PR #439, opened 2013-11-22
FLUID-5220

jobara:
Improves the performance of composite panels by not requiring double creation of each sub panel.
http://issues.fluidproject.org/browse/FLUID-5220


amb26:
Horrifying.... I "reviewed" this blunderous test and merged it myself!
See commentary on https://issues.fluidproject.org/browse/FLUID-5220

PR #544, opened 2014-07-11
FLUID-5458: Updated npmignore (for master)

jobara:
Removed ant related files that no longer exist, also ignoring the preferences framework.

http://issues.fluidproject.org/browse/FLUID-5458

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,9 +5,6 @@ demos
 examples
 Gruntfile.js
 products
-project.xml
-project.properties
-pom.xml
 
 src/components
 src/lib/fonts
@@ -15,6 +12,7 @@ src/lib/jquery/plugins
 src/lib/jquery/ui
 src/lib/json
 src/framework/fss
+src/framework/preferences
 
 tests/3rd-party-tests
 tests/component-tests


colinbdclark:
@jobara and @amb26, what needs to be done to this pull request to finish it up and get us to the point where we can publish Infusion as an npm module?

amb26:
We need to come to an agreement about what parts of the framework's image will be excluded from our npm module image through entries in .npmignore - at least for our initial "beta" releases of infusion-2.0.0

colinbdclark:
So is the primary issue whether or not parts of the Preferences Framework should be included? And is there any refactoring necessary in order to do so?

amb26:
Well, the issue is rather theoretical, since I'm not aware of any practical use that these files could be put to - but I think we just need some clear guidelines about the kind of reasoning we will use in order to either include or exclude parts of the framework from the module. My model was that any code which could even in theory execute without a DOM document should be included, but I'm open to other models.

colinbdclark:
Yeah, my model was that it was only portions of the code that could be put to practical use would be included, but of course "practical use" itself is somewhat fuzzy. I guess we probably have two main clients of the Infusion node module: the GPII and Flocking, neither of whom would likely use portions of the Preferences Framework, but who knows?

I'm willing to go with a model where anything that can be included should be included, but presumably we need some means of testing whether the module actually works. As far as I know, we don't have any tests for it currently, other than to run either the GPII or Flocking and confirm that they work?

jobara:
@colinbdclark and @amb26 we did have the other use case of the OAuth2 security work for GPII. For the client side portion we intended to also make use of the Infusion framework supplied via NPM (which is used for the server side), but had to also pull in missing pieces with a [custom build](https://github.com/GPII/universal/blob/master/gpii/node_modules/gpii-oauth2/gpii-oauth2-authz-server/public/lib/infusion/VERSION.md). Of course we also needed the preferences framework for the security work. If we strip the Infusion NPM module all the way down to just the framework, users who want to do server and client side work from the same repo will have to manually mash them together. 

The options I'm thinking of are:
* Have two NPM infusion modules (framework only and everything)
* Include all of Infusion in the NPM module
* Framework only NPM module, client side will have to pull in anything they need manually
  * For this case we'll need to come up with a best practice (e.g. share part of Infusion from NPM and bring in the rest manually, bring in two copies of infusion, etc.)
  * We should actually start to consider bower and other client side package managers.

colinbdclark:
Those three options seem like a reasonable summary of the situation. Do you have a preference for one of them, @jobara?

Wouldn't a client-side developer pretty much always have to make an Infusion custom build of some kind, since they probably aren't using every component included in Infusion, and will want to deliver a minified (and minimal) script to web clients?

jobara:
@colinbdclark good point about wanting a custom version for the client-side. I'd argue for last option then. But maybe the distinction shouldn't have been Framework/Other, but about requiring the DOM or not, as @amb26 mentioned above.

amb26:
Ok, well I'm happy for the time being to accept @colinbdclark 's suggested model above, "this module contains only code which we have ever successfully run in node". (What Wittgenstein might call a "finitist" model - "Turing - how many numerals have you learned to write? One possible answer is - 'as many numerals as I ever have written'" : P) - we can think about extending this as we progress towards the 2.0 release. I'll merge this pull as-is once we have the big merge. I agree that it looks like we need a completely separate packaging system for infusion on the client, probably involving bower and/or a build.

colinbdclark:
Sounds like a good compromise, @amb26 and @jobara. I, for one, will be really happy to have Infusion and related projects released into the npm registry for easier versioning.

PR #572, opened 2014-11-19
FLUID-5512: Fix for compound valueMapper issue, by consolidating deepEq algorithm with the one now implemented in DataBinding. 

amb26:
This mildly extends the algorithm to count unchanged values as well as changed ones, and implements a new ranking algorithm and option accepted by the valueMapper named "partialMatches" which may appear on an individual option as well as on the valueMapper transform as a whole. Rationalised naming of "transform" to "transformer" as the last argument of a transform


amb26:
@cindyli - the first comment (11 Dec) doesn't have an attached line number but perhaps it is referring to "iff" on line 1096 of DataBinding.js - this is not a typo but a standard abbreviation signifying "if and only if" https://en.wikipedia.org/wiki/If_and_only_if

amb26:
For 2nd comment - unchanged is always counted during every comparison.

cindyli:
Merged @ c860fe56239a70aa057532426a36457af11279e3

PR #577, opened 2015-01-13
FLUID-5585: Fixes for failure to relay removal of model material

amb26:


amb26:
We should hold up this pull request for a while, since I think we would like the behaviour which it changes to be configurable - between the "old model" where a relay applies a MERGE at its endpoints, and the behaviour in this pull where it applies an ADD. This implies we should wait until we properly have both MERGE and ADD back in the ChangeApplier system.

amb26:
@the-t-in-rtf - this has now been dusted off and updated with master - please do report back with results of experiments and if possible further test cases for cases you find it doesn't handle

jobara:
@amb26 sounds like this needs some testing/verification that it addresses the issues. Will it actually be ready in time for the Infusion 2.0 release and is it required to be a part of that?

the-t-in-rtf:
The best place to test this is in gpii-binder, which is currently building against [my branch with the range of new transforms](https://github.com/GPII/gpii-binder/pull/5).  I'll have to create a new branch that merges in the fix, commit that, and test.  As long as no code changes are needed in the merged branch, there doesn't need to be a separate PR, I'd just toss the branch out once both PRs are merged.

the-t-in-rtf:
I have updated the branch of gpii-binder under PR to include tests for deleting values and relaying deletions.  You can see the results here:

https://github.com/the-t-in-rtf/gpii-binder/tree/GPII-2077 

Hilariously, my previous tests had a typo in a grade name, which prevented me from seeing that they were also completely broken.  I discovered this when I added an initial check to confirm that the value exists before it is cleared.  I fixed the typo, which revealed other problems with handling of empty strings.  Once I fixed those, the "clear a value" tests worked as expected.

I then added a child component and added a check to confirm that the child component is made aware of the deletion.  This test fails with my previous branch of infusion, as expected.  The applier deletes the value in the "local" model, but this change is not relayed.

When I updated to [the new branch that merges Antranig's work with mine](https://github.com/the-t-in-rtf/infusion/tree/FLUID-5971-5585), I get immediate failures in running testem:

```
Tonys-MacBook-Pro:gpii-binder duhrer$ testem
Infusion at path /Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion is at top level 
/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/enhancement/js/ContextAwareness.js:210
        return navigator.platform ? navigator.platform : undefined;
               ^

ReferenceError: navigator is not defined
    at fluid.contextAware.getPlatformName (/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/enhancement/js/ContextAwareness.js:210:16)
    at Object.fluid.invokeGlobalFunction (/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/core/js/Fluid.js:1112:25)
    at Object.<anonymous> (/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/enhancement/js/ContextAwareness.js:56:30)
    at transformInternal (/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/core/js/Fluid.js:468:34)
    at Object.fluid.transform (/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/core/js/Fluid.js:497:17)
    at Function.fluid.contextAware.performChecks (/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/enhancement/js/ContextAwareness.js:43:22)
    at Function.fluid.contextAware.makeChecks (/Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/enhancement/js/ContextAwareness.js:78:47)
    at /Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/enhancement/js/ContextAwareness.js:213:24
    at /Users/duhrer/Source/rtf/gpii-binder/node_modules/infusion/src/framework/enhancement/js/ContextAwareness.js:219:3
    at Object.exports.runInContext (vm.js:44:17)
```

If I open the test file directly without using Testem, both tests now fail, i.e. the change applier does not delete the value from the "local" model, and the change is not relayed.

Anyway, you should be able to see both the testem and browser test failures in [the gpii-binder branch](https://github.com/the-t-in-rtf/gpii-binder/tree/GPII-2077).  Please advise.

amb26:
@waharnum - looks like this was caused by the merge of https://github.com/fluid-project/infusion/pull/757 for FLUID-5973, which has broken trunk in all environments other than plain browsers - node tests/node-tests/basic-node-tests.js is also broken. Would you like to prepare a fix?
If only we had a CI system which caught such problems :)

waharnum:
@amb26 Sorry about that - a very basic error on my part (the code functionally assumes Infusion is run only in browser and has the navigator object available, which is obviously not true). Will prepare a quick patch.

waharnum:
@amb26 / @the-t-in-rtf - PR for this issue at https://github.com/fluid-project/infusion/pull/766

Sorry again for such an elementary error.

colinbdclark:
Regarding @amb26's "if only we had a CI system" comment: Isn't it less an issue of not having CI to catch this—since we do have Vagrant-automated tests for Infusion that can be run manually by a developer prior to committing—and more a question of making sure we have [tests that cover the use of Infusion within Node.js](https://issues.fluidproject.org/browse/FLUID-5744)? Or am I confused? :)

waharnum:
@colinbdclark - agreed. as a basic guard (this would have caught the issue in FLUID-5973, as I was running the VM tests as a final check before PR submission), I think we should do something like the following as part of `grunt runTests`:

*Current*: `vagrant ssh -c 'cd /home/vagrant/sync/; DISPLAY=:0 testem ci --file tests/testem.json'` (this runs all the browser-based tests in the Vagrant VM)

*Proposed*: `vagrant ssh -c 'cd /home/vagrant/sync/; DISPLAY=:0 testem ci --file tests/testem.json' && node tests/node-tests/basic-node-tests.js`

This will run the basic node tests "to verify the basic packaging of Fluid Infusion as a node module
performed by the material in src/module/fluid.js" in addition to the browser tests.

I'll create a separate JIRA for this and make a PR, as the change seems straightforward.

jobara:
@waharnum, the testem tests conveniently report back the set of test runs. I wonder if you could fold in the results of the node tests into this report. My concern is that running the node tests separate will obscure the reports from the other tests.

waharnum:
@jobara yes, I should have tagged that as pseudocode work in progress :) I'll see what it would take to report the Node tests as part of the overall testem run.

colinbdclark:
@jobara, it's a good point, but may not be entirely trivial to implement in a short time. I think @waharnum's pseudocode is a good first step if it turns out to be onerous to integrate this more smoothly. It's how some of the GPII tests work as well, and tests I've written for my other libraries, too. Not ideal, but better than nothing.

waharnum:
@colinbdclark - I've actually just about got this working. Should have a PR in the next 45 minutes or so.

the-t-in-rtf:
@waharnum, let me know the branch and hash once you have this fixed.  I suspect I will see the same failures as I do in a plain old browser, but it would be good to confirm.

the-t-in-rtf:
nm, I see it in the PR for FLUID-5985.

the-t-in-rtf:
Thanks for the quick fix, @waharnum.  With the most recent master merged with my local branch, I'm now able to run testem and confirm that the second "relay" test is still failing.  The non-relay test (that deletes content from the local model) works.

the-t-in-rtf:
The other failure was apparently simply related to not running within testem.

the-t-in-rtf:
I have committed the updates to [the gpii-binder branch](https://github.com/the-t-in-rtf/gpii-binder/tree/GPII-2077) so that you can see the test failures in context.

colinbdclark:
In a call with @cindyli, @the-t-in-rtf, and @amb26, we agreed that we should defer this issue until after Infusion 2.0. It shouldn't block us from freezing the code for release.

the-t-in-rtf:
@amb26, I have confirmed that all the tests that rely on deletion-relaying in gpii-binder now pass in all supported browsers.  Thanks for the investigative work and the quick fix.

tests/framework-tests/core/js/DataBindingTests.js
```diff
@@ -2076,6 +2076,170 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+    fluid.tests.fluid5585.fixtureFuncs = [
+        "fluid.tests.fluid5585.verifyBackwardRelay",
+        "fluid.tests.fluid5585.verifyForwardRelay",
+        "fluid.tests.fluid5585.verifyForwardPartialDeleteRelay",
+        "fluid.tests.fluid5585.verifyBackwardPartialDeleteRelay"
+    ];
+
+    fluid.each(fluid.tests.fluid5585.fixtureFuncs, function (fixtureFunc) {
+        fluid.tests.fluid5585.runOneConfiguration("fluid.tests.fluid5585.explicitRelay", fixtureFunc);
```
cindyli: Line 2206 and 2207 are identical. Did you mean to have one of them test `fluid.tests.fluid5585.implicitRelay`?

amb26: Very well spotted! Pushed fix for this horrifying blunder... thankfully the tests still pass at least :)

tests/framework-tests/core/js/DataBindingTests.js
```diff
@@ -2297,6 +2461,6 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
         return togo;
     };
 
-    fluid.test.runTests(["fluid.tests.fluid5659root"]);
+    // fluid.test.runTests(["fluid.tests.fluid5659root"]);
```
cindyli: Remove this line?

amb26: Oops... I often comment out that line when debugging since even when disabled, IoC framework tests generate so much "construction junk" at startup ...

cindyli:
Merged at 07cc0d5ba6fe85a4b910751c9db8cec6c501e3bc

PR #580, opened 2015-01-26
FLUID-5592: Added a test case for FLUID-5592

cindyli:
Even though this test case didn't manage to expose the issue.


amb26:
Closed for https://github.com/fluid-project/infusion/pull/581

PR #581, opened 2015-01-27
FLUID-5592 Fix for case where an earlier modelListener destroys the component holding an applier notified by a later listener in the same stack

amb26:
This implementation is quite a mess and requires three separate pathways to be guarded. In theory we could clean up by inserting a wrapper around each listener, but this would i) create memory and stack-frame garbage as well as ii) still not guarding the "classic" pathway leading through fluid.resolveModelListener


amb26:
This is one of the unavoidable "costs of doing business" in a synchronous world

cindyli:
Thanks for the fix, @amb26. I've tested this pull request with my FLUID-5552 branch (https://github.com/cindyli/infusion/tree/FLUID-5552) where this issue was originated from. The fix does resolve the problem.

cindyli:
Merged into the master branch @ 605b0bcfaedf6193c31456236d3aa4e42cf36a96

PR #582, opened 2015-01-27
FLUID-5552: Upgrade the prefs framework to use the core model relay functionality.

cindyli:
JIRA: http://issues.fluidproject.org/browse/FLUID-5552

This pull request is dependent on the other pull request https://github.com/fluid-project/infusion/pull/581 that needs to be merged in first.

As some styling for emphasizing links and enlarging inputs have been changed, before testing prefs framework demos or examples, run "grunt buildStylus" first.


demos/prefsFramework/js/SelfVoicingEnactor.js
```diff
@@ -34,6 +34,9 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
         model: {
             value: false
```
amb26: Now that we have moved exclusively to the new relay system, we can eliminate all of this wrapping with the spurious path "value" everywhere - all of these cooperating components can save on configuration and runtime garbage by simply declaring model: false etc.

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -66,123 +62,117 @@ var fluid_2_0 = fluid_2_0 || {};
...
-        that.events.afterRender.fire(that);
-    };
+    });
 
     fluid.defaults("fluid.textfieldSlider.textfield", {
-        gradeNames: ["fluid.viewComponent", "autoInit"],
+        gradeNames: ["fluid.viewRelayComponent", "autoInit"],
+        range: {}, // should be used to specify the min, max range e.g. {min: 0, max: 100}
+        modelRelay: {
+            target: "value",
```
amb26: Remember to eliminate "value" from this component as everywhere

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -66,123 +62,117 @@ var fluid_2_0 = fluid_2_0 || {};
...
+            }
+        },
+        modelListeners: {
+            value: {
+                "this": "{that}.container",
+                "method": "val",
+                args: ["{change}.value"]
+            },
+            // When the input field receives a invalid input such as a string value, ignore it and populate the field with the previous value
+            stringValue: {
```
amb26: This is a very interesting setup, that we need to talk over and write up. It depends on an odd subtlety in order to work - that an update to the value held in our model only occurs at a coarse granularity - say, whenever we receive a "change" event or some other similar checkpoint. If these updates occurred any faster, the user would be infuriated by being unable to type into the field. This is an area in which UX has spilled over into our model configuration, which to some extent we should try to prevent - especially since with our current implementation, the user could never reconfigure this listener. At the very least, we need to give a namespace to this listener - unfortunately the current dataBinding implementation will ignore it. Still, we should add the namespace with a comment that explains that having this listener here represents an interaction decision rather than an implementation decision, and that we need to be able to revisit the issue of what the behaviour is here.

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -66,123 +62,117 @@ var fluid_2_0 = fluid_2_0 || {};
...
-        gradeNames: ["fluid.viewComponent", "autoInit"],
+        gradeNames: ["fluid.viewRelayComponent", "autoInit"],
+        range: {}, // should be used to specify the min, max range e.g. {min: 0, max: 100}
         selectors: {
             thumb: ".ui-slider-handle"
         },
-        events: {
-            modelChanged: null
+        members: {
+            slider: {
```
amb26: Just convert this into a genuine Infusion component named fluid.slider and we can eliminate the oddity of the member + expander and the strange function which just dispatches to its function argument. This can eventually become elevated to a standard wrapper along the same lines as the existing "fluid.tooltip" etc.

amb26: Sorry, this comment wasn't very clear. It's clear that this component really IS the slider - therefore the entire thing should just be renamed "fluid.slider". You should be aware that all modern jQuery UI widgets don't actually return anything from their "creator" - if you look into this member called "slider" you will find that it either contains nothing or just the same DOM element as container. So we can eliminate this expander in favour of just calling initSlider (which we should probably rename as initJQuerySlider or some such) as a lifecycle action - that is, putting back the onCreate that we used to have. We can break out the initAria call into another listener.

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -385,7 +390,7 @@ var fluid_2_0 = fluid_2_0 || {};
      * Resets the selections to the integrator's defaults and fires onReset
      */
     fluid.prefs.prefsEditor.reset = function (that) {
-        that.updateModel(fluid.copy(that.rootModel));
+        that.applier.change("", fluid.copy(that.rootModel));
```
amb26: Leave a note here that this may not be completely effective if the root model is smaller than the current one. Given our previous discoveries re "model shrinkage" (http://issues.fluidproject.org/browse/FLUID-5585 ), the proper thing to do here is to apply a DELETE to the root before putting in the new model. And this should be done within a transaction in order to avoid notifying the tree more than necessary. However, the transactional model of the changeApplier is going to change radically soon (re. "New New Notes") and this implementation doesn't seem to be causing a problem at present so we had just better leave it the way it is for now, with a comment pointing to the potential problem and requirements.

src/framework/preferences/js/StarterGrades.js
```diff
@@ -97,23 +97,15 @@ var fluid_2_0 = fluid_2_0 || {};
...
                 type: "fluid.prefs.enactor.textSize",
                 container: "{uiEnhancer}.container",
                 options: {
                     fontSizeMap: "{uiEnhancer}.options.fontSizeMap",
-                    rules: {
-                        "textSize": "value"
-                    },
                     model: {
-                        value: "{fluid.prefs.rootModel}.rootModel.textSize"
+                        value: "{uiEnhancer}.model.textSize"
```
amb26: This is odd - why does the parent's component model bind to {fluid.prefs.rootModel} whereas all the subcomponents bind to {uiEnhancer}.model?

src/framework/preferences/js/StarterGrades.js
```diff
@@ -97,23 +97,15 @@ var fluid_2_0 = fluid_2_0 || {};
 
     fluid.defaults("fluid.uiEnhancer.starterEnactors", {
         gradeNames: ["fluid.uiEnhancer", "fluid.uiEnhancer.cssClassEnhancerBase", "fluid.uiEnhancer.browserTextEnhancerBase", "autoInit"],
-        connectionsGrade: "fluid.prefs.uiEnhancerConnections",
-        distributeOptions: {
-            source: "{that}.options.connectionsGrade",
-            removeSource: true,
-            target: "{that > fluid.prefs.enactor}.options.gradeNames"
-        },
+        model: "{fluid.prefs.rootModel}.rootModel",
```
amb26: I've got a little lost here and forgotten what the "rootModel" component actually does - it appears that it acts here as a source of initial values (defaults?) for this grade. We should probably rename it to something a little more descriptive now that we have other models at the "root". Does "defaultRootModel" or "rootDefaultModel" seem reasonable?

src/framework/preferences/js/Store.js
```diff
@@ -131,7 +131,11 @@ var fluid_2_0 = fluid_2_0 || {};
     });
 
     fluid.tempStore.set = function (settings, applier) {
-        applier.requestChange("", settings);
+        if ($.isEmptyObject(settings)) {
+            applier.fireChangeRequest({path: "", type: "DELETE"});
+        } else {
+            applier.change("", settings);
```
amb26: As per the previous discussion, we should do the DELETE first _every_ time, not just if the model is empty - it might have shrunk for some other reason other than being entirely empty. As a general "code smells" note, any call to $.isEmptyObject needs to be flagged up as suspicious - I've almost never seen a call to it that was really necessary. Although KASPARNET did manage to convince me of one in the GPII once.

demos/prefsFramework/js/SelfVoicingEnactor.js
amb26: Following my later comment - I think we can't eliminate "value" from absolutely everywhere - but I think we can from all the enactors.

src/framework/preferences/js/Panels.js
```diff
@@ -143,7 +133,8 @@ var fluid_2_0 = fluid_2_0 || {};
      * this is all done, the onDomBind event is fired.
      */
     fluid.prefs.subPanel.resetDomBinder = function (that) {
-        that.container = $(that.container.selector);
+        var userJQuery = that.container.constructor;
```
amb26: This is unusual - was this change required by a bug that was discovered? This code is copied from the framework and should be marked with a TODO to be eliminated once we have the new renderer

src/framework/preferences/js/Panels.js
```diff
@@ -317,13 +296,28 @@ var fluid_2_0 = fluid_2_0 || {};
...
             }
-        });
 
+            // Construct the model relay btw the composite panel and its subpanels
+            var componentRelayRules = {};
+            var preferenceMap = fluid.get(fluid.prefs.compositePanel.prefetchComponentOptions(componentOptions.type, componentOptions.options), ["preferenceMap"]);
+            fluid.each(preferenceMap, function (prefObj, prefKey) {
+                $.each(prefObj, function (prefRule) {
+                    if (prefRule.indexOf("model.") === 0) {
+                        fluid.set(componentRelayRules, prefRule.slice(6), "{compositePanel}.model." + fluid.prefs.subPanel.safePrefKey(prefKey));
```
amb26: Write "model.".length here rather than 6 to make the meaning of the magic constant clear

src/framework/preferences/js/Panels.js
```diff
@@ -317,13 +296,28 @@ var fluid_2_0 = fluid_2_0 || {};
...
+                    }
+                });
+            });
+            fluid.set(relayOption, [componentName], componentRelayRules);
```
amb26: Given the path seems to have just one segment, couldn't this more simply just be 
relayOption[componentName] = componentRelayRules?

src/framework/preferences/js/Panels.js
```diff
@@ -317,13 +296,28 @@ var fluid_2_0 = fluid_2_0 || {};
                     target: "{that > " + componentName + "}.options"
                 });
             }
-        });
 
+            // Construct the model relay btw the composite panel and its subpanels
+            var componentRelayRules = {};
+            var preferenceMap = fluid.get(fluid.prefs.compositePanel.prefetchComponentOptions(componentOptions.type, componentOptions.options), ["preferenceMap"]);
```
amb26: Split this complex operation into two lines

src/framework/preferences/js/Panels.js
```diff
@@ -317,13 +296,28 @@ var fluid_2_0 = fluid_2_0 || {};
...
+            fluid.each(preferenceMap, function (prefObj, prefKey) {
+                $.each(prefObj, function (prefRule) {
```
amb26: Why $.each rather than fluid.each here?

src/framework/preferences/js/Panels.js
```diff
@@ -310,6 +288,7 @@ var fluid_2_0 = fluid_2_0 || {};
     fluid.prefs.compositePanel.assembleDistributeOptions = function (components) {
         var gradeName = "fluid.prefs.compositePanel.distributeOptions_" + fluid.allocateGuid();
         var distributeRules = [];
```
amb26: This can now be named "distributeOptions" to make its function completely clear

tests/framework-tests/preferences/js/BuilderTests.js
```diff
@@ -52,8 +52,8 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
      ************************************************/
 
     fluid.tests.testparseAuxSchema = function (expected, funcArgs) {
```
amb26: Should be testParseAuxSchema

tests/framework-tests/preferences/js/EnactorsTests.js
```diff
@@ -17,18 +17,27 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 
     fluid.registerNamespace("fluid.tests");
 
-    fluid.tests.testStyle = function (that, container, expectedDefaultFlag, expectedCssClass) {
-        var elements = that.getElements();
+    fluid.tests.containSubstring = function (string, subString) {
```
amb26: Should be containsSubstring - but a clearer model should use this jQuery utility http://api.jquery.com/hasclass/ rather than string-bashing

tests/framework-tests/preferences/js/ModelRelayTests.js
```diff
@@ -1,161 +0,0 @@
-/*
```
amb26: yay yay yay they are alle ***GGGONNE***

src/framework/preferences/js/Panels.js
cindyli: Yes, the bug discovered was in the separated panel version of the UIO, dom binders of sub-panels were not reset due to the use of the incorrect jQuery instance. I'll add a TODO comment for this line.

src/framework/preferences/js/StarterGrades.js
cindyli: Binding the parent's component model with {fluid.prefs.rootModel} is to set the default root model.

Binding all the sub-components' model to {uiEnhancer}.model is to establish model relay in btw the uiEnhancer and actual enactors.

amb26: Ok - could we rename rootModel to something more descriptive? Such as initialModel or defaultModel

cindyli: Sure. will rename to initialModel.

src/components/textfieldSlider/js/TextfieldSlider.js
cindyli: The "change" event fires when the user leaves the input field, so the user can finish typing without having to worry about the model getting updated too fast.

Another question, I tried to add namespace to the model listener like `"stringValue.ignoreStringInput"`, however, this entire listener is ignored without being executed. Does the current dataBinding not yet support the namespaced model listeners?

src/components/textfieldSlider/js/TextfieldSlider.js
cindyli: The member "slider" does return a jQuery slider object that has all slider functions attached onto. If I move initSlider to an "onCreate" listener, the instantiation of the slider would happen too late that the initial triggering of the model listener to set the initial slider value (https://github.com/cindyli/infusion/blob/FLUID-5552/src/components/textfieldSlider/js/TextfieldSlider.js#L163) complaints the undefined function of slider().

As a result, the expander for initializing the slider member option is kept. But the code is refactored to get rid of the oddity of passing setSliderAria() as an argument to the invoker setSliderValue().

cindyli:
Thanks for the comments, @amb26. This pull request is ready for another review.

src/components/textfieldSlider/js/StringToNumberTransform.js
```diff
@@ -0,0 +1,30 @@
...
+var fluid_2_0 = fluid_2_0 || {};
+
+(function ($, fluid) {
+    "use strict";
+
+    /**
+     * Transform a string to a number
+     **/
+
+    fluid.defaults("fluid.transforms.stringToNumber", {
```
amb26: This should just be added to the main ModelTransformationTransforms file

src/components/textfieldSlider/js/TextfieldSlider.js
amb26: Yes, namespaces are currently ignored in databinding. However we should leave a comment here to explain the problems that might arise with this implementation.

cindyli:
@amb26, this pull request is ready for another round of review.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
```diff
@@ -39,93 +39,67 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-            slider.slider("value", valToTest);
-            jqUnit.assertEquals("Slider value should be " + expected, expected, slider.slider("value"));
             textfield.val(valToTest);
             textfield.change();
             jqUnit.assertEquals("Textfield value should be the " + expected, expected, textfield.val());
             jqUnit.assertEquals("The ARIA value now should be " + expected, expected, thumb.attr("aria-valuenow"));
+            jqUnit.assertEquals("Slider value should be " + expected, expected, slider.slider("value"));
+        };
+
+        var testSlider = function (valToTest, expected) {
```
amb26: Sorry not to catch this in earlier rounds of the review - could you put all these global functions into a fluid.tests.textfieldSlider namespace? Thanks

cindyli: Good catch. Addressed and back ready.

amb26:
Thanks so much for all the really excellent work on this essential framework upgrade, Cindy! This sets us forward enormously - I will be able to remove support for the old ChangeApplier in my upcoming branch next month.

jobara:
merged at 18939c0dd614901bfc434a29f880aa332600a239

PR #583, opened 2015-01-29
FLUID-5589: Text-To-Speech (component wrapper, prefsFramework integration)

jobara:
Created a wrapper component around the SpeechSynthesis api, including a new method that can be used by the progressive enhancement to check for support.

Also updated the enactor and panel that was used by prefsFramework demo and migrated these into the Infusion prefsFramework and added tests.

http://issues.fluidproject.org/browse/FLUID-5589


src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -0,0 +1,105 @@
...
+            start: null,
+            end: null,
+            error: null,
+            pause: null,
+            resume: null,
+            mark: null,
+            boundary: null
+        },
+        utteranceOpts: {
+            // text: "", // text to synthesize. avoid as it will override any other text passed in
```
amb26: Why don't we supply actual values for these defaults - is the component usable without them?

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -0,0 +1,105 @@
...
+
+var fluid_2_0 = fluid_2_0 || {};
+
+(function ($, fluid) {
+    "use strict";
+
+    fluid.registerNamespace("fluid.textToSpeech");
+
+    fluid.textToSpeech.isSupported = function () {
+        return !!(window && window.speechSynthesis);
```
amb26: This should contain a link in a comment to the relevant HTML5 API spec (or whatever form it takes) so that the reader can know what API is expected here

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -0,0 +1,105 @@
...
+            },
+            resume: {
+                "this": "speechSynthesis",
+                "method": "resume"
+            },
+            getVoices: {
+                "this": "speechSynthesis",
+                "method": "getVoices"
+            },
+            isPending: {
```
amb26: We should see if we can instead represent this using a model - by relaying the relevant events from "utterance" onto model listeners which change these three flags. Then the user can just subscribe to those events instead - or else just inspect the model state.

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -0,0 +1,151 @@
...
+        preferenceMap: {
+            "fluid.prefs.speak": {
+                "model.value": "default"
+            }
+        },
+        model: {
+            value: false
+        },
+        events: {
+            onStart: null,
```
amb26: If this were done via model relay from the model from the TTS component, we could eliminate all these listeners and event definitions etc. This of course would require cindyli's prefs framework update to go in, but I think that that was extremely close to being ready.

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -0,0 +1,151 @@
...
+                        "pause.hoist": "{speakEnactor}.events.onPause",
+                        "resume.hoist": "{speakEnactor}.events.onResume",
+                        "error.hoist": "{speakEnactor}.events.onError"
+                    }
+                }
+            }
+        },
+        invokers: {
+            announce: {
+                funcName: "fluid.prefs.enactor.speakEnactor.announce",
```
amb26: This should really be named something like "queueAnnounce" or "queueSpeech" and the underlying method in TTS should be given the same name for consistency - to emphasise to the reader that it does not speak the supplied text immediately but adds it to a queue.

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -0,0 +1,119 @@
...
+
+            jqUnit.assertTrue("The Text to Speech component should have initialized", that);
+            jqUnit.assertFalse("Nothing should be speaking", that.isSpeaking());
+            jqUnit.assertFalse("Nothing should be pending", that.isPending());
+            jqUnit.assertFalse("Shouldn't be paused", that.isPaused());
+        });
+
+        jqUnit.asyncTest("Start and Stop Events", function () {
+            jqUnit.expect(8);
+            var that = fluid.tests.textToSpeech({
```
amb26: Make this into a proper grade as usual - in general avoid large amounts of in-line configuration in functions

tests/framework-tests/preferences/js/SelfVoicingEnactorTests.js
```diff
@@ -0,0 +1,209 @@
...
+            selfVoicingTester: {
+                type: "fluid.tests.selfVoicingTester"
+            }
+        }
+    });
+
+    fluid.defaults("fluid.tests.selfVoicingTester", {
+        gradeNames: ["fluid.test.testCaseHolder", "autoInit"],
+        testOptions: {
+            // Because events are fired for each utterance
```
amb26: We should improve the event API of TTS so that this can be tested properly. It should have an "onQueueEmpty" event that fires once all speech has stopped.

tests/framework-tests/preferences/js/PanelTestUtils.js
```diff
@@ -0,0 +1,45 @@
...
+        var newval = fluid.get(newModel, path);
+        jqUnit.assertEquals("Expected model value " + expectedValue + " at path " + path, expectedValue, newval);
+    };
+
+    fluid.tests.panels.utils.setCheckboxState = function (element, state) {
+        element.prop("checked", state).change();
+    };
+
+    fluid.tests.panels.utils.verifyCheckboxState = function (message, expectedState, checkbox) {
+        jqUnit[expectedState ? "assertTrue" : "assertFalse"](message, checkbox.is(":checked"));
```
amb26: jqUnit.assertEquals(message, expectedState, checkbox.is(":checked")) will do fine here

src/components/textToSpeech/js/TextToSpeech.js
jobara: yes, these are all optional. And depending on the synthesizer used could have different results.. For example, volume and rate may or may not be respected by the synthesizer. Specifying the text would override any text actually passed in. so this one is a bit dangerous to use. Unless you really have a specific use case for it i guess.

src/components/textToSpeech/js/TextToSpeech.js
jobara: Is this comment referring specifically to the isSupported function or the file as a whole?

src/framework/preferences/js/SelfVoicingEnactor.js
jobara: Do you have an ETA for when that will go in. We will need the TTS stuff for our First Discovery Tool work. The "deliverable" of the First Discovery Tool is due Feb 13.

src/components/textToSpeech/js/TextToSpeech.js
amb26: Placing it near the head of the file close to this first use of the API should be good.

src/framework/preferences/js/SelfVoicingEnactor.js
jobara: This modelization work will need to wait till after https://github.com/fluid-project/infusion/pull/582. For now I will just add a TODO as a comment in the code.

See: https://botbot.me/freenode/fluid-work/2015-01-30/?tz=America/Toronto

src/components/textToSpeech/js/TextToSpeech.js
jobara: I have a couple question about this. Should the related events be removed ( https://github.com/jobara/infusion/blob/FLUID-5589/src/components/textToSpeech/js/TextToSpeech.js#L50-L58 ). Also for something like isPending, there isn't a related event. The best i could do is try to track what has been read. For example keep track of start and stop events and try to calculate what's left. Does doing that make sense?

amb26: Yes, that's correct - the related events should be removed, and you will need to synthesize the "isPending" model state yourself. But all of this may as well be headed off until we can modelise the entire hierarchy by using @cindyli 's work

jobara: Okay, so I should just make this another TODO? or is there useful work here? For example related to your comment on the tests.

amb26: I guess what you could do is just modelise this component and just connect it to the enactor using modelListeners rather than relay. That might at least let you improve the tests.

jobara:
@amb26 ready for another round of review

demos/prefsFramework/js/SelfVoicingEnactor.js
```diff
@@ -1,170 +0,0 @@
...
-        var nodes = $(that.currentElement).contents();
-        var next = fluid.find(nodes, function (node) {
-            if (that.seen.indexOf(node) > -1) {return;}
-
-            if (node.nodeType === 8 || node.nodeName === "SCRIPT" || node.nodeName === "IFRAME") {
-                that.seen.push(node);
-                return;
-            }
-
-            if (node.nodeType === 3) {
```
amb26: Could you add comments here describing the symbolic values of these nodeTypes?

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -0,0 +1,178 @@
...
+        utteranceOpts: {
+            // text: "", // text to synthesize. avoid as it will override any other text passed in
+            // lang: "", // the language of the synthesized text
+            // voiceURI: "" // a uri pointing at a voice synthesizer to use. If not set, will use the default one provided by the browser
+            // volume: 1, // a value between 0 and 1
+            // rate: 1, // a value from 0.1 to 10 although different synthesizers may have a smaller range
+            // pitch: 1, // a value from 0 to 2
+        },
+        // TODO: When this is updated to model relay component
+        // remove this manual realying of modelListeners to listeners
```
amb26: typo here

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -0,0 +1,178 @@
...
+        // in favour of having integrators use the modelListeners directly.
+        // This system is currently setup because when a the entire model
```
amb26: typo "when a the"

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -0,0 +1,119 @@
...
+        jqUnit.test("Initialization", function () {
+            var that = fluid.tests.textToSpeech();
+
+            jqUnit.assertTrue("The Text to Speech component should have initialized", that);
+            jqUnit.assertFalse("Nothing should be speaking", that.model.speaking);
+            jqUnit.assertFalse("Nothing should be pending", that.model.pending);
+            jqUnit.assertFalse("Shouldn't be paused", that.model.paused);
+        });
+
+        fluid.defaults("fluid.tests.textToSpeech.startStop", {
```
amb26: This defaults block should be moved out to top level

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -0,0 +1,119 @@
...
+            jqUnit.expect(10);
+            var that = fluid.tests.textToSpeech.startStop();
+            that.queueSpeech("Testing start and end events");
+        });
+
+        // Chrome doesn't properly support pause which causes this test to break.
+        // see: https://code.google.com/p/chromium/issues/detail?id=425553&q=SpeechSynthesis&colspec=ID%20Pri%20M%20Week%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified
+        if (!window.chrome) {
+
+            fluid.defaults("fluid.tests.textToSpeech.pauseResume", {
```
amb26: Defaults block should be at top level, even if the test code is conditional

jobara:
@amb26 ready for another round of review

PR #588, opened 2015-03-17
FLUID-5600: loaderGrades passed in via aux schema

jobara:
Rather than taking in a prefsEditorType option at instantiation, grades for the prefsEditorLoader are now passed in through the schema. This not only keeps all the configuration in a centralized location, but also allows for more configuration on the prefsEditorLoader.

http://issues.fluidproject.org/browse/FLUID-5600


jobara:
@amb26 could you please review?

PR #589, opened 2015-03-18
FLUID-5597: Pass message bundles to individual panels

jobara:
Rather than combining all of the message bundles into a single bundle that gets passed around, the specific message bundles for each panel are now directly passed to them.

http://issues.fluidproject.org/browse/FLUID-5597


jobara:
@amb26 could you please review this pull request?

src/framework/preferences/js/Panels.js
```diff
@@ -546,12 +559,18 @@ var fluid_2_0 = fluid_2_0 || {};
                 var valuebinding = tree.valuebinding;
                 var modelValue = fluid.get(model, fluid.prefs.compositePanel.rebaseValueBinding(valuebinding, modelRelayRules));
                 return modelValue !== undefined ? modelValue : val;
+            } else if (key === "messagekey") {
+                // converts the "UIMessage" renderer component into a "UIBound"
+                // and passes in the resolved message as the value.
+                rebasedMessagekey.componentType = "UIBound";
```
amb26: This should be done in place, replacing the existing structure rather than the ad-hoc scheme which assigns to an outside closure variable

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -177,22 +154,27 @@ var fluid_2_0 = fluid_2_0 || {};
...
-        delete resources.expander;   // A work-around for FLUID-5117
+    fluid.prefs.resourceLoader.loadResources = function (that, resources, transformFn) {
         fluid.fetchResources(resources, function () {
-            that.resources = resources;
+            that.resources = transformFn(resources);
             that.events.onResourcesLoaded.fire(resources);
         });
     };
 
+    fluid.prefs.resourceLoader.parseMessages = function (resources) {
```
amb26: More efficient to have this done by the jQuery ajax machinery - can we arrange to have this done by a suitable "dataType" option passed through as we do the fetch? Then we can lose the whole "transformResources" workflow

jobara:
@amb26 I've made the change to use the jquery machinery to parse the resource. This is ready for another round of review.

jobara:
@amb26 ready for another round of review.

src/framework/preferences/js/Panels.js
```diff
@@ -528,28 +540,55 @@ var fluid_2_0 = fluid_2_0 || {};
...
-                var modelValue = fluid.get(model, fluid.prefs.compositePanel.rebaseValueBinding(valuebinding, modelRelayRules));
-                return modelValue !== undefined ? modelValue : val;
-            } else {
-                return val;
-            }
-        });
+    fluid.prefs.compositePanel.rebaseTreeComp = function (msgResolver, model, treeComp, memberName, modelRelayRules) {
+        var rebased = fluid.copy(treeComp);
+
+        if (rebased.children) {
```
amb26: I think this would look much clearer with "else" blocks expressing that only one branch is possible at a time - unless there is indeed some case whereby more than one could execute - which would be good to highlight if so!

jobara: @amb26 yes, it is possible for more than one to execute. For example ID is often hit. value and valuebinding would be another example. It had worked before with `else` blocks because i was looping through these properties but now i'm looping through the parent and getting the entire object back.

jobara:
@amb26 I've added a comment about the if statements. Ready for more review.

jobara:
@amb26 refactored rebaseTreeComp to use as many `else` blocks as possible. Ready for more review.

jobara:
@amb26 more cleanup on rebaseTreeComp. Ready for more review.

jobara:
merged at 809231c7525798ead15f214497721b6ea9e619f7

PR #597, opened 2015-04-29
FLUID-5659: Fix for relay saturation issue in multiple relay - new core utility fluid.flatten

amb26:
This resolves the relay issue you encountered with the language panel yesterday, @cindyli


cindyli:
Thanks for the fix, @amb26. I tested it with my branch where the issue occurred, this pull request does fix the problem.

jobara:
merged at c0c17164e10e46ac498af9d758c0cfcb93dac7d5

PR #598, opened 2015-05-04
FLUID-5599: Improve the preferences framework to fetch localised message bundles

cindyli:
@amb26, can you review this pull request? Thanks.

PR #600, opened 2015-05-06
FLUID-5662: Fix for resource stall in case two requests are queued to the same nonexistent resource

amb26:


tests/framework-tests/core/data/messages4.json
```diff
@@ -0,0 +1,3 @@
+{
+    "same": "Are you Mr. Sherlock Holmes?"
+}
```
cindyli: Need a new line char at the end of the file.

cindyli:
Tested with https://github.com/fluid-project/infusion/pull/598 and the first discovery tool where the issue was discovered, this pull request does fix the problem.

cindyli:
Merged @ 7b1fdc80a06339fb84ae8c07a128aefc5380c68c

PR #602, opened 2015-05-12
FLUID-4257: Fix for at least one case of failure to clear injected listener

amb26:
from destroyed component (in the case the raw function handle was the same as that registered by another listener), expanded test case coverage

@colinbdclark  - there's a chance that this corrects your issue from last night. If not, please could you supply an alternative test case.


amb26:
This should probably actually be attached to https://issues.fluidproject.org/browse/FLUID-5546, FLUID-4257 looks like an irrelevant accidental branch name

amb26:
@colinbdclark - this one is also now ready

PR #610, opened 2015-06-17
FLOE-4138: fixed file name for keyboard a11y tests

jobara:
Renamed keyboard a11y test files to follow the naming convention.

https://issues.fluidproject.org/browse/FLUID-4138


cindyli:
Shall demos/keyboard-a11y directory also be renamed for consistency?

jobara:
@cindyli, the underlying file is called jquery.keyboard-a11y.js. So I think it might make sense to leave the demo. We could also change the tests to be Keyboard-a11y-test.html and Keyboard-a11yTests.js. What do you think?

cindyli:
I think it makes sense to change tests to be Keyboard-a11y-test.html and Keyboard-a11yTests.js

jobara:
@cindyli renamed the files, ready for more review.

cindyli:
The last commit message should use "FLUID-4138" rather than "FLOE-4138".

jobara:
@cindyli this is ready for another round of review.

cindyli:
Merged @ 47b15a3f8a2c16ee3474c7baf13c8a2798dce244

PR #612, opened 2015-06-19
FLUID-5675: Add a mock version of text to speech

jobara:
Added a mock version of fluid.textToSpeech which mocks the calls to the Web Speech API. This will allow for easier integration testing and remove the need for these tests to be run in a browser that supports the Web Speech API (e.g. Firefox).
    
The unit tests for the selfVoicingEnactor have been updated to make use of the MockTTS.

https://issues.fluidproject.org/browse/FLUID-5675


src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -46,7 +47,11 @@ var fluid_2_0 = fluid_2_0 || {};
         str.replace(/\s{2,}/gi, " ");
 
         if (that.model.enabled && str) {
-            fluid.textToSpeech.queueSpeech(that, str, interrupt, options);
+            if (typeof(speechFn) === "string") {
+                fluid.invokeGlobalFunction(speechFn, [that, str, interrupt, options]);
+            } else {
+                speechFn(that, str, interrupt, options);
```
amb26: This change in contract needs to be documented and explained

tests/component-tests/textToSpeech/js/MockTTS.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        speechRecord.push(record);
+
+        that.queue.push(text);
+        that.events.onSpeechQueued.fire(text);
+
+        // mocking speechSynthesis speak
+        handleStart();
+        // using setTimeout to preserve asynchronous behaviour
```
amb26: If this is required, the specification that handleEnd is invoked asynchronously should be added to fluid.textToSpeech

tests/component-tests/textToSpeech/js/MockTTS.js
```diff
@@ -0,0 +1,89 @@
...
+*/
+
+// Declare dependencies
+/* global fluid */
+
+(function () {
+    "use strict";
+
+    // Mocks the fluid.textToSpeech component, removing calls to the
+    // Web Speech API. This will, amongst other things, remove asynchrony
```
amb26: Looks like it won't remove it entirely now

tests/framework-tests/preferences/js/SelfVoicingEnactorTests.js
```diff
@@ -148,11 +129,21 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
      *******************************************************************************/
 
     fluid.defaults("fluid.tests.prefs.enactor.selfVoicingEnactor", {
-        gradeNames: ["fluid.prefs.enactor.selfVoicing", "fluid.tests.prefs.enactor.speakEnactor", "autoInit"],
+        gradeNames: ["fluid.viewRelayComponent", "fluid.tests.mock.textToSpeech", "fluid.prefs.enactor.selfVoicing", "autoInit"],
```
amb26: Perhaps the need for this awkward grade list would be best met by refactoring fluid.prefs.enactor.selfVoicing so that it uses containment rather than grade merging to bind to the TTS impl. If it had a fluid.textToSpeech as a subcomponent rather than having it merged on to itself, it would be easier for this to be replaced, e.g. by distributeOptions or subcomponent overriding. The corruption of the grade structure emerged because it was ambiguous whether this component was intended to be a view component or not - the enactor is one, and the TTS isn't. This seems to be a clear sign that the relationship should be refactored.

jobara:
@amb26 this is ready for another round of review.

jobara:
Another issue that has come up, while trying to run tests in sauce labs, is that the feature exists in the browser but the audio drivers have been stripped from the OS. This prevents the web speech api from being engaged. We should also implement a detection for this case.

jobara:
Added a check to ensure that if the SpeechSynthesis API is available, that for a given delay, a call to speak actually triggers speech events. This has been implemented in the TextToSpeechTests for the following reasons:

* It is the only test that will make use of the SpeechSynthesis API, integrations can use a mock.
* This is primarily for running the tests in SauceLabs, which doesn't provide an audio driver
* Because the check is asynchronous, it was necessary to use a promise to handle the running/not running of tests. Typical progressive enhancement checks are all synchronous. Therefor it didn't seem reasonable to make this a general check as it wouldn't conform to normal usage.

jobara:
@amb26 ready for more review, includes new check for tests to work in SauceLabs.

colinbdclark:
@avtar can you also test this with your Sauce Labs-based test automation system and verify for @amb26 and @jobara that it works in that context?

jobara:
@amb26 could you also review this related pull request to update the build site https://github.com/fluid-project/build.fluidproject.org/pull/16

avtar:
@amb26, @colinbdclark, @jobara These two tests don't timeout anymore:

SelfVoicingEnactor-test.html https://saucelabs.com/tests/dec6557ad2da495b987fb0485f60981d
TextToSpeech-test.html https://saucelabs.com/tests/7e4ba925fcb942aa90f65b7c2c28e928

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -11,13 +11,40 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 // Declare dependencies
-/* global fluid, jqUnit, speechSynthesis */
+/* global fluid, jqUnit, speechSynthesis, SpeechSynthesisUtterance */
 
 (function () {
     "use strict";
 
     fluid.registerNamespace("fluid.tests");
 
+    // Ensures that TTS is supported in the browser, including cases wehre the
```
amb26: typo wehre -> where

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -11,13 +11,40 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+    // feature is detected, but where the underlying audio engine is missing.
+    // For example VM's on SauceLabs.
```
amb26: Typo VM's -> VMs

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -11,13 +11,40 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+    //
+    // returns a promise
```
amb26: What does the returned promise represent? If this is intended to be a supported API, we should add a proper doc-style comment /** ... @param, @return etc. */ explaining the implementation and its strategy. What does the "delay" argument do?

jobara: I've only included this check in the tests because it runs asynchronously. All of the other standard progressive enhancement checks are all synchronous, so introducing this inline with those could/would cause confusion for the time that they would resolve. So I'm not sure what you mean by supported. I believe it is useful to the tests, but not necessarily for general use. What do you think?

amb26: I think that i) this code provides a feature detection which is generally useful, and could well be useful outside just the testing scenario, and that ii) it should be moved out of the testing files into the main implementation, and iii) its documentation should be improved

jobara: I did push a change to the comment, could you let me know how it is now. I'll look into moving it into the main implementation if the comment is okay.

jobara:
@amb26 I've made the requested changes. This is ready for another code review. I did move the checkTTSSuport method from the tests to the implemention. @avtar you may want to test that it still addresses the issue you had with test automation on sauce labs.

jobara:
merged at 70ea372fd461554567ae453c5523a4b30d8cc438

PR #614, opened 2015-06-24
FLUID-5697: provide a means of excluding headers

jobara:
Added a class that can be used to exclude a header or section of headers from being included in the Table of Contents.

https://issues.fluidproject.org/browse/FLUID-5697

### DIFF
--- a/demos/prefsFramework/index.html
+++ b/demos/prefsFramework/index.html
@@ -90,7 +90,7 @@
             });
         </script>
 
-        <div class="flc-overviewPanel fl-overviewPanel-container"></div>
+        <div class="flc-overviewPanel flc-toc-exclude fl-overviewPanel-container"></div>
         <!-- BEGIN markup for Preference Editor -->
         <div class="flc-prefsEditor-separatedPanel fl-prefsEditor-separatedPanel">
             <!-- This is the div that will contain the Preference Editor component -->
--- a/demos/prefsFramework/index.html
+++ b/demos/prefsFramework/index.html
@@ -90,7 +90,7 @@
             });
         </script>
 
-        <div class="flc-overviewPanel flc-toc-exclude fl-overviewPanel-container"></div>
+        <div class="flc-overviewPanel fl-overviewPanel-container"></div>
         <!-- BEGIN markup for Preference Editor -->
         <div class="flc-prefsEditor-separatedPanel fl-prefsEditor-separatedPanel">
             <!-- This is the div that will contain the Preference Editor component -->
--- a/demos/tableOfContents/index.html
+++ b/demos/tableOfContents/index.html
@@ -39,68 +39,85 @@
             <nav class="row">
                 <div class="flc-toc-tocContainer demo-toc large-4 small-centered columns"></div>
             </nav>
-
-            <h1>Amphibians</h1>
+            <h1>Species</h1>
+            <h2>Amphibians</h2>
             <p>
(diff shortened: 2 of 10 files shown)


tests/component-tests/tableOfContents/js/TableOfContentsTests.js
```diff
@@ -620,5 +620,35 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+        /**
+        * #FLUID-5697: Test the exclusion of headers
+        */
+        jqUnit.asyncTest("FLUID-5697: Header exclusion", function () {
+            // craft headingInfo so renderTOCTest() can use it
+            var testHeadings = {
+                headingInfo: [{
+                    level: "2",
+                    text: "Inculded",
```
amb26: Typo here Inculded

demos/tableOfContents/index.html
```diff
@@ -39,68 +39,85 @@
             <nav class="row">
                 <div class="flc-toc-tocContainer demo-toc large-4 small-centered columns"></div>
             </nav>
-
-            <h1>Amphibians</h1>
+            <h1>Species</h1>
+            <h2>Amphibians</h2>
```
amb26: Why were all of these headings moved down one level in index, and why didn't this affect any test output?

demos/prefsFramework/index.html
```diff
@@ -90,7 +90,7 @@
             });
         </script>
 
-        <div class="flc-overviewPanel fl-overviewPanel-container"></div>
+        <div class="flc-overviewPanel flc-toc-exclude fl-overviewPanel-container"></div>
```
amb26: I think this style of configuration may have been implemented backwards. Rather than requiring to intrude on all the markup in the page with a "flc-toc-exclude" class, I think it would make more sense for this to be the integrator's job - there should be a cleaner way, for example, of specifying that flc-overviewPanel, the existing selector, represents a thing should should be excluded. This requires "unbundling" the giant selector that TOC uses for exclusions into a form which is easier to configure piece by piece

src/components/tableOfContents/js/TableOfContents.js
```diff
@@ -112,7 +112,7 @@ var fluid_2_0 = fluid_2_0 || {};
             tocHeader: "Table of Contents"
         },
         selectors: {
-            headings: ":header:visible:not(.flc-toc-tocContainer :header)",
+            headings: ":header:visible:not(.flc-toc-tocContainer :header, .flc-toc-exclude :header, .flc-toc-exclude)",
```
amb26: See earlier comment - I think this feature would be more usable if it was possible for users to freely contribute "things to ignore" into a map, rather than having to unpick the logic of a single giant selector. Unfortunately the configuration "selectorsToIgnore" is taken already : P I suggest something like 

ignoreForToC: {
   overview: ".flc-overviewPanel",
   key2: "mySelector2"
}
etc
The implementation would then collect together all the values in this structure and ignore them collectively.

demos/tableOfContents/index.html
jobara: As part of the a11y review we had for UIO it was mentioned that it is atypical to have multiple `<h1>`'s on a page. To avoid any issues in this regard for the demo, i just made a new top level `<h1>` and dropped everything down a level. This did not affect the tests because it is the demo, not the tested html file.

jobara:
@amb26 this pull request is ready for another round of review.

PR #615, opened 2015-06-24
FLUID-4977: Changed A in colour/contrast label to pseudo element for AT

maozillah:
Prevent AT from reading letter 'A' in label by replacing it with pseudo element


src/framework/preferences/js/Panels.js
```diff
@@ -841,7 +841,7 @@ var fluid_2_0 = fluid_2_0 || {};
             theme: ["default", "bw", "wb", "by", "yb", "lgdg"]
         },
         markup: {
-            label: "<span class=\"fl-preview-A\">A</span><span class=\"fl-hidden-accessible\">%theme</span><div class=\"fl-crossout\"></div>"
+            label: "<span class=\"fl-preview-A\"></span><span class=\"fl-hidden-accessible\">%theme</span><div class=\"fl-crossout\"></div>"
```
jobara: While the "a" is no longer in the markup. It seems that the screen reader is still picking it up. This is probably because the "a" content added via css is still picked up by the screen reader as part of the label.

klown: @jobara It's likely due to the accessible name computation that browsers implement.  As a first approximation:  include text specified in CSS content rules.  The details are here:
http://w3c.github.io/aria/accname-aam/accname-aam.html#step2F, especially bullet ii.

Thus, it's not necessarily that the screen reader is picking it up; rather, the browser is telling the screen reader:  the name is, "...", and the "..." includes CSS text content.

jobara: @klown thanks for the clarification.

klown: @jobara You're welcome.

I can't remember if, in the past, I suggested using aria-label as a solution in this case.  Maybe I did, and you had reasons for not using it.  FWIW, consider the following:

&lt;label for="radio" aria-label="Default"&gt;A&lt;/label&gt; &lt;input type="radio" id="radio"&gt;

With that markup:
- the name of the radio button is "Default".
- the text "Default" is not visually rendered.
- the text "A" is visually rendered.

Hope that helps.

maozillah: @jobara Tested nvda on all the browsers and it seems like the only one that is reading the "a" is firefox. gonna do some research see if there's a quick workaround before trying @klown 's suggestion

jobara:
It seems like the default contrast is still trying to read a special character. Possibly for the "crossout". Any chance you could remove that as well.

screen shot 2015-07-03 at 10 58 08 am

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -169,6 +169,10 @@
         .fl-preview-A {
             font-size: 2em;
         }
+        
+        .fl-preview-A:before {
+            content: "a";
```
jobara: I wonder if the content should be moved back to the markup for the label. It would be easier to change this all in one place and for localization purposes as well. What do you think?

src/framework/preferences/js/Panels.js
```diff
@@ -841,7 +841,7 @@ var fluid_2_0 = fluid_2_0 || {};
             theme: ["default", "bw", "wb", "by", "yb", "lgdg"]
         },
         markup: {
-            label: "<span class=\"fl-preview-A\">A</span><span class=\"fl-hidden-accessible\">%theme</span><div class=\"fl-crossout\"></div>"
+            label: "<span class=\"fl-preview-A\" role=\"presentation\" aria-hidden=\"true\"></span><span class=\"fl-hidden-accessible\">%theme</span><div class=\"fl-crossout\"></div>"
```
jobara: do we need the "role" and "aria-hidden" attributes?

jobara:
Could you please provide unit tests for the aria-label that is set?

src/framework/preferences/css/stylus/PrefsEditor.styl
maozillah: Hey Justin, this is needed for screenreaders to read it properly in browsers other than firefox. Otherwise all other browsers will still say the 'a'

src/framework/preferences/js/Panels.js
maozillah: The aria-hidden attribute is needed for screen readers on safari to ignore the "a"

jobara: can we remove the role="presentation"? Also please leave a comment for why aria-hidden and aria-label are set.

src/framework/preferences/css/stylus/PrefsEditor.styl
jobara: Could you please leave a comment for why the content is set in the CSS.

src/framework/preferences/js/Panels.js
maozillah: Yep have removed the role, and for sure I'll go in and leave those comments

maozillah:
Ready for review

src/framework/preferences/css/PrefsEditor.css
```diff
@@ -1 +1 @@
...
...or .fl-choice input:focus ~ label{outline:.2em solid #000}.fl-prefsEditor .fl-choice label.fl-theme-by,.fl-prefsEditor .fl-choice label.fl-theme-by span,.fl-prefsEditor .fl-choice label.fl-theme-by .fl-crossout{background-color:#ff0 !important;border-color:#000 !important;color:#000 !important}.fl-theme-lgdg .fl-prefsEditor.fl-slider{background:none;background-color:#bdbdbb;}.fl-theme-lgdg .fl-prefsEditor.fl-slider a,.fl-theme-lgdg .fl-prefsEditor.fl-slider a:hover{border:1px solid #bdbdbb;background-color:#555}.fl-theme-lgdg .fl-prefsEditor .fl-prefsEditor-switch,.fl-theme-lgdg .fl-prefsEditor .fl-prefsEditor-switch-inner,.fl-theme-lgdg .fl-prefsEditor input + label .fl-prefsEditor-switch:before{box-shadow:none !important;text-shadow:none;color:inherit;background-image:none;background-color:inherit}.fl-theme-lgdg .fl-prefsEditor input:focus + label,.fl-theme-lgdg .fl-prefsEditor .fl-choice input:focus ~ label{outline:.2em solid #bdbdbb}.fl-prefsEditor .fl-choice label.fl-theme-lgdg,.fl-prefsEditor .fl-choice label.fl-theme-lgdg span,.fl-prefsEditor .fl-choice label.fl-theme-lgdg .fl-crossout{background-color:#555 !important;border-color:#bdbdbb !important;color:#bdbdbb !important}
```
jobara: Could you remove this file from this pull request too?

maozillah: Removed, not sure why my commits consistently include the *.css

jobara:
@maozillah, could you please update the copyright in the PanelsTests.js file. It should be "Copyright 2013-2015 OCAD University" now.

maozillah:
Updated all OCAD copyright notices

jobara:
@maozillah thanks, but you should only update the copyrights for the files that were changed this year. Were all of those modified?

maozillah:
@jobara Ah gotcha, reverted the commit and made the single copyright change

jobara:
merged at 4c2ba7903a0543fdc0f78edc79fd436ef4bee0ae

PR #617, opened 2015-07-10
FLUID-5665: Move preferences data into a specific model path "preferences"

cindyli:
@amb26, can you review this pull request.

The corresponding pull request for the infusion-docs change is https://github.com/fluid-project/infusion-docs/pull/70


src/framework/preferences/js/PrefsEditor.js
```diff
@@ -362,38 +363,60 @@ var fluid_2_0 = fluid_2_0 || {};
...
 
     fluid.prefs.prefsEditor.saveAndApply = function (that) {
-        that.save();
-        that.events.onPrefsEditorRefresh.fire();
-        that.applyChanges();
+        var prevSettings = that.getSettings(),
+            newSelections = that.save();
+
+        // Only when preferences are changed, re-render panels and trigger enactors to apply changes
+        if (!fluid.model.diff(fluid.get(newSelections, "preferences"), fluid.get(prevSettings, "preferences"))) {
```
amb26: There's no need for an expensive diff here - the plan was always just to have a standard model listener listening at path "preferences"

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -362,38 +363,60 @@ var fluid_2_0 = fluid_2_0 || {};
...
-                delete savedSelections[key];
-            }
-        });
-        that.events.onSave.fire(savedSelections);
-        that.setSettings(savedSelections);
+        var initialModel = that.initialModel,
+            userSelections = fluid.copy(that.model);
+
+        // Only save the changed preferences so the future default value change on preferences can still be correctly merged with changed preferences
+        if (fluid.model.diff(userSelections.preferences, initialModel.preferences)) {
```
amb26: Much more efficient to just make a single call to diff and make use of the "changeMap" structure which is built up in its options.

src/framework/preferences/js/PrefsEditor.js
cindyli: In the full page demo: http://build.fluidproject.org/infusion/tests/manual-tests/framework/preferences/fullPage/, the changed preferences should only be applied when "save and apply" button is clicked. If we use a standard model listener, every model change on "preferences" path will cause enactors to applied changes to the page, which is not desired.

src/framework/preferences/js/PrefsEditor.js
cindyli: This comment has been addressed.

tests/framework-tests/preferences/js/PrefsEditorTests.js
```diff
@@ -109,8 +126,8 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
             prefsEditor("#ui-options");
         };
 
-        var resetSaveCalled = function () {
-            saveCalled = false;
+        var resetSaveCalled = function (that) {
+            that.options.saveCalled = false;
```
amb26: Must not write to a component's options - this was already corrected in my branch to the following:

https://github.com/amb26/infusion/blob/FLUID-5249/tests/framework-tests/preferences/js/PrefsEditorTests.js#L90

tests/framework-tests/preferences/js/PrefsEditorTests.js
```diff
@@ -68,9 +66,20 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
                     }
                 },
+                saveCalled: false,
+                prefsEditorRefreshed: 0,
                 listeners: {
-                    onSave: function () {
-                        saveCalled = true;
+                    "onSave.called": {
+                        listener: function (that) {
+                            that.options.saveCalled = true;
```
amb26: See below - could you try to harmonise these changes with the ones in my branch?

cindyli:
@amb26, improved prefsEditorTests.js. Ready for another look.

PR #619, opened 2015-08-05
Fluid 5682: Added ARIA for Hide/show frame to explain behaviour to AT

maozillah:
Added 
* aria-controls, dynamic aria-pressed to hide/show button
* dynamic aria-expanded to panel
* aria-controls to reset button

As well as tests for the corresponding additions

PR #620, opened 2015-08-11
Fluid 5680: Hide icons from AT and added aria-label for hide/show button

maozillah:
* aria-labels for icons removed with the addition of `display = "presentation"`
* aria-labels added for show/hide button

This pull request should be merged **after** [FLUID-5682](https://github.com/fluid-project/infusion/pull/619)

### DIFF
--- a/src/components/slidingPanel/js/SlidingPanel.js
+++ b/src/components/slidingPanel/js/SlidingPanel.js
@@ -25,10 +25,6 @@ var fluid_2_0 = fluid_2_0 || {};
             toggleButton: ".flc-slidingPanel-toggleButton",
             toggleButtonLabel: ".flc-slidingPanel-toggleButton"
         },
-        strings: {
-            showText: "show",
-            hideText: "hide"
-        },
         events: {
             onPanelHide: null,
             onPanelShow: null,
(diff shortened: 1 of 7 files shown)


maozillah:
Tested and working with by tabbing with voiceover on chrome, firefox, and safari and NVDA on windows with chrome, firefox, and IE

src/components/slidingPanel/js/SlidingPanel.js
```diff
@@ -25,10 +25,6 @@ var fluid_2_0 = fluid_2_0 || {};
             toggleButton: ".flc-slidingPanel-toggleButton",
             toggleButtonLabel: ".flc-slidingPanel-toggleButton"
         },
-        strings: {
-            showText: "show",
-            hideText: "hide"
-        },
```
jobara: I think you need to keep some default text, for the sliding panel. Actually you might want to add default strings for all of the string options. It seems the sliding panel tests are failing because of this.

jobara:
@maozillah is this updated from the FLUID-5682 branch? Also, you should add tests for the arai-label in the sliding panel tests.

maozillah:
@jobara Yes it is, I needed the code from that to do this.

maozillah:
Not sure why, but the aria-label is not being added to the toggleButton in the test case while the other attributes are. Tried writing the aria-label functionality in in jquery similar to aria-pressed but it's still not showing up.

jobara:
@amb26 will you be taking this over to update to the latest framework changes?

PR #623, opened 2015-08-17
FLUID-5711: Softens references to "window" for use within Workers.

colinbdclark:
This pull request includes a basic unit test that at least verifies that a) Infusion is correctly evaluated within a Web Worker and b) that a component can be instantiated correctly and the response from an invoker sent as a message back to the main thread.


colinbdclark:
I have tested this on:
* Mac OS X: Firefox, Safari, Chrome
* Windows 10: Edge, IE 11
* Node.js using the basic-node-tests.js (is there a way to run all the non-view framework tests in Node.js, too?)

PR #630, opened 2015-08-27
FLUID-5755: Fix to failed detection of non-plain objects in 2nd branch

amb26:
of mergeOneImpl - this logic all needs revisiting for FLUID-4982

Thanks for the report and test case, @colinbdclark


colinbdclark:
Thanks for the speedy fix, @amb26!

PR #631, opened 2015-09-02
Fluid 5745: Includes all useful portions of Infusion in its npm module.

colinbdclark:
This pull request modifies .npmignore so that all of Infusion's source is available to npm, so that users can use npm to source Infusion for both Node and browser-based projects.


colinbdclark:
@amb26 or @jobara, can one of you review this?

.npmignore
```diff
@@ -1,25 +1,9 @@
...
-tests/framework-tests
-tests/manual-tests
-tests/all-tests.html
-tests/lib/jquery-ui
-tests/lib/mockjax
-tests/lib/README
-tests/test-core/utils/js/DebugFocus.js
+.gitignore
+.jshintignore
+.jshintrc
```
jobara: Should we leave the jshint files? They aren't really necessary for a build, but at least the .jshintrc file is used by a task in Gruntfile.js.

.npmignore
```diff
@@ -1,25 +1,9 @@
...
+.npmignore
+.project
+demos/
+examples/
+tests/
```
jobara: Did we want to leave in the testing infrastructure and just remove the actual tests. This seems to wipeout all of the tests including the ones specified in [devIncludes.json](https://github.com/fluid-project/infusion/blob/master/src/module/devIncludes.json)

jobara:
Should we also add the "products" directory to prevent any build artifacts from being published?

.npmignore
colinbdclark: I'm not sure I understand. Can you elaborate? Thanks!

colinbdclark:
@jobara I've updated this pull request based on your review.

.npmignore
```diff
@@ -1,25 +1,13 @@
...
+.project
+build/
+demos/
+examples/
+products/
+tests/component-tests/
+tests/framework-tests/
+tests/manual-tests/
+tests/node-tests/
+tests/node-tests/all-tests.html
```
jobara: was "tests/node-tests/all-tests.html" supposed to be "tests/all-tests.html"?

jobara: we should probably also remove "tests/test-core/testTests" which are the unit tests for the IoC Testing Framework. This file should probably actually be under the jqUnit tests or somewhere similar in the framework-tests directory.

colinbdclark:
Good eye once again, @jobara! Updated again based on your comments.

jobara:
merge at a1e3d601478e57276d4143fca5bcbe00f9c6b0cd

PR #632, opened 2015-09-03
FLOE-5756: Moved MockTTS to TextToSpeech src dir

jobara:
https://issues.fluidproject.org/browse/FLUID-5756


cindyli:
The moved file needs to be added into the dependency json file (https://github.com/fluid-project/infusion/blob/master/src/components/textToSpeech/textToSpeechDependencies.json)

jobara:
@cindyli it's not a dependency of the TextToSpeech, it shouldn't be included in the concatenated build. It's really just a useful tool for testing.

cindyli:
Merged @ 6273ecc911990511481e25d8da1c11ab4b9f74b3

PR #637, opened 2015-09-18
FLUID-5770: Fix an issue with the prefsEditor save function

cindyli:
The issue was fluid.prefs.prefsEditor.save() only saves changes on the model path "preferences" instead of other parts of the prefsEditor model.

JIRA: https://issues.fluidproject.org/browse/FLUID-5770

### DIFF
--- a/src/framework/preferences/js/PrefsEditor.js
+++ b/src/framework/preferences/js/PrefsEditor.js
@@ -372,7 +372,7 @@ var fluid_2_0 = fluid_2_0 || {};
 
     /**
      * Saves the current model and fires onSave
- */
+     */
     fluid.prefs.prefsEditor.save = function (that) {
         if (!that.model) {  // Don't save a reset model
             return;
@@ -380,7 +380,7 @@ var fluid_2_0 = fluid_2_0 || {};
 
         var initialModel = that.initialModel,
             userSelections = fluid.copy(that.model),
-            changedSelections = {};
+            changedPrefs = {};
 
         // Only save the changed preferences so the future default value change on preferences can still be correctly merged with changed preferences
         if (fluid.model.diff(userSelections.preferences, initialModel.preferences)) {
@@ -391,13 +391,17 @@ var fluid_2_0 = fluid_2_0 || {};
             fluid.model.diff(userSelections.preferences, fluid.get(that.initialModel, ["preferences"]), stats);
 
             fluid.each(stats.changeMap, function (state, path) {
-                fluid.set(changedSelections, ["preferences", path], userSelections.preferences[path]);
+                fluid.set(changedPrefs, path, userSelections.preferences[path]);
             });
+
+            if (stats.changes > 0) {
+                userSelections.preferences = changedPrefs;
+            }
         }
 
-        that.events.onSave.fire(changedSelections);
-        that.setSettings(changedSelections);
-        return changedSelections;
+        that.events.onSave.fire(userSelections);
+        that.setSettings(userSelections);
+        return userSelections;
     };
 
     fluid.prefs.prefsEditor.saveAndApply = function (that) {
--- a/src/framework/preferences/js/PrefsEditor.js
+++ b/src/framework/preferences/js/PrefsEditor.js
@@ -379,29 +379,30 @@ var fluid_2_0 = fluid_2_0 || {};
         }
 
         var initialModel = that.initialModel,
-            userSelections = fluid.copy(that.model),
+            modelToSave = fluid.copy(that.model),
             changedPrefs = {};
 
-        // Only save the changed preferences so the future default value change on preferences can still be correctly merged with changed preferences
-        if (fluid.model.diff(userSelections.preferences, initialModel.preferences)) {
-            delete userSelections.preferences;
+        // Only save the changed preferences so the future initial model value change on unchanged
+        // preferences can still be correctly merged with changed preferences
+        if (fluid.model.diff(modelToSave.preferences, initialModel.preferences)) {
+            delete modelToSave.preferences;
         } else {
             var stats = {changes: 0, unchanged: 0, changeMap: {}};
 
-            fluid.model.diff(userSelections.preferences, fluid.get(that.initialModel, ["preferences"]), stats);
+            fluid.model.diff(modelToSave.preferences, fluid.get(that.initialModel, ["preferences"]), stats);
 
             fluid.each(stats.changeMap, function (state, path) {
-                fluid.set(changedPrefs, path, userSelections.preferences[path]);
+                fluid.set(changedPrefs, path, modelToSave.preferences[path]);
             });
 
             if (stats.changes > 0) {
-                userSelections.preferences = changedPrefs;
+                modelToSave.preferences = changedPrefs;
             }
         }
 
-        that.events.onSave.fire(userSelections);
-        that.setSettings(userSelections);
-        return userSelections;
+        that.events.onSave.fire(modelToSave);
+        that.setSettings(modelToSave);
+        return modelToSave;
     };
 
     fluid.prefs.prefsEditor.saveAndApply = function (that) {
--- a/src/framework/preferences/js/PrefsEditor.js
+++ b/src/framework/preferences/js/PrefsEditor.js
@@ -372,7 +372,7 @@ var fluid_2_0 = fluid_2_0 || {};
 
     /**
      * Sends the prefsEditor.model to the store and fires onSave
-     * @param that: An fluid.prefs.prefsEditor instance
+     * @param that: A fluid.prefs.prefsEditor instance
      * @return the saved model
      */
     fluid.prefs.prefsEditor.save = function (that) {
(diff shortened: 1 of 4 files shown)


src/framework/preferences/js/PrefsEditor.js
```diff
@@ -391,13 +391,17 @@ var fluid_2_0 = fluid_2_0 || {};
             fluid.model.diff(userSelections.preferences, fluid.get(that.initialModel, ["preferences"]), stats);
 
             fluid.each(stats.changeMap, function (state, path) {
-                fluid.set(changedSelections, ["preferences", path], userSelections.preferences[path]);
+                fluid.set(changedPrefs, path, userSelections.preferences[path]);
```
jobara: Do you think it would make sense if an integrator could decide what they wanted to save? It might make sense for some cases to only save the preferences and for others to save everything. One possible way to facilitate this would be to allow the path to be passed in.

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -380,7 +380,7 @@ var fluid_2_0 = fluid_2_0 || {};
 
         var initialModel = that.initialModel,
             userSelections = fluid.copy(that.model),
-            changedSelections = {};
+            changedPrefs = {};
```
jobara: The name seems wrong, if the changes are more than just what's stored under model.preferences.

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -391,13 +391,17 @@ var fluid_2_0 = fluid_2_0 || {};
...
+                userSelections.preferences = changedPrefs;
+            }
         }
 
-        that.events.onSave.fire(changedSelections);
-        that.setSettings(changedSelections);
-        return changedSelections;
+        that.events.onSave.fire(userSelections);
+        that.setSettings(userSelections);
+        return userSelections;
```
jobara: why do we pass along userSelections instead of changedPrefs?

src/framework/preferences/js/PrefsEditor.js
cindyli: One reason that the save function saves the entire model is because of the special handling of the "preferences" path that only changed preferences are saved instead of a direct save of all preferences. This allows the ability when the initial model gets changed later for unchanged preferences, the new initial values can still be merged correctly with the saved prefs. This special handling is currently transparent to integrators. Allowing users to pass in the path parameter would increase the complexity of maintaining this transparency, in terms of figuring out where "preferences" are since they need a special handling, merging the value of the given path into the saved model, the possibility of parts of the saved model and the component model could be out of sync etc. Also, as far as I know, we don't have an use case so far that requires to save a specific path of the prefsEditor model.

I think this topic worths more discussions and likely a separate JIRA.

src/framework/preferences/js/PrefsEditor.js
cindyli: To inform what has been saved. The saved pieces are changed prefs and other parts in the model.

src/framework/preferences/js/PrefsEditor.js
cindyli: Whatever changes on other parts of the model will be saved and no special handling is needed for those parts.

src/framework/preferences/js/PrefsEditor.js
jobara: We've only been saving the preferences, there wasn't a need to save other model value before. Currently that's just a special case by the firstDiscovery tool. Could we always save the preferences, and have an option to save the rest of the model or portion there of?

src/framework/preferences/js/PrefsEditor.js
jobara: I see now, the userSelections has been updated with the changedPrefs.

src/framework/preferences/js/PrefsEditor.js
jobara: I see that the userSelections has everything. I think the names are still confusing though. Maybe we should change userSelections to modelToSave or something. As there may be other things in there than user Selection which were analogous to preference selections before.

src/framework/preferences/js/PrefsEditor.js
jobara: So i think this comment is at the wrong spot in the code. I guess i want an option where we just save changedPrefs and not the entire userSelections. Before I believe something like {contrast: "bw"} would have been saved, but in the current implementation "{preferences: {contrast: "bw"}}" would be saved. However, that will require existing implementations to change their data structure even though they aren't saving any other material.

src/framework/preferences/js/PrefsEditor.js
cindyli: Makes sense. Changed the variable name to the suggested.

src/framework/preferences/js/PrefsEditor.js
cindyli: Which comment do you refer to for being at the wrong spot in the code.

To answer your question in the previous comment of "have an option to save the rest of the model or portion there of", I would say, for the data that integrators are not intended to be saved, don't relay to the prefsEditor model. The prefsEditor model is empty on its own but being populated from the initialModel whose data are meant to be tracked.

cindyli: I refactored the save function to save for specific model path. As long as the path is relative to the model root, we are able to find and handle preferences path specially. Let me know if the refactored is what you were thinking. Thanks.

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -371,38 +371,49 @@ var fluid_2_0 = fluid_2_0 || {};
...
         if (!that.model) {  // Don't save a reset model
             return;
         }
 
-        var initialModel = that.initialModel,
-            userSelections = fluid.copy(that.model),
-            changedSelections = {};
+        var path = fluid.makeArray(modelPath).join("."),
+            currentModel = fluid.copy(that.model),
+            modelToSave = path === "" ? currentModel : that.getSettings();
```
jobara: It seems strange that undefined, null, and "" will all cause the entire model to save. I'm thinking we either support a flag like saveEntireModel (boolean) or we allow for an array of paths such as [["path, "one"], "path.two"] where "" represents the entire model and undefined/null represent only the preferences. It seems like we only have a use case for the first example at this point, so it is probably enough. How does that sound?

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -435,7 +446,9 @@ var fluid_2_0 = fluid_2_0 || {};
         var bindHandlers = function (that) {
             var saveButton = that.locate("save");
             if (saveButton.length > 0) {
-                saveButton.click(that.saveAndApply);
+                saveButton.click(function () {
+                    that.saveAndApply("");
```
jobara: Should we leave the default as it has been, that is, only saving the preferences and not the entire model?

cindyli:
@jobara, According to our discussion with Colin and Simon, I've modified the save function to send the entire prefsEditor.model to the store.

This pull request is ready for a look. Thanks.

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -371,33 +371,36 @@ var fluid_2_0 = fluid_2_0 || {};
...
-        } else {
-            var stats = {changes: 0, unchanged: 0, changeMap: {}};
+        var modelToSave = fluid.copy(that.model),
+            initialModel = that.initialModel,
+            stats = {changes: 0, unchanged: 0, changeMap: {}},
+            changedPrefs = {};
 
-            fluid.model.diff(userSelections.preferences, fluid.get(that.initialModel, ["preferences"]), stats);
+        // Only save the changed preferences so the future initial model value change on unchanged
+        // preferences can still be correctly merged with changed preferences
```
jobara: I'm not sure I understand this comment.

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -371,33 +371,36 @@ var fluid_2_0 = fluid_2_0 || {};
     };
 
     /**
-     * Saves the current model and fires onSave
- */
+     * Sends the prefsEditor.model to the store and fires onSave
+     * @param that: An fluid.prefs.prefsEditor instance
```
jobara: should be "a" instead of "an"

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -371,33 +371,36 @@ var fluid_2_0 = fluid_2_0 || {};
...
+        // preferences can still be correctly merged with changed preferences
+        fluid.model.diff(modelToSave.preferences, fluid.get(initialModel, ["preferences"]), stats);
 
-            fluid.each(stats.changeMap, function (state, path) {
-                fluid.set(changedSelections, ["preferences", path], userSelections.preferences[path]);
+        if (stats.changes === 0) {
+            delete modelToSave.preferences;
+        } else {
+            fluid.each(stats.changeMap, function (state, pref) {
+                fluid.set(changedPrefs, pref, modelToSave.preferences[pref]);
```
jobara: can we ever have a case where there are fewer preferences selections in the changeMap than in the current model.preferences object? If so, should they remove the extra ones?

cindyli: changeMap generated by fluid.model.diff() only contains the changed preferences. It doesn't list unchanged preferences. Let me know if I didn't understand your question correctly.

cindyli:
@jobara, comments addressed. Ready for another look.

src/framework/preferences/js/PrefsEditor.js
jobara: Never mind, i missed that modelToSave.preferences was being replaced by changedPrefs.

jobara:
merged at 9cd0887c677fd00d28f148777a928823ec2e1416

PR #640, opened 2015-09-30
FLUID-5682: Aria for hide/show of sliding panel

jobara:
Added
* aria-controls, dynamic aria-pressed to hide/show button
* dynamic aria-expanded to panel
* aria-controls to reset button
* tests for the corresponding additions

Replaces: https://github.com/fluid-project/infusion/pull/619

https://issues.fluidproject.org/browse/FLUID-5682


src/components/slidingPanel/js/SlidingPanel.js
```diff
@@ -65,6 +73,26 @@ var fluid_2_0 = fluid_2_0 || {};
...
+                    funcName: "fluid.allocateSimpleId",
+                    args: "{that}.dom.panel"
+                }
+            }
+        },
+        model: {
+            isShowing: false
+        },
+        modelListeners: {
+            "{that}.model.isShowing": {
```
cindyli: Can be simplified to use ```"isShowing"``` directly.

src/components/slidingPanel/js/SlidingPanel.js
```diff
@@ -45,6 +45,14 @@ var fluid_2_0 = fluid_2_0 || {};
                 listener: "{that}.applier.modelChanged.addListener",
                 args: ["isShowing", "{that}.refreshView"]
             },
+            "onCreate.setAriaControls":{
```
cindyli: According to the ARIA example: http://oaa-accessibility.org/example/20/, not only setting ```role="button"``` for the toggle button, also the sliding panel should have ```role="region"``` set. The example also suggests to set ```aria-labelledby``` on the panel. We probably should consider to add a component option for the panel label.

jobara: Talked to "clown" in the channel. He suggested a role of group.
https://botbot.me/freenode/fluid-work/2015-10-01/?msg=50937400&page=1

src/components/slidingPanel/js/SlidingPanel.js
```diff
@@ -112,4 +139,22 @@ var fluid_2_0 = fluid_2_0 || {};
         that.events[that.model.isShowing ? "onPanelShow" : "onPanelHide"].fire();
     };
 
+    // panelId is passed in to ensure that it is evaluated before this
+    // function is called.
+    fluid.slidingPanel.setAriaProperties = function (that, panelId) {
+        that.locate("toggleButton").attr({
```
amb26: This is a very ancient discussion, but I wonder if we ever came to a definitive conclusion about whether we should be using "attr" or "prop" for setting ARIA...... "properties" .... I imagine we did, but I don't seem to find a writeup anywhere. My sense is that it simply doesn't matter, since this presumably is one of that majority of properties that are "in 1-1 correspondence with attributes" but it would be good to be clear about this - especially since the name of this function is "setAriaProperties". It does seem that all the code I can find in trunk uses "attr" rather than "prop". We could even add a paragraph to the ILDH handbook here http://handbook.floeproject.org/index.php?title=WAI-ARIA_-_Beyond_semantic_tags about this

src/framework/preferences/messages/prefsEditor.json
```diff
@@ -1,4 +1,5 @@
 {
     "slidingPanelShowText": "+ Show Display Preferences",
-    "slidingPanelHideText": "- Hide"
+    "slidingPanelHideText": "- Hide",
+    "slidingPanelPanelLabel": "User Prefrences"
```
amb26: Typo here Prefrences

src/components/slidingPanel/js/SlidingPanel.js
michelled: @amb26 I suspect that this is a case of habit rather than a thought out choice. It seems to me that as a general practice we should lean towards using "prop". The only reason I see to stick with "attr" is that it's all over the code base. What do you think? Should I change this use?

klown: @michelled @amb26 
It depends on your point of view, and what you want to accentuate.  For example, if you're thinking of it as a DOM manipulation, then attr() is fine.  All aria-* are attributes (so is 'role').   But, from the point of view of ARIA, you're setting/getting a state or a property.  Yes, ARIA makes a distinction between states and properties [1], where, to a first approximation, states are subject to change, whereas properties are less so.  In that case, ariaProp() and ariaState() might make more sense (and role()?).  Then again, maybe that's overkill.
[1] http://www.w3.org/TR/wai-aria-1.1/#statevsprop

PR #641, opened 2015-10-06
FLUID-5786: Changing SNAPSHOT to dev

jobara:
Renaming our in development pre-release identifier to dev.

Also performed listing.

https://issues.fluidproject.org/browse/FLUID-5786

### DIFF
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -26,8 +26,8 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 
 /* global console */
 
-var fluid_2_0 = fluid_2_0 || {};
-var fluid = fluid || fluid_2_0;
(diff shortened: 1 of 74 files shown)


src/framework/core/js/Fluid.js
```diff
@@ -32,7 +32,7 @@ var fluid = fluid || fluid_2_0;
 (function ($, fluid) {
     "use strict";
 
-    fluid.version = "Infusion 2.0-SNAPSHOT";
+    fluid.version = "Infusion 2.0-dev";
```
amb26: Shouldn't this now agree with the version in the package.json file?

colinbdclark: I think it probably should, yes. Is there any reason not to? If I remember, this is just a helpful human-readable string, right? If, so, we might as well make it consistent.

PR #648, opened 2015-11-02
FLUID-5800: Basics for fixing "baseless merge" issues

amb26:
The configuration at least no longer throws on registration, even if merged result is still incorrect. However, the removal of the cognitively lost "mergeModel" and other cruft from the mergeOneImpl loop is highly worthwhile since this loop is very hot as well as poorly understood. Also able to remove fluid.isDOMish since the new fluid.isPlainObject is powerful enough to reject markers as well as anything DOMlike even in IE9 (which is itself 1 revision below our minimum support).
These were never entered into our documentation as inherently suspicious.

### DIFF
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -1821,6 +1821,9 @@ var fluid = fluid || fluid_2_0_0;
     // it has long been disused by DataBinding itself
     fluid.model.mergeModel = function (target, source) {
         if (fluid.isPlainObject(target)) {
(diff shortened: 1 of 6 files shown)


amb26:
Ok @colinbdclark  - I've done my best with this issue for now. I've added a test which (surprisingly) demonstrated that the failure is no longer demonstrable in a live component after the previous fix, and it looks like any further improvements will either make a significant mess in the existing codebase or require the complete rewrite that the JIRA indicates is required.

PR #655, opened 2015-11-19
FLUID-5759: A bug fix for the auxiliary schema builder

cindyli:
When more than one panel names defined in the aux schema use single letters, and these panels compose a composite panel using the expression of "panels: ["x", "y", ...], the builder mistakenly sets the renderOnPrefs option to true for some panels even though all panels should always be rendered.

@jobara, can you review this pull request? Thanks.

### DIFF
--- a/src/framework/preferences/js/AuxBuilder.js
+++ b/src/framework/preferences/js/AuxBuilder.js
@@ -248,7 +248,7 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             // 2. an object that describes what panels should be always rendered,
             //    and what panels should be rendered when a preference is turned on
             // The loop below is only needed for processing the latter.
-            if (!fluid.isPrimitive(thisCompositeOptions.panels) && !fluid.isArrayable(thisCompositeOptions.panels)) {
+            if (fluid.isPlainObject(thisCompositeOptions.panels) && !fluid.isArrayable(thisCompositeOptions.panels)) {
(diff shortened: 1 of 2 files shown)


src/framework/preferences/js/AuxBuilder.js
```diff
@@ -243,9 +243,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
 
-            // panels can contain an array of always on panels, or an object
-            // describing which panels are always and which are initialized by a preference value
-            if (!fluid.isPrimitive(thisCompositeOptions.panels)) {
+            // thisCompositeOptions.panels can be in two forms:
+            // 1. an array of names of panels that should always be rendered;
+            // 2. an object that describes what panels should be always rendered,
+            //    and what panels should be rendered when a preference is turned on
+            // The loop below is only needed for processing the latter.
+            if (!fluid.isPrimitive(thisCompositeOptions.panels) && !fluid.isArrayable(thisCompositeOptions.panels)) {
```
jobara: Because we are trying to only run the code against an object, I wonder if this make more sense to be written as:

```javascript
if (fluid.isPlainObject(thisCompositeOptions.panels) && !fluid.isArrayable(thisCompositeOptions.panels)) {
```

cindyli:
@jobara, addressed the comment. Ready for another look. Thanks.

jobara:
merged at 133da8e1de28b46cd7964a946c459f212ec6ab3d

PR #656, opened 2015-11-20
FLUID-5666: Added a new function fluid.changeElementValue() into FluidDocument.js

cindyli:
It sets the value to a DOM node and fires the change event.

@amb26, can you review this pull request? Thanks.

### DIFF
--- a/src/framework/core/js/FluidDocument.js
+++ b/src/framework/core/js/FluidDocument.js
@@ -174,4 +174,10 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         };
     });
 
+    // Sets the value to the node and fires the change event
+    fluid.changeElementValue = function (node, value) {
+        node = $(node);
+        node.val(value).change();
+    };
+
 })(jQuery, fluid_2_0_0);
--- a/src/framework/core/js/FluidDocument.js
+++ b/src/framework/core/js/FluidDocument.js
@@ -174,12 +174,13 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         };
     });
 
-    /* Sets the value to the node and triggers the change event on the node.
+    /* Sets the value to the DOM element and triggers the change event on the element.
      * Note: when using jQuery val() function to change the node value, the change event would
      * not be fired automatically, it requires to be initiated by the user.
      *
-     * @param node - A jQueryable DOM element (a selector, a DOM node, or a jQuery instance)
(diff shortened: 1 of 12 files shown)


src/framework/core/js/FluidDocument.js
```diff
@@ -174,4 +174,10 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         };
     });
 
+    // Sets the value to the node and fires the change event
```
amb26: Better docs would be helpful here - up to the standard of the JSDoc-style comments that are the basis of the newer-style documentation - indicating i) the types and names of the arguments, ii) any return value, iii) what the function actually does - "fires the change event" is too vague to allow the user to interpret the meaning of this function, even though in theory it is pretty obvious

cindyli: @amb26, I've improved the comment. Let me know what you think. Thanks.

src/framework/core/js/FluidDocument.js
```diff
@@ -174,4 +174,17 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
 
+    /* Sets the value to the DOM element and triggers the change event on the element.
+     * Note: when using jQuery val() function to change the node value, the change event would
+     * not be fired automatically, it requires to be initiated by the user.
+     *
+     * @param node {A jQueryable DOM element} A selector, a DOM node, or a jQuery instance
+     * @param value {String|Number|Array} A string of text, a number, or an array of strings
+     * corresponding to the value of each matched element to set in the node
+     */
+    fluid.changeElementValue = function (node, value) {
```
amb26: Thanks - these docs look good now.
In checking over the diff, I see that there are still actually no calls to this utility outside test cases - which suggests that the proper place for it is inside our test utilities rather than the core framework. 

I suggest it should go into the file jqUnit-browser.js under the name jqUnit.changeElementValue

cindyli: @amb26, the use case that triggers the extraction of changeElementValue() (https://github.com/GPII/first-discovery/pull/47#discussion-diff-30366584) uses this function in the code that implements actual functionalities instead of code for testing, so I wonder if it makes sense to have this function in the core framework for a broader use. 

What do you think, @amb26?

amb26: Hi @cindyli - yes, sorry for my forgetfulness that I had already recommended the creation of this utility - this is fine

PR #658, opened 2015-11-27
FLUID-5730: removed qunit-fixture element

jobara:
The test markup had been inside a qunit-fixture; however, qunit was clearing out
the markup causing the test component to initialize without all of the markup it
was expecting. This was only an issue in IE. I've removed the fixture as there
is only one test that uses the markup.

https://issues.fluidproject.org/browse/FLUID-5730

### DIFF
--- a/tests/framework-tests/preferences/html/SelfVoicingEnactor-test.html
+++ b/tests/framework-tests/preferences/html/SelfVoicingEnactor-test.html
@@ -45,11 +45,9 @@
         <div id="qunit-testrunner-toolbar"></div>
         <h2 id="qunit-userAgent"></h2>
         <ol id="qunit-tests"></ol>
-        <div id="qunit-fixture">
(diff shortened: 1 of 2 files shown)


amb26:
The correct approach for an fixture using the IoC testing framework to target some markup is for the root of its markup to be registered in the "markupFixture" option of the testEnvironment - see docs at 
https://wiki.fluidproject.org/display/docs/The+IoC+Testing+Framework#TheIoCTestingFramework-AMoreComplexExample

This will direct the testing framework to ensure that the markup is correctly torn down and restored between different test runs (similar to the qunit-fixture behaviour, but supporting multiple tests targetted at multiple pieces of markup)

jobara:
@amb26 Added the markupfixture. Ready for more review.

amb26:
Thanks for this fix, @jobara - I was able to verify it in IE11

PR #660, opened 2015-12-16
FLUID-5831: Corrected broadcast of storeType so that it properly displaces default and test grades

amb26:
@acheetham - here's the fix for the "storeType" issue we discussed earlier

### DIFF
--- a/examples/framework/preferences/minimalEditor/js/prefsEditor.js
+++ b/examples/framework/preferences/minimalEditor/js/prefsEditor.js
@@ -9,7 +9,7 @@ You may obtain a copy of the ECL 2.0 License and BSD License at
 https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 */
 
-/* global fluid */
+/* global fluid, awesomeCars */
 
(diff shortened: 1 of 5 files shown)


acheetham:
@amb26, just noticing that the JIRA number in your commit log has some digits transposed...

acheetham:
@amb26, I've tried this change in my sample code - works beautifully. Specifying a custom store is very straightforward now, and seems quite reasonable, from an integrator's perspective.

jobara:
@amb26 the commit log has the wrong JIRA number. It should have been FLUID-5831. You should be able to correct this with `git commit --amend` or `git reset --soft` and then push back up with `git push origin -f`.

jobara:
@amb26 the UIO demo is broken, and it has the "gear" icon in the bottom right of the page, which i guess is for the Infusion debugger thing.

amb26:
Thanks, @jobara - I've corrected both the commit comment and the demo failure

jobara:
merged at e51c67226e2aa7a758e1a2c364d2448dbe947539

PR #661, opened 2015-12-25
FLUID-5833: Added check for "name" field in testCase to avoid anomalous behaviour issuing a QUnit module with falsy name

amb26:


cindyli:
Tried the change with my GPII-1546 branch(https://github.com/cindyli/universal/tree/GPII-1546), an upgraded GPII-1245 branch where the issue was originated, but using the latest infusion so this change can be applied. 

The change didn't fix the issue because the testCase.name variable verified in the if condition is the existing test module name (an example: https://github.com/cindyli/universal/blob/GPII-1245-odd/gpii/node_modules/gpii-oauth2/gpii-oauth2-authz-server/webTests/components/privacySettings/js/PrivacySettingsWithPrefsTests.js#L45) instead of the "name" field required in the "tests" block (an example is starting from the line 46 in the same example js file), so that, the fluid.fail() is never called.

amb26:
Thanks for testing this so carefully, @cindyli - I've updated the branch with more validation, let me know if it seems to work for you

tests/test-core/utils/js/IoCTestUtils.js
```diff
@@ -611,6 +614,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             // might enter the queue and immediately leave it as a result of only ever issuing
             // asynchronous tests
             var oldLength = QUnit.config.queue.length;
+            if (!fixture.name) {
+                fluid.fail("Error in configuration of test fixture - required field \"name\" is missing in ", fixture, " at index " + index + " of test case ", testCase);
```
cindyli: As the variable ```index``` is undefined, when hitting this line, the error is ```ReferenceError: index is not defined```.

amb26:
Oops, terribly sorry @cindyli - this is what I get for cutting and pasting this code back from the obsolete version of Infusion.... now fixed

cindyli:
Tested and confirmed this pull request does solve the issue. Thanks, @amb26.

cindyli:
Merged @ 09fac7f7ca172073cd2628e8af578e47852b4977

PR #662, opened 2016-01-14
FLUID-5832: Using Uglify to generate a source map for the concatenate…

michelled:
…d file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ The Infusion library is the cornerstone of a number of projects. You can see som
 
 ## How Do I Create an Infusion Package? ##
 
-Strictly speaking, Infusion can be used directly from source (i.e. by including each individual required file). However, for simplicity and performance reasons, you may wish to create a concatenated, minified file. The Grunt build options described below will also allow you to remove any unneeded features or libraries that you may already have in your project.
+Strictly speaking, Infusion can be used directly from source (i.e. by including each individual required file). However, for simplicity and performance reasons, you may wish to create a concatenated, minified file. Source maps for the minified file are automatically generated to make debugging easier. The Grunt build options described below will allow you to remove any unneeded features or libraries that you may already have in your project.
 
 ### Dependencies ###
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
   },
   "devDependencies": {
     "grunt": "~0.4.2",
-    "grunt-contrib-uglify": "~0.2.2",
+    "grunt-contrib-uglify": "~0.11.0",
     "grunt-contrib-clean": "~0.5.0",
     "grunt-contrib-copy": "~0.4.1",
     "grunt-contrib-concat": "~0.3.0",
```
jobara: It seems that concat isn't being used any more, it should be removed from the dev dependencies.

jobara:
@michelled is there anything that needs to be done to make use of the source maps? If so we should document that somewhere in the README.

michelled:
@jobara there is nothing special that needs to happen to use the source maps. In the FF dev tools debugger, there is a flag called 'Show Original Sources' that you can toggle to see individual unminified files instead of the concatenated minified file. Do you think I should put this into the readme?

jobara:
@michelled probably worth adding something about it. Even documenting in the readme that we provide the source maps is useful.

michelled:
@jobara I finally updated the README :)

README.md
```diff
@@ -36,7 +36,7 @@ The Infusion library is the cornerstone of a number of projects. You can see som
 
 ## How Do I Create an Infusion Package? ##
 
-Strictly speaking, Infusion can be used directly from source (i.e. by including each individual required file). However, for simplicity and performance reasons, you may wish to create a concatenated, minified file. The Grunt build options described below will also allow you to remove any unneeded features or libraries that you may already have in your project.
+Strictly speaking, Infusion can be used directly from source (i.e. by including each individual required file). However, for simplicity and performance reasons, you may wish to create a concatenated, minified file. Source maps for the minified file are automatically generated to make debugging easier. The Grunt build options described below will allow you to remove any unneeded features or libraries that you may already have in your project.
```
amb26: Could you add a little more to give the user some idea how to use source maps? Perhaps i) some pointers to the relevant docs for at least FF, Chrome - and, presumably, the user has to arrange for the built file to be hosted from the place it was generated by the build in a statically hosted image of the whole project?

jobara:
@michelled running a build with the `--source` flag set doesn't seem to produce an infusion-all.js file. I ran this command "grunt --source=true".

jobara:
@michelled could you also add source maps for the stylus generation of css files.

jobara:
Closed in favour of https://github.com/fluid-project/infusion/pull/723

PR #663, opened 2016-01-22
FLUID-2845, FLUID-5669: Improved the binaryOp model transformation

kaspermarkus:
* Now uses fluid.model.isSameValue form comparison instead of === and !==
* the logical or (||) operator now always returns either true or false as opposed to returning the first value in case of false (which javascript does).
* Unit tests for new features


amb26:
The part of the fix for FLUID-5669 looks great.
The FLUID-5845 is more puzzling - 
i) Why should the || operator return logical true or false, when this isn't the natural behaviour in JavaScript?
ii) Why should the || be fixed in this way, but not the && operator?

kaspermarkus:
I removed the FLUID-5845 related changes - will continue discussion on it in the JIRA

PR #664, opened 2016-02-01
GPII 1550: Adds changes required for automated tests

avtar:
These changes let anyone launch installed browsers and have them run Infusion tests and also run the tests in a Fedora VM. Test results are displayed in the host operating system's console in both cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,8 @@ To test a build it is advisable to serve the contents of this directory using a 
 
 Automated tests can be run by first [installing Testem](https://github.com/testem/testem/#installation) and then running ``testem ci --file tests/testem.json`` in this directory. Any browsers that Testem detects on your platform will be launched with test results returned in the [TAP](https://testanything.org/) format. You can use the ``testem launchers`` command to get a list of available browsers.
 
+A [Fedora VM](https://github.com/idi-ops/packer-fedora) can be automatically created using tools provided by the [Prosperity4All Quality Infrastructure](https://github.com/GPII/qi-development-environments/). Please ensure the [requirements](https://github.com/GPII/qi-development-environments/#requirements) have been met. The ``vagrant up`` command can then be used to provision a new VM. Typing ``grunt tests`` will run the Infusion tests using Testem in the VM and report the results in your console.
+
 ### Modules ###
 
 #### Framework Modules ####
--- a/README.md
+++ b/README.md
@@ -105,6 +105,12 @@ By default, custom packages are given a name with the form _infusion-custom-<ver
 
     grunt custom --name="myPackage"    # this produces infusion-myPackage.js
 
+### How Do I Run Tests? ###
+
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -105,6 +105,14 @@ By default, custom packages are given a name with the form _infusion-custom-<ver
 
     grunt custom --name="myPackage"    # this produces infusion-myPackage.js
 
+### How Do I Run Tests? ###
+
+To test a build it is advisable to serve the contents of this directory using a web server and then opening [tests/all-tests.html](https://github.com/fluid-project/infusion/blob/master/tests/all-tests.html) in a web browser. The browser will need to remain in the foreground because some of the tests require window focus. 
+
+Automated tests can be run by first [installing Testem](https://github.com/testem/testem/#installation) and then running ``testem ci --file tests/testem.json`` in this directory. Any browsers that Testem detects on your platform will be launched with test results returned in the [TAP](https://testanything.org/) format. You can use the ``testem launchers`` command to get a list of available browsers.
```
colinbdclark: Somehow I feel like this paragraph and the next one could be clarified. Am I correct in understanding that there are two different ways that one could use this new Testem support?

1. You can install Testem yourself on your local machine and then run it on your host, and it will run all the tests using whichever browsers you have installed
2. You can run the tests inside a vagrant box, in which case Testem will launch all the browsers inside this Linux-based VM. Which browsers will it run?

Is that correct?

avtar: @colinbdclark yes, that's correct. I'll add a revision.

avtar:
@colinbdclark I've updated the README. Thoughts?

colinbdclark:
I think it's perfect!

colinbdclark:
One other question: is there any good way to ensure that the VM will contain the latest versions of each of the browsers? I noticed that it was running Firefox 38, while version 44 was just recently released. Does the <code>vagrant provision</code> step ensure that the latest packages are installed? Or will we have to something more to ensure relatively maintenance-free updates?

avtar:
The browsers are packaged with the IDI Fedora Vagrant box. This approach was used in an effort to reduce the amount of time taken to provision a test environment. I can add a command to the ``Vagrantfile`` that will attempt to update the browsers when a VM is first provisioned or if explicitly asked to do so using ``vagrant provision``.

colinbdclark:
I think it would be great to have a command people can run to ensure that their vagrant box has the latest browsers. Go for it!

avtar:
Please give it another shot. It might be best to ``vagrant destroy -f`` and start with a new VM.

```
Upgrading:
 firefox                               x86_64                  44.0-3.fc22                         updates                         75 M
 google-chrome-stable                  x86_64                  48.0.2564.97-1                      google-chrome                   46 M
```

colinbdclark:
<code>vagrant provision</code> was enough to update my vagrant box. This pull request is, in my mind, ready to go.

PR #668, opened 2016-02-09
FLUID-5837: The client side of the UIO and GPII integration

cindyli:
Adds a new GPII panel as well as an export button on the button bar, for the communication between UIO and GPII.

To launch the demo, access demos/uiOptions/gpii.html


src/framework/preferences/js/Builder.js
```diff
@@ -196,12 +196,22 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+    fluid.defaults("fluid.prefs.create", {
+        gradeNames: ["fluid.viewComponent"],
+        builder: {}, // options provided by users
+        prefsEditor: {}, // options provided by users
+        members: {
+            constructedGrades: "@expand:fluid.prefs.builder({that}.options.build)"
+        },
+        components: {
+            prefsEditor: {
+                type: "{fluid.prefs.create}.constructedGrades.options.assembledPrefsEditorGrade",
```
jobara: @cindyli this is an interesting approach. I wonder if you could do this with an invoker and dynamic grade instead? But i'm also not sure that that would be a better solution than what's here. @amb26 do you have any thoughts on this approach.

src/framework/preferences/js/GpiiPanel.js
```diff
@@ -0,0 +1,107 @@
...
+            auto: ".flc-prefsEditor-gpii-auto",
+            autoLabel: ".flc-prefsEditor-gpii-auto-label"
+        },
+        strings: {
+            importLabel: "{that}.msgLookup.importLabel",
+            exportLabel: "{that}.msgLookup.exportLabel",
+            importTooltipContent: "{that}.msgLookup.importTooltipContent",
+            exportTooltipContent: "{that}.msgLookup.exportTooltipContent"
+        },
+        selectorsToIgnore: ["exportButton", "importButton"],
```
jobara: Why are these ignored, shouldn't their text be localized through the renderer?

tests/all-tests.html
```diff
@@ -72,13 +52,7 @@
...
         "./component-tests/tooltip/html/Tooltip-test.html",
         "./component-tests/uiOptions/html/UIOptions-test.html",
-        "./component-tests/uploader/html/DemoUploadManager-test.html",
-        "./component-tests/uploader/html/ErrorPanel-test.html",
-        "./component-tests/uploader/html/FileQueue-test.html",
-        "./component-tests/uploader/html/FileQueueView-test.html",
-        "./component-tests/uploader/html/HTML5UploaderSupport-test.html",
-        "./component-tests/uploader/html/Uploader-test.html",
-        "./component-tests/uploader/html/UploaderCompatibility-test.html",
+        "./component-tests/uploader/all-tests.html",
```
jobara: Does this work in all browsers. I feel like I had issues before with the nesting of iframes in iframes.

src/framework/preferences/js/Builder.js
amb26: This is a better approach than invoker + dynamic grade - as our docs suggest, "raw dynamic grades" are a last-ditch technique. 
http://docs.fluidproject.org/infusion/development/ComponentGrades.html#raw-dynamic-grades

tests/all-tests.html
amb26: Yes, I would prefer we didn't do this, just on general principles. If we did, we would at least need to make sure to validate it with @avtar 's integration server (at least, that it didn't result in more test failures than we already have). The cost of cut'n paste maintenance here I think is small compared to the risks, or at least our current apprehension of them

src/framework/preferences/js/Builder.js
amb26: Sorry I am a bit below par today. The approach itself is fine, but we might want to worry about the fact that the returned component from this grade does not do anything useful until the user looks at the "prefsEditor" inside it. Do we have any examples in our documentation of expecting the user to be able to do anything with the component directly?

cindyli: Our documentation of how to use ```fluid.prefs.create``` to construct preferences editor:
http://docs.fluidproject.org/infusion/development/PreferencesEditor.html

```fluid.prefs.create``` is primarily used by users to provide options that controls the assembling of the preferences editor. For its returned grade, users can also find some builder information such as grade names of assembled prefsEditor or uiEnhancer. But I doubt if users would care about those things. :)

cindyli:
@jobara, @amb26, thanks for the reviews. all addressed and ready for another look.

PR #670, opened 2016-02-12
FLUID-5821: Fix for DOM Binder "localContainer" problem causing Reorderer mouse interaction failure

amb26:


amb26:
@jobara - here is a fix for the grid reorderer demo issue we discovered on Wednesday

jobara:
@amb26 I'm running into a bug in IE 11 Win 8.1.

Steps to reproduce:

1. Open the grid reorderer demo

2. Tab to the grid

3. Move selection to the letter 'D'

4. Using ctrl-arrow keys move the 'D' around the 4 spots in the lower right corner.

5. Using the ctrl-arrow keys move the 'D' in the opposite direction around the 4 spots in the lower right corner.

Notice that after, up to a few cycles, that the the letter in the lower right corner disappears. If you perform these actions while the developer tools are open, the browser just crashes and reloads.

jobara:
Screen shot of the letter in the lower right corner disappearing.

screen shot 2016-02-17 at 10 55 38 am

jobara:
Screen shot of the browser crashing when the developer tools are open.

screen shot 2016-02-17 at 10 56 08 am

amb26:
Hi @jobara - I tested this with IE11 in Windows 7 and couldn't reproduce either of these problems. The fact that you can get a crash with the dev tools suggests that there may be a problem with your installation of it - would it be possible to try in a different VM or with different updates?
I am at IE11 version 11.0.9600.16428

colinbdclark:
I experience the same issue as @jobara on a Windows 10 VMWare VM with IE11, including the browser crashing with the dev tools open. Perhaps even more interestingly, Edge is also exhibits the same behaviour.

colinbdclark:
@jhung has offered to test this on a native Windows host to try to take a few variables out of the equation.

jhung:
@amb26 I tested on a Surface Pro 3 on Windows 10 with IE11 and was able to reproduce the problem in both the FLUID-5821 branch and on the build site. So it doesn't seem to be a VM specific issue.

amb26:
Hi - I just tested on a distinct, and native IE11 machine (my laptop with IE also verison 11.0.9600.16428) and still can't reproduce it. Given that @jhung can reproduce this issue on the trunk-based build site I suggest we merge this branch in since it seems it has not caused a regression and is holding up work on other issues since it causes our demos to be broken on all platforms (see #672).

We should make a separate JIRA for this issue and pursue it. From what I can see, the only remaining variable is the OS - all of my tests have been on Windows 7 and all of the failures have been on Windows 10. I'll take a look on my other laptop now to try and reproduce

colinbdclark:
Okay, I think that is a reasonable approach. @jobara or @amb26, who has time to create a JIRA to cover this issue?

amb26:
Just tested on Edge on Windows 10 - I can't reproduce there either, but I do notice what seems to be a separate issue, than when you try to drag the active element "out of the square", the entire page will refresh rather than a noop on other platforms.

amb26:
I've written this up as https://issues.fluidproject.org/browse/FLUID-5870

colinbdclark:
Thanks @amb26, hugely appreciated. I'm about to merge this.

PR #671, opened 2016-02-14
FLUID-5695, etc.: Numerous improvements to ChangeApplier and model transforms

amb26:
FLUID-5695: Allow "invalidation-style" model listeners to listen to multiple paths
FLUID-5848: Detect indirect references to model components from model listeners and relay rules

### DIFF
--- a/src/framework/core/js/DataBinding.js
+++ b/src/framework/core/js/DataBinding.js
@@ -356,7 +356,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
                 });
             }
             var shadow = fluid.shadowForComponent(that);
-            shadow.modelComplete = true; // technically this is a little early, but this flag is only read in fluid.connectModelRelay
+            if (shadow) { // Fix for FLUID-5869 - the component may have been destroyed during its own init transaction
+                shadow.modelComplete = true; // technically this is a little early, but this flag is only read in fluid.connectModelRelay
+            }
         });
 
         transac.commit(); // committing one representative transaction will commit them all
--- a/src/framework/core/js/DataBinding.js
+++ b/src/framework/core/js/DataBinding.js
@@ -380,23 +380,34 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         }
     };
 
-    fluid.transformToAdapter = function (transform, targetPath) {
-        var basedTransform = {};
-        basedTransform[targetPath] = transform;
-        return function (trans, newValue /*, sourceSegs, targetSegs */) {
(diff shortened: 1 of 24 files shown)


amb26:
FLUID-5490: Restore "source tracking" facility to ChangeApplier
FLUID-5294: Inconsistency in transforms using input named "value"
FLUID-5586: Implementation of "changePath" records is incomplete

amb26:
FLUID-5361: Allow modelListeners to respond to priority records (in this case constraint priorities)
FLUID-5866: "priorityHolder" component and IoC-resolved priorities for modelListeners

amb26:
FLUID-5867: Revert ancient behaviour of interpreting the interior of a transform rule as an output-structured document
FLUID-5865: Working implementation of PRIORITISED MODEL RELAY

amb26:
FLUID-5869: Fix for two causes of failure when destroying a component during its own init transaction

amb26:
FLUID-5847: Diagnostics for case of relay rule with both "source" and model dependency, improved diagnostics in malformed IoC configuration

@colinbdclark - this one is now ready to go!

amb26:
Work is in progress on a matching docs pull at https://github.com/fluid-project/infusion-docs/pull/84

src/framework/core/js/DataBinding.js
```diff
@@ -374,56 +377,94 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
-    };
-
     fluid.parseModelReference = function (that, ref) {
         var parsed = fluid.parseContextReference(ref);
         parsed.segs = that.applier.parseEL(parsed.path);
         return parsed;
     };
 
-    fluid.parseValidModelReference = function (that, name, ref) {
+    /** Given a string which may represent a reference into a model, parses it into a structure holding the coordinates for resolving the reference. It specially
```
colinbdclark: Robust documentation, even for internal functions like this, makes the review process so much easier. Thank you for including this!

colinbdclark:
I have reviewed this pull request and I think it's great.

However, bummer, the FluidPromise tests are failing in Internet Explorer 11 with the following error:

FluidPromises-test.html
SCRIPT5007: Object.freeze: argument is not an Object
Fluid.js (814,9)

It looks like IE throws an error when trying to freeze non-Object types, whereas Firefox, Chrome, and Safari are all fine with it. Is it possible that somewhere we're trying to freeze <code>undefined</code> or some other primitive type?

Tests pass in Edge and all other browsers on Mac and Windows. Won't it be great when we eventually pull the plug on IE entirely?

amb26:
Cheers, @colinbdclark - thanks for catching this. Now updated with a guard condition just for IE11!

PR #676, opened 2016-03-04
FLUID-5868: Allow the localization of message texts used at the preferences editor level

cindyli:
This pull request is to support the localization of message texts used at the preferences editor level (outside of the preferences panels), such as the "reset" button text in the separated panel and "save", "cancel" button texts in the full page versions.


jobara:
@cindyli from the pattern i see in the code and from our discussion it seems that we have multiple means of localizing text depending on which component is handling it. I'm worried if this will add to the complexity and cause confusion for integrators and for future maintenance. It also may be the case that the prefsEditorLoader is taking on too much responsibility as it's doing more than just loading the templates and message bundles. In addition to this, we have to concern ourselves with controls such as "reset" being rendered in different parts of the UI depending on which variant of prefs editor is being used (e.g. separated panel, full page). Is there a way to simplify things to have a more common approach to localizing text that would make it straightforward to know where to modify configuration as an integrator, and to also be able to add/remove ui elements that need to be localized. Perhaps this is already the best approach, and/or maybe some of these questions really stem from larger issues with the structure of the prefs framework in general. @amb26 do you have thoughts on this pull request and the prefs framework itself?

cindyli:
Closed since a more proper solution is covered by https://issues.fluidproject.org/browse/FLUID-5878

PR #677, opened 2016-03-05
FLUID-5689: Pager demo tooltips out of sync with table

dinukadesilva:
Fixed the tooltip to sync with model changes


Dinuka2013513:
Hi @jobara ,

According to the current implementation, it doesn't show the tool-tip if it's the current page. Even if that is the feature how it should be, it behaves like that only at the time. after switching to other pages, it's not working like that.

But, I feel like it's nice to have the tool-tip even when it's the current page or not. And I made that change also along with this. 

Appreciate to know what you think about this.

jobara:
@Dinuka2013513 from story card #6 ( https://wiki.fluidproject.org/display/Infusion13/Pager+Storycards ) it seems like it should have the tooltip for the current page link even on page load. @dayotte what do you think?

dayotte:
@Dinuka2013513 @jobara I agree

Dinuka2013513:
@jobara @dayotte that sounds good................ and I have also done that.
I'l update test cases also accordingly.

[a small suggestion]
I think it's nice to have a trigger to show test results as well in pull requests.

jobara:
@Dinuka2013513 I pinged you about this in another pull request as well. Could you please fill out the Fluid CLA and send it to me. https://wiki.fluidproject.org/display/fluid/Fluid+Licensing You can ping me in the IRC channel (Justin_o) if you have more questions on that.

For background reading, here's a good overview of what CLAs are and why they're important: http://oss-watch.ac.uk/resources/cla

src/components/pager/js/Pager.js
```diff
@@ -493,7 +493,7 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             }
         }],
         modelListeners: {
-            "": "{that}.events.onModelChange.fire({change}.value, {change}.oldValue, {that})"
+            "{fluid.pager}.model": "{that}.events.onModelChange.fire({change}.value, {change}.oldValue, {that})"
```
jobara: This change doesn't look like it's needed. In fact it looks like it will produce the same outcome as the we are currently in the `fluid.pager` component. (See: http://docs.fluidproject.org/infusion/development/ChangeApplierAPI.html#model-path-references)

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -182,11 +194,8 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
                     jQueryTrigger: "focus"
                 });
 
-                if (link.hasClass(pager.pagerBar.options.styles.currentPage)) { // assumption that this will not change between now and then!
-                    assertVisibleTips("There shouldn't be any tooltips visible when the currentPage is focused", []);
-                } else {
-                    assertVisibleTips("Only the tooltip for page link " + (idx + 1) + ", in the " + location + " page bar is visible", [linkId]);
-                }
+                
+                assertVisibleTips("There should be tooltip visible when none of the pageLinks are focused", [linkId]);
```
jobara: I'm a bit confused about what this is testing. Why is there a tooltip if none of the pageLinks are focused? Is something hovered?

jobara:
Please run the `grunt lint` from the project root. It appears that there are linting errors.

Dinuka2013513:
@jobara  I fixed them. Can you look in again.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -193,9 +193,7 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
                     element: link,
                     jQueryTrigger: "focus"
                 });
-
-                
-                assertVisibleTips("There should be tooltip visible when none of the pageLinks are focused", [linkId]);
+                assertVisibleTips("The tooltip for page link " + (idx + 1) + ", in the " + location + " page bar is visible", [linkId]);
```
jobara: Could you move the `idx + 1` to an appropriately named variable. Looking at this line is a bit hard to understand because it's not immediately apparent what the `idx +1` is for and the overloaded use of `+` for addition and string concatenation.

Dinuka2013513: Fixed

Dinuka2013513:
@jobara I just updated the code. Could you please look in again

cindyli:
The test case you changed is to cover the fact the focus on the current page number also triggers the display of its tooltip.

Another fact that needs to be tested is to demonstrate the issue described in [the bug report](https://issues.fluidproject.org/browse/FLUID-5689) has been fixed. That is, when the page size or the order of the displayed rows are changed, the tooltip content for the current page number is adjusted accordingly.

cindyli:
@Dinuka2013513, do you plan to continue working on this pull request? If not, this pull request will be closed due to its inactivity.

Dinuka2013513:
@cindyli , I'm working on writing test cases for bug reported. I'l update it soon.

Dinuka2013513:
@cindyli @jobara ,

I just added test cases to cover the fix related to this bug fix. Could you please look into this again.

Thanks.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,59 +159,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 
-        var tooltipTest = function (location) {
-            return function (idx, linkEl) {
-                var linkId = linkEl.id;
-                var link = $(linkEl);
+        var tooltipTest = function (location, index, expectedTooltipContents) {
+            var link =  "#rendered-ioc .flc-pager-" + location + " .flc-pager-links .flc-pager-pageLink a:eq(" + index + ")";
+            var linkNumber = index + 1;
+            var linkId = "page-link:link" + linkNumber + (location === "bottom" ? "-1" : "");
+            sequence.push({
```
cindyli: I found it's difficult to follow all these dynamically generated test sequence blocks. It defeats the purpose of having well outlined test sequences with IoC testing framework - http://docs.fluidproject.org/infusion/development/IoCTestingFramework.html

In your case that does not require to hook up event listeners or model listeners on the fly, like what line 184-191 does, as well as not having asynchronous calls, IoC testing framework might not be necessary. Consider to use jqunit.test() directly.

Also, instead of adding to existing fluid.tests.tooltipModuleSource() function, consider using a separate test as you were meant to cover another use case.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,59 +159,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
                     });
                 }
             });
         };
 
-        var tooltipTest = function (location) {
-            return function (idx, linkEl) {
-                var linkId = linkEl.id;
-                var link = $(linkEl);
+        var tooltipTest = function (location, index, expectedTooltipContents) {
```
cindyli: Use namespaced public functions whenever possible, such as line 161 does. It would help to keep the code open and more accessible. Thanks.

tests/component-tests/pager/js/PagedTableTests.js
Dinuka2013513: Actually this can be reused for verifying tool-tips when the page size has changed. Thats why I updated it accordingly. Wouldn't that be a good idea....

tests/component-tests/pager/js/PagedTableTests.js
Dinuka2013513: yeah.... that's true @cindyli

tests/component-tests/pager/js/PagedTableTests.js
cindyli: BTW, an example of how to define the test sequence with IoC testing framework: https://github.com/fluid-project/infusion/blob/master/tests/framework-tests/preferences/js/PanelsTests.js#L1066-L1093

You will find more examples by searching infusion repo with the keyword "fluid.test.testCaseHolder".

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,59 +159,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+                jQueryTrigger: "blur"
+            });
+            assertVisibleTips("There shouldn't be any tooltips visible when none of the pageLinks are focused", [], expectedTooltipContents, index);
+        };
+
+        /* 
+         * Change the page size if it has provided and
+         * verify whether the displaying links and their tooltips are synced are with page size
+         */
+        var verifyToolTips = function (pageSize) {
```
cindyli: This function verifies the collaboration among page size selection and the number of links on the top and bottom page bars. This would be a good integration test addition to Pager tests at PagerTests.js.

Regarding adding the unit test for the paged table, the change on the page size can be done via for example `pager.applier.change("pageSize", 2)` rather than going thru the DOM change.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,59 +159,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
         var sequence = [];
 
-        var assertVisibleTips = function (message, targetIds) {
+        var assertVisibleTips = function (message, targetIds, expectedTooltipContents, index) {
             sequence.push({
                 event: "{trackTooltips}.events.notifyFocusChange",
                 listener: function () {
                     fluid.tests.tooltip.assertVisible(message, pager, targetIds, null, function (tooltip) {
-                        jqUnit.assertNode("The contents of the tooltip should be set", tooltipContents[0], $("b", tooltip));
+                        jqUnit.assertNode("When the page size is " + pager.model.pageSize + ", The contents of the tooltip should be set", expectedTooltipContents[index], $("b", tooltip));
```
cindyli: Instead of using an unknown `pager.model.pageSize`, check on the expected page size as well.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,59 +159,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-                assertVisibleTips("There shouldn't be any tooltips visible when none of the pageLinks are focused", []);
-            };
+            } else {
+                pageSize = pager.model.pageSize;
+            }
+            
+            var expectedNumberOfLinks = Math.ceil(pager.dataModel.length / pageSize);
+            sequence.push({
+                func: function () {
+                    $("#rendered-ioc .flc-pager-page-size").val(pageSize);
```
cindyli: When accessing the DOM element `$("#rendered-ioc .flc-pager-page-size")` in the context of Pager.js, the proper way is to use `pager.locate("pageSize")`.

tests/component-tests/pager/js/PagedTableTests.js
cindyli: Running the test shows output test messages like "Expected number of nodes When the page size is 2, The contents of the tooltip should be set", which does not make much sense.

This comment also applies to other output messages.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,59 +159,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        
         return {
             name: "Pager tooltip tests",
-            tests: {
-                name: "Tooltip visibility",
-                sequence: sequence
-            }
+            tests: [
+                {
+                    name: "Tooltip visibility",
```
cindyli: Probably use "Tooltip visibility and contents".

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,59 +159,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
                 sequence.push({
-                    element: link,
-                    jQueryTrigger: "blur"
+                    element: "#rendered-ioc .flc-pager-page-size",
+                    jQueryTrigger: "change"
                 });
-                assertVisibleTips("There shouldn't be any tooltips visible when none of the pageLinks are focused", []);
-            };
+            } else {
+                pageSize = pager.model.pageSize;
```
cindyli: Is this in use? All test cases are and should be dealing with a pre-defined `pageSize`.

tests/component-tests/pager/js/PagedTableTests.js
Dinuka2013513: @cindyli

1) Here I have just shown the size of the currently selected page size. So thats why I have use `pager.model.pageSize`. I have not mentioned about an expected page size here.
  
2) This message has come from `{trackTooltips}.events.notifyFocusChange`. Which is related to tool tips. Wouldn't it better to take it to another jira sice it's a different scope.

tests/component-tests/pager/js/PagedTableTests.js
Dinuka2013513: @cindyli @jobara 

This is the use case where the test is executed without changing the page size and considering the current page size. Which mean if the pageSize parameter has not passed.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,60 +159,104 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-                    assertVisibleTips("There shouldn't be any tooltips visible when the currentPage is focused", []);
-                } else {
-                    assertVisibleTips("Only the tooltip for page link " + (idx + 1) + ", in the " + location + " page bar is visible", [linkId]);
+    /*
+     * Change the page size if it has provided and
+     * verify whether the displaying links and their tooltips are synced are with page size
+     */
+    fluid.tests.verifyToolTips = function (pager, sequence, pageSize, expectedTooltipContent) {
+        if (pageSize) {
+            sequence.push({
```
jobara: I think @cindyli mentioned this before, but I'm also having trouble following along all of the sequence generation parts. I wonder if you could make some of these more static and reuse through concatenation, grade merging, or the like. Also it isn't obvious from the function name and description that the function will actually modify the sequence you pass into it. This could cause some unintended side effects for someone using it.

Dinuka2013513:
@jobara 

Agree @jobara , So, separated each different flow into functions. And added more readable comments and function names which help others to understand easily.

Let me know whether this is fine. 

But, still I think we should not hard code or duplicate the code even though these are test cases. Let me know your thoughts.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,60 +159,113 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 
-        var tooltipTest = function (location) {
-            return function (idx, linkEl) {
-                var linkId = linkEl.id;
-                var link = $(linkEl);
-                sequence.push({
-                    element: link,
-                    jQueryTrigger: "focus"
+    fluid.tests.assertVisibleTips = function (pager, sequence, message, targetIds, expectedTooltip) {
+        sequence.push({
```
jobara: In the old implementation the modification of the sequence was all done within the scope of a larger function. However, now that you have factored out this function and added new ones they are available publicly. This is good, but we need to make sure that the external api is expected. As I mentioned before, these all modify the sequence array directly and could cause an issue for someone who isn't expecting that. I wouldn't for example expect something called assertVisibleTips to modify a sequence array. I would expect it to perform some test to make sure that the expected tooltips are visible. I'd suggest a couple things, both here and for your other functions:

- Rename the functions to indicate what they are actually doing.
- Consider not modifying the sequence in place, but creating a new one and returning it.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,60 +159,113 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 
-        pageLinksTop.each(tooltipTest("top"));
-        pageLinksBottom.each(tooltipTest("bottom"));
-        return {
-            name: "Pager tooltip tests",
-            tests: {
-                name: "Tooltip visibility",
-                sequence: sequence
+    /**
+     * verify whether the displaying links and their tooltips
```
jobara: There must be a typo here. You've used the word "whether" but you haven't indicated what is being the display links and the tooltips are being tested for.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,60 +159,113 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+     */
+    fluid.tests.verifyToolTips = function (pager, sequence, pageSize, expectedTooltips) {
+        var expectedNumberOfLinks = Math.ceil(pager.dataModel.length / pageSize);
+        sequence.push({
+            func: function () {
+                pager.locate("pageSize").val(pageSize);
+                jqUnit.assertEquals("When the page size is " + pager.model.pageSize + ", Top pageLinks",
+                    expectedNumberOfLinks,
+                    $("#rendered-ioc .flc-pager-top .flc-pager-links .flc-pager-pageLink a").length);
```
jobara: You should not use jQuery directly like this. This will limit the ability to reuse this function for another test. You should make use of the Dom Binder that the component supplies to look up any elements. In this case you'll probably need to go through the pageBar subcomponents to access the elements. So something like:

```javascript
pager.pagerBar.locate("pageLinks").length;
pager["pagerBar-1"].locate("pageLinks").length;
```

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,60 +159,113 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        }
+    };
+
+    /**
+     * Verify whether the tooltip of the given index is matched with the provided tooltip
+     */
+    fluid.tests.verifyToolTip = function (pager, sequence, location, index, expectedTooltip) {
+        var link =  "#rendered-ioc .flc-pager-" + location + " .flc-pager-links .flc-pager-pageLink a:eq(" + index + ")";
+        var linkNumber = index + 1;
+        var linkId = "page-link:link" + linkNumber + (location === "bottom" ? "-1" : "");
```
jobara: Similar to above, you should avoid constructing these jQuery selectors. Also I think you should just pass in the pageBar component you are interested in, instead of passing in the location.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,60 +159,113 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+     * Change the page size
+     */
+    fluid.tests.setPageSize = function (pager, sequence, pageSize) {
+        sequence.push({
+            func: function () {
+                pager.locate("pageSize").val(pageSize);
+            }
+        });
+        sequence.push({
+            element: "#rendered-ioc .flc-pager-page-size",
```
jobara: As mentioned elsewhere you should avoid using jQuery selectors like this, as it will reduce reusability. Instead, make use of the components DOM Binder.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,60 +159,113 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            jQueryTrigger: "change"
+        });
```
jobara: Rather than having separate sequences for this, I think you should make a general function here, because it really does represent an atomic action. Something like:

```javascript
var makeSelection = function (selectElm, newValue) {
    selectElm = $(selectElm);
    selectElm.val(newValue);
    selectElm.trigger("change");
};

Comment (amb26) [inline tests/component-tests/pager/js/PagedTableTests.js]: We already have a utility for this

https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/FluidDocument.js#L185

File: tests/component-tests/pager/js/PagedTableTests.js [hunk 18]
```diff
@@ -150,60 +159,113 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            [ {nodeName: "b", nodeText: "dog"}, {nodeName: "b", nodeText: "bird"} ],
+            [ {nodeName: "b", nodeText: "fish"}, {nodeName: "b", nodeText: "dragon"} ],
+            [ {nodeName: "b", nodeText: "ant"}, {nodeName: "b", nodeText: "ant"} ]
+        ]);
+
+        fluid.tests.setPageSize(pager, sequence, 4);
+        fluid.tests.verifyToolTips(pager, sequence, 4, [
+            [ {nodeName: "b", nodeText: "dog"}, {nodeName: "b", nodeText: "fish"} ],
+            [ {nodeName: "b", nodeText: "camel"}, {nodeName: "b", nodeText: "ant"} ]
+        ]);
```
jobara: In looking at these now, I think you actually want to direct the tests with a structure that looks like:

```javascript
fluid.tests.tooltipTestDriver  = [
    {
        pageSize: 2,
        expectedTooltips: [
            {nodeName: "b", nodeText: "dog"}, {nodeName: "b", nodeText: "cat"}
        ], [ 
            {nodeName: "b", nodeText: "bird"}, {nodeName: "b", nodeText: "fish"}
        ], [
            {nodeName: "b", nodeText: "camel"}, {nodeName: "b", nodeText: "dragon"}
        ], [
            {nodeName: "b", nodeText: "ant"}, {nodeName: "b", nodeText: "ant"}
        ]
    },
    // and so on
];
```
This structure should be in the public namespace for potential reuse. You'd pass this structure into a function that loops over the top level array, and constructs and returns the sequence.

jobara: actually i think you'd just modify `fluid.tests.tooltipModuleSource` to take in the above structure as an argument.

Dinuka2013513:
Thanks for all the support @jobara @cindyli @amb26 ,

I updated the test cases and removed the sequence structure too. Now, all these methods can be used directly. Also I separated this into some more functions to improve the readability and the maintainability.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -160,30 +160,22 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 
     fluid.tests.tooltipModuleSource = function (pager) {
         var sequence = [];
+        var linksList = pager.dataModel.map(function(data) {
+            return {text: data.animal};
+        });
```
jobara: Could you use the `fluid.transform` utility for this. http://docs.fluidproject.org/infusion/development/CoreAPI.html#fluid-transform-source-fn1-fn2-

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -160,30 +160,22 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+        /* Change the page size and verify whether the tooltips and links are synced accordingly */
+        for (var pageSize = 2; pageSize <= 4; pageSize++) {
```
jobara: maybe you can take the pageSize range as an argument in `fluid.tests.tooltipModuleSource`

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-            [
-                {nodeName: "b", nodeText: "bird"},
-                {nodeName: "b", nodeText: "fish"}
-            ]
-        ];
-
         var sequence = [];
+        var linksList = pager.dataModel.map(function(data) {
+            return {text: data.animal};
+        });
```
jobara: Could you use the `fluid.transform` utility for this. http://docs.fluidproject.org/infusion/development/CoreAPI.html#fluid-transform-source-fn1-fn2-

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 
-        var assertVisibleTips = function (message, targetIds) {
+        /* Change the page size and verify whether the tooltips and links are synced accordingly */
+        for (var pageSize = 2; pageSize <= 4; pageSize++) {
```
jobara: maybe you can take the pageSize range as an argument in `fluid.tests.tooltipModuleSource`

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+    fluid.tests.assertVisibleTips = function (pager, message, targetIds, expectedTooltip) {
+        fluid.tests.tooltip.assertVisible(message, pager, targetIds, null, function (tooltip) {
+            jqUnit.assertNode(tooltip.length, expectedTooltip, $("b", tooltip));
+        });
+    };
+
+    fluid.tests.setPageSize = function (pager, pageSize) {
+        fluid.changeElementValue(pager.locate("pageSize"), pageSize);
+    };
```
jobara: You don't need `fluid.tests.setPageSize` you can just use fluid.changeElementValue directly

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        for (var pageSize = 2; pageSize <= 4; pageSize++) {
             sequence.push({
-                event: "{trackTooltips}.events.notifyFocusChange",
-                listener: function () {
-                    fluid.tests.tooltip.assertVisible(message, pager, targetIds, null, function (tooltip) {
-                        jqUnit.assertNode("The contents of the tooltip should be set", tooltipContents[0], $("b", tooltip));
-                    });
-                }
+                func: "fluid.tests.setPageSize",
+                args: [pager, pageSize]
```
jobara: You can avoid using `fluid.tests.setPageSize` here with the following IoC expression.

```javascript
{
    func: "fluid.changeElementValue",
    args: ["pager.dom.pageSize", pageSize]
}
```

jobara: above `"pager.dom.pageSize"` in an IoC expression is equivalent to saying `pager.locate("pageSize");`

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            ]
         };
     };
 
+    fluid.tests.getTooltipContents = function(listOfData, pageSize) {
+        var tooltipContents = [];
+        for (var i = 0; i < listOfData.length; i += pageSize) {
+            var tooltipContent = [];
+            var pageStartIndex = i;
+            var pageEndIndex = (i + pageSize -1) < listOfData.length ? (i + pageSize -1) : (listOfData.length - 1);
```
jobara: typo should be `(i + pageSize - 1)

jobara: I'd also suggest you capture this in a variable instead of performing the operation twice.

jobara: It seems like you are just looking for the smaller of the two values, so you might be able to simplify this by using `Math.min`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/min

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+        return tooltipContents;
+    };
+
+    fluid.tests.verifyLinksAndTooltips = function(pager, pageSize, linksList) {
+        fluid.tests.setPageSize(pager, pageSize);
+        fluid.tests.verifyPagerBar(pager, pageSize, "top");
+        fluid.tests.verifyPagerBar(pager, pageSize, "bottom");
+        fluid.tests.verifyToolTipsOfPagerBar(pager, pageSize, "top", fluid.tests.getTooltipContents(linksList, pageSize));
+        fluid.tests.verifyToolTipsOfPagerBar(pager, 4, "bottom", fluid.tests.getTooltipContents(linksList, pageSize));
```
jobara: where does "4" come from?

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        fluid.changeElementValue(pager.locate("pageSize"), pageSize);
+    };
+
+    fluid.tests.verifyPagerBar = function (pager, pageSize, location) {
+        var expectedNumberOfLinks = Math.ceil(pager.dataModel.length / pageSize);
+
+        //Verify the number of links in the pager bar
+        jqUnit.assertEquals("The number of links in the pager should be according to the page size",
+            expectedNumberOfLinks,
+            pager[location === "top" ? "pagerBar" : "pagerBar-1"].locate("pageLinks").length);
```
jobara: You only use location to determine which pagerBar to use. I think it would be better to just pass in the pagerBar name or the component itself.

jobara: Actually an argument structure of (pagerBar, dataLength, pageSize) would make more sense, or even just (pagerBar, numExpectedLinks)

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+    fluid.tests.verifyToolTipsOfPagerBar = function (pager, pageSize, location, expectedTooltips) {
+        var expectedNumberOfLinks = Math.ceil(pager.dataModel.length / pageSize);
+
+        for (var x = 0; x < expectedNumberOfLinks; x++) {
+            fluid.tests.verifyPagerToolTip(pager, location, x, expectedTooltips[x]);
+        }
+    };
+
+    fluid.tests.verifyPagerToolTip = function (pager, location, index, expectedTooltip) {
+        var link =  $(pager[location === "top" ? "pagerBar" : "pagerBar-1"].locate("pageLinks").find("a")[index]);
```
jobara: Again try to factor out needing the location.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -150,62 +159,99 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        var linkId = "page-link:link" + linkNumber + (location === "bottom" ? "-1" : "");
+
+        link.trigger("focus");
+        fluid.tests.assertVisibleTips(pager,
+            "The tooltip of page link " + linkNumber + ", in " + location + " page bar is visible",
+            [linkId], expectedTooltip);
+
+        link.trigger("blur");
+        fluid.tests.assertVisibleTips(pager,
+            "There shouldn't have any tooltips visible when none of the pageLinks are focused",
```
jobara: Reword this as "There should not be any tooltips visible when no pageLinks are focused."

tests/component-tests/pager/js/PagedTableTests.js
Dinuka2013513: The `location` has been used to identify classes and messages as well. So, we cannot move it. But, that also can be generalized.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -195,77 +188,69 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-                        $("b", tooltip));
-                });
-            }
-        });
+    fluid.tests.getTooltipContents = function(listOfData, pageSize) {
+        var tooltipContents = [];
+        for (var i = 0; i < listOfData.length; i += pageSize) {
+            var tooltipContent = [];
+            var pageStartIndex = i;
+            var pageEndIndex = Math.min(pageStartIndex + pageSize -1, listOfData.length - 1);
```
jobara: missing a space between "-" and "1"

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -149,63 +158,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            tooltipContent.push({nodeName: "b", nodeText: listOfData[pageStartIndex].text});
+            tooltipContent.push({nodeName: "b", nodeText: listOfData[pageEndIndex].text});
+            tooltipContents.push(tooltipContent);
+        }
+
+        return tooltipContents;
+    };
+
+    fluid.tests.verifyLinksAndTooltips = function(pager, pageSize, linksList) {
```
jobara: whitespace issue. please make sure you define functions like `some.namesapce = function (arg1, arg2) {};` _**Note:** the space between the "function" and "("_  Please make sure to fix this up throughout.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -149,63 +158,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        fluid.changeElementValue(pager.locate("pageSize"), pageSize);
+        fluid.tests.verifyPagerBar(pager, "top", pageSize);
+        fluid.tests.verifyPagerBar(pager, "top", pageSize);
```
jobara: i think you meant to test "bottom"

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -149,63 +158,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+    fluid.tests.verifyToolTipsOfPagerBar = function (pager, location, pageSize, expectedTooltips) {
+        var expectedNumberOfLinks = Math.ceil(pager.dataModel.length / pageSize);
+
+        for (var x = 0; x < expectedNumberOfLinks; x++) {
+            fluid.tests.verifyPagerToolTip(pager, location, x, expectedTooltips[x]);
+        }
+    };
+
+    fluid.tests.verifyPagerToolTip = function (pager, location, index, expectedTooltip) {
+        var link =  $(fluid.tests.getPagerBar(pager, location).locate("pageLinks").find("a")[index]);
```
jobara: instead of using `[index]` you can use `.eq(index)` which will give you back a jQuery object. Also there is a lot of chaining here. Maybe separate out the pageBar from the link into separate variables.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -222,7 +269,7 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
                 options: {
                     moduleSource: {
                         func: "fluid.tests.tooltipModuleSource",
-                        args: "{pager}"
+                        args: ["{pager}", [2, 3, 4]]
```
jobara: Please pull out the array into a public variable that is named appropriately so that it is clear what it is.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -149,63 +158,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            tooltipContent.push({nodeName: "b", nodeText: listOfData[pageStartIndex].text});
+            tooltipContent.push({nodeName: "b", nodeText: listOfData[pageEndIndex].text});
+            tooltipContents.push(tooltipContent);
+        }
+
+        return tooltipContents;
+    };
+
+    fluid.tests.verifyLinksAndTooltips = function(pager, pageSize, linksList) {
+        fluid.changeElementValue(pager.locate("pageSize"), pageSize);
```
cindyli: A duplicate call? Hasn't it been performed at line 171?

tests/component-tests/pager/js/PagedTableTests.js
Dinuka2013513: @jobara since this parameter has been defined in the function with a meaningful name, do you think that it's required to move this to another external variable. I mean this is just a test data passed to the function. So, I think it would be complected if we separate this parameter in all the time we use this.

jobara: If we break it out into a variable it will be easier to know what it is just by reading the code, and it allows us to use it in other places if need be.

Dinuka2013513: But, there cannot be any instance where a sequence of page sizes is reusable. I mean it's a test data.

jobara: This discussion was picked up in the [IRC channel](https://botbot.me/freenode/fluid-work/2016-05-31/?msg=67025052&page=1).

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -195,77 +188,69 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-            element: "#rendered-ioc .flc-pager-page-size",
-            jQueryTrigger: "change"
-        });
+    fluid.tests.verifyLinksAndTooltips = function (pager, pageSize, linksList) {
+        fluid.tests.verifyPagerBar(pager, "top", pageSize);
+        fluid.tests.verifyPagerBar(pager, "bottom", pageSize);
+        fluid.tests.verifyToolTipsOfPagerBar(pager, "top", pageSize,
+            fluid.tests.getTooltipContents(linksList, pageSize));
+        fluid.tests.verifyToolTipsOfPagerBar(pager, "bottom", pageSize,
+            fluid.tests.getTooltipContents(linksList, pageSize));
```
jobara: fluid.tests.getTooltipContents seems to be called twice, you should just store the result as a variable and use that as an argument to fluid.tests.verifyToolTipsOfPagerBar

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -195,77 +188,69 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
         }
     };
 
-    /**
-     * Verify whether the tooltip of the given index is matched with the provided tooltip
-     */
-    fluid.tests.verifyToolTip = function (pager, sequence, location, index, expectedTooltip) {
-        var link =  "#rendered-ioc .flc-pager-" + location + " .flc-pager-links .flc-pager-pageLink a:eq(" + index + ")";
+    fluid.tests.verifyPagerToolTip = function (pager, location, index, expectedTooltip) {
+        var link =  $(fluid.tests.getPagerBar(pager, location).locate("pageLinks").find("a").eq(index));
```
jobara: you don't need to wrap this in a jQuery `$()` it's already a jQuery element because of using `eq(index)`

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -282,9 +267,10 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
             fixtures: {
                 type: "fluid.test.testCaseHolder",
                 options: {
```
jobara: This should all be in a separate grade that uses `fluid.test.testCaseHolder` as a gradeName and then passed in here as the type.

dinukadesilva: is it needed to move all the methods also into that.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -149,63 +158,101 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-
+    fluid.tests.tooltipModuleSource = function (pager, sequencesOfPageSizes) {
         var sequence = [];
+        var linksList = fluid.transform(pager.dataModel, function (data) {
+            return {text: data.animal};
+        });
 
-        var assertVisibleTips = function (message, targetIds) {
+        /* Change the page size and verify whether the tooltips and links are synced accordingly */
+        for (var i = 0; i < sequencesOfPageSizes.length; i++) {
```
jobara: You should be able to use `fluid.transform

tests/component-tests/pager/js/PagedTableTests.js
jobara: I was referring to the configuration specified in this options block. You can pull all of that out into a separate grade and just set the type to the new grade.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -149,63 +158,100 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-
+    fluid.tests.tooltipModuleSource = function (pager, sequencesOfPageSizes) {
         var sequence = [];
+        var linksList = fluid.transform(pager.dataModel, function (data) {
+            return {text: data.animal};
+        });
 
-        var assertVisibleTips = function (message, targetIds) {
+        /* Change the page size and verify whether the tooltips and links are synced accordingly */
+        fluid.transform(sequencesOfPageSizes, function (pageSize) {
```
jobara: Instead of `fluid.transform`, use `fluid.each`. `fluid.each` is for iteration over an object or array. `fluid.transform` creates a new array or object by performing a function over each member.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -158,24 +158,67 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+                type: "fluid.tests.renderedPager",
+                container: "#rendered-ioc",
+                options: {
+                    gradeNames: ["fluid.tests.tooltip.trackTooltips", "fluid.tests.focusNotifier"]
+                }
+            },
+            fixtures: {
+                type: "fluid.tests.pagerTooltipTests",
+                options: {
+                    sequenceOfPageSizes: [2, 3, 4]
```
jobara: You have a default value of `[2, 3]` in the grade and then override it here. Is there a specific reason to do this, and not just put `[2, 3, 4]` in the grade definition?

dinukadesilva: Actually.... the default value is just an example.....

dinukadesilva: But, I think it make sense to make it empty since there can be errors when the default value has not been overridden.

tests/component-tests/pager/js/PagedTableTests.js
```diff
@@ -218,17 +170,123 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+        //Verify the number of links in the pager bar
+        jqUnit.assertEquals("The number of links in the pager should be according to the page size",
+            expectedNumberOfLinks,
+            fluid.tests.getPagerBar(pager, location).locate("pageLinks").length);
+    };
+
+    fluid.tests.assertVisibleTips = function (pager, message, targetIds, expectedTooltip) {
+        fluid.tests.tooltip.assertVisible(message, pager, targetIds, null, function (tooltip) {
+            jqUnit.assertNode(tooltip.length, expectedTooltip, $("b", tooltip));
```
jobara: Need to have a proper message here, not just `tooltip.length` so that we know what is being tested and can easily identify what is going wrong.

dinukadesilva: @jobara the message passed to this is used to show the number of nodes as well. 
"Expected number of nodes "

jobara: What I'm seeing when I run these tests is something like the following:

> Expected number of nodes 1 - at sequence position 4 of 15
> 1: node 0: - node name expected value: b actual: b - at sequence position 4 of 15
> 1: node 0: - attribute nodeText expected value: dog actual: dog - at sequence position 4 of 15
> 1: node 1: - node name expected value: b actual: b - at sequence position 4 of 15
> 1: node 1: - attribute nodeText expected value: cat actual: cat - at sequence position 4 of 15

It's the "1" at the front that is particularly confusing. I don't know the I should equate this with a value from the preceding test. It's also confusing that if there are supposed to be only 1 node, why are two nodes tested. A real message should be able to explain what's being tested here.

dinukadesilva: @jobara there are two nodes displaying per tool-tip
They are the starting and ending items of the selected page.

jobara: @dinukadesilva why is the first test "Expected number of nodes 1 - at sequence position 4 of 15" shouldn't the expected number of nodes be two then?

dinukadesilva: There is only one node. That is the tool-tip.
And there are another two nodes inside that node. They are starting and ending items of the page

jobara: @dinukadesilva okay thanks for the explanation, but you can see how that is unclear from test message. We should be able to just read the test output and have a good understanding of what is going on.

dinukadesilva: That's true...... we need to have a look to improve this.

tests/test-core/jqUnit/js/jqUnit-browser.js
```diff
@@ -134,8 +134,8 @@ var jqUnit = jqUnit || {};
                 node = node[0];
             }
             else if (node.length !== undefined) {
-                jqUnit.assertEquals("Expected number of nodes " + message, expected.length, node.length);
-                for (var i = 0; i < node.length; ++i) {
+                jqUnit.assertEquals("Expected number of nodes ", expected.length, node.length);
+                for (var i = 0; i < node.length; ++ i) {
```
jobara: We tend to write it as `++i` as opposed to `++ i`. Please revert this change.

tests/test-core/jqUnit/js/jqUnit-browser.js
```diff
@@ -134,8 +134,8 @@ var jqUnit = jqUnit || {};
                 node = node[0];
             }
             else if (node.length !== undefined) {
-                jqUnit.assertEquals("Expected number of nodes " + message, expected.length, node.length);
-                for (var i = 0; i < node.length; ++i) {
+                jqUnit.assertEquals("Expected number of nodes ", expected.length, node.length);
```
jobara: Rather than remove the message entirely, i feel that it is useful for grouping the set of assertions performed here. Perhaps you could rewrite it as:

```javascript
jqUnit.assertEquals(message + ": Expected number of nodes: " + expected.length, expected.length, node.length);
```

Please also double check all of the uses of jqUnit.assertNode to ensure that changing this assertion doesn't break any of the existing uses. If it does, please rethink the change here or update those locations as needed.

jobara:
@dinukadesilva could you please update the copyright sections for all of the JS files you've touched in this pull request.

Basically that you should be able to follow the others in the file and add one for yourself.

dinukadesilva:
Hi @jobara I updated the copyright sections.

jobara:
merged at dacc70fc5318f4f944c54ea7c316b61a307be221

PR #678, opened 2016-03-06
FLUID-4051: Uploader : Added message to use a browser that supports HTML5

sakshamsaxena:
Updated the HTML files to include the message which would be
appropriately displayed.

### DIFF
--- a/demos/uploader/index.html
+++ b/demos/uploader/index.html
@@ -57,7 +57,7 @@
             <p>Use the Browse button to add a file, and the Save button to upload it.</p>
             <input name="fileData" type="file" />
             <input class="fl-uploader-basic-save" type="submit" value="Save"/>
-            <p class="fl-uploader-no-js-note">NOTE : You are using an outdated version of browser. Please upgrade your browser which supports HTML5 <noscript>and JavaScript </noscript>features.</p>
+            <p class="fl-uploader-no-js-note">NOTE : To use the full featured Uploader, please use a browser that supports HTML5 <noscript>and JavaScript </noscript>features.</p>
         </form>
 
         <!-- The Uploader's template will be injected via AJAX into this container. -->
--- a/demos/uploader/index.html
+++ b/demos/uploader/index.html
@@ -52,11 +52,12 @@
     </head>
     <body>
         <div class="flc-overviewPanel fl-overviewPanel-container"></div>
-        <!-- Basic upload controls, used when JavaScript is unavailable -->
+        <!-- Basic upload controls, used when JavaScript/HTML5 is unavailable -->
         <form method="post" enctype="multipart/form-data" class="fl-progEnhance-basic">
             <p>Use the Browse button to add a file, and the Save button to upload it.</p>
             <input name="fileData" type="file" />
             <input class="fl-uploader-basic-save" type="submit" value="Save"/>
+            <p class="fl-uploader-no-js-note">NOTE : You are using an outdated version of browser. Please upgrade your browser which supports HTML5 <noscript>and JavaScript </noscript>features.</p>
         </form>
 
         <!-- The Uploader's template will be injected via AJAX into this container. -->
--- a/src/components/uploader/css/Uploader.css
+++ b/src/components/uploader/css/Uploader.css
@@ -485,6 +485,10 @@ html>body .fl-uploader span.fl-uploader-browse {
     display:block;
 }
 
+.fl-uploader-no-js-note {
+    font-weight: bold;
+}
+
 /* hidden */
 .fl-uploader-hidden {
     display: none;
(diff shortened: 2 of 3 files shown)


jobara:
@sakshamsaxena thank you for your pull request. We ask that all our contributors sign a CLA. Could you please fill out the Fluid CLA and send it to me. https://wiki.fluidproject.org/display/fluid/Fluid+Licensing You can ping me in the IRC channel (Justin_o) if you have more questions on that.

For background reading, here's a good overview of what CLAs are and why they're important: http://oss-watch.ac.uk/resources/cla

jobara:
@sakshamsaxena thanks for sending me the CLA

src/components/uploader/css/Uploader.css
```diff
@@ -485,6 +485,10 @@ html>body .fl-uploader span.fl-uploader-browse {
     display:block;
 }
 
+.fl-uploader-no-js-note {
```
jobara: I think this class should probably be renamed. 

* Look at the [naming conventions](https://wiki.fluidproject.org/display/docs/FSS+Naming+Conventions). I think what you have here is quite close to the naming structure.
* The name seems specific to the case where JS isn't enabled, but also is used when HTML5 features aren't present. This should have a more generic name like `fl-uploader-basic-note` or `fl-uploader-basic-message`

demos/uploader/index.html
```diff
@@ -52,11 +52,12 @@
     </head>
     <body>
         <div class="flc-overviewPanel fl-overviewPanel-container"></div>
-        <!-- Basic upload controls, used when JavaScript is unavailable -->
+        <!-- Basic upload controls, used when JavaScript/HTML5 is unavailable -->
         <form method="post" enctype="multipart/form-data" class="fl-progEnhance-basic">
             <p>Use the Browse button to add a file, and the Save button to upload it.</p>
             <input name="fileData" type="file" />
             <input class="fl-uploader-basic-save" type="submit" value="Save"/>
+            <p class="fl-uploader-no-js-note">NOTE : You are using an outdated version of browser. Please upgrade your browser which supports HTML5 <noscript>and JavaScript </noscript>features.</p>
```
jobara: Maybe a message more like "To use the full featured Uploader, please use a browser that supports HTML5 and JavaScript". @colinbdclark what do you think?

sakshamsaxena: True, that did come across my mind, but then, wouldn't it be simply better if the user _does_ upgrade it for once and all? :sweat_smile: Let me know what you think

src/components/uploader/css/Uploader.css
sakshamsaxena: The expanse of your wiki is quite daunting to look at first :sweat_smile: I'll change this in a while.

demos/uploader/index.html
jobara: Well it's up to the user, they may have specific reasons why they do not want to or cannot upgrade their browser. Or they may prefer to navigate sites without scripts running all the time.

sakshamsaxena:
Hey. Did some changes like you said. I hope they're okay now.

demos/uploader/index.html
```diff
@@ -52,11 +52,12 @@
     </head>
     <body>
         <div class="flc-overviewPanel fl-overviewPanel-container"></div>
-        <!-- Basic upload controls, used when JavaScript is unavailable -->
+        <!-- Basic upload controls, used when JavaScript/HTML5 is unavailable -->
         <form method="post" enctype="multipart/form-data" class="fl-progEnhance-basic">
             <p>Use the Browse button to add a file, and the Save button to upload it.</p>
             <input name="fileData" type="file" />
             <input class="fl-uploader-basic-save" type="submit" value="Save"/>
+            <p class="fl-uploader-no-js-note">NOTE : To use the full featured Uploader, please use a browser that supports HTML5 <noscript>and JavaScript </noscript>features.</p>
```
jobara: The class `fl-uploader-no-js-note` needs to be updated to match `fl-uploader-basic-note` which is in the css.

jobara: You don't need the extra space between "Note" and ":". You could write "Note: ".

src/components/uploader/css/Uploader.css
```diff
@@ -485,6 +485,10 @@ html>body .fl-uploader span.fl-uploader-browse {
     display:block;
 }
 
+.fl-uploader-basic-note {
+    font-weight: bold;
+}
+
```
jobara: If the styling is only going to make the text bold, maybe use a `<strong>` element around the text instead.

demos/uploader/index.html
jobara: The above comments also apply to changes to the Uploader.html file

jobara:
@sakshamsaxena are you still working on this pull request?

PR #680, opened 2016-03-07
FLUID-4770 : Resolved Browse Files Button misalignment in Uploader

NRDa27:
The problem was their since we were using a fake button over the top of the new button

### DIFF
--- a/src/components/uploader/css/Uploader.css
+++ b/src/components/uploader/css/Uploader.css
@@ -237,12 +237,21 @@ tr.fl-uploader-file-state-uploaded button:before {
  */
 
 /* Fake add files button wrapper ("Browse files") */
+.fl-force-right{
+    width: 9.1em;
+    text-align: center;
+}
+
 .fl-uploader-browse {
     display: block;
     overflow: hidden;
     height: 2em;
-    width: 9em;
+    width: 9.1em;
     cursor: default;
+}
+
+span.fl-uploader-button-browse-text {
+    padding-left: 7px;
 }
 
 /* Fake add files button wrapper ("Add more")
--- a/src/components/uploader/css/Uploader.css
+++ b/src/components/uploader/css/Uploader.css
@@ -241,9 +241,8 @@ tr.fl-uploader-file-state-uploaded button:before {
     display: block;
     overflow: hidden;
     height: 2em;
-    width: 7em;
+    width: 9em;
     cursor: default;
-    padding-right: 30px;
 }
 
 /* Fake add files button wrapper ("Add more")
--- a/src/components/uploader/css/Uploader.css
+++ b/src/components/uploader/css/Uploader.css
@@ -243,6 +243,7 @@ tr.fl-uploader-file-state-uploaded button:before {
     height: 2em;
     width: 7em;
     cursor: default;
+    padding-right: 30px;
 }
 
 /* Fake add files button wrapper ("Add more")
--- a/src/components/uploader/html/Uploader.html
+++ b/src/components/uploader/html/Uploader.html
@@ -91,7 +91,7 @@
                 <div class="fl-text-align-right fl-force-right">
                     <span class="flc-uploader-button-browse fl-uploader-browse">
                         <span class="fl-infusionIcon-plus"></span>
-                        <span class="flc-uploader-button-browse-text fl-uploader-button-browse-text">Browse files</span>
+                        <span class="flc-uploader-button-browse-text fl-uploader-button-browse-text">Browse Files</span>
                     </span>
                 </div>
                 <!-- Total progress bar -->


jobara:
@NRDa27 I tried this is IE 11 and am still getting the clipping on the right side. In fact it seems that the "s" at the end of "Browser files" is always clipped. Also notice that originally the button label is "Browser files" and after clearing the queue the button is "Browser Files" which could be why it is taking up more space. It should probably have the same label in both cases.

src/components/uploader/css/Uploader.css
```diff
@@ -243,6 +243,7 @@ tr.fl-uploader-file-state-uploaded button:before {
     height: 2em;
     width: 7em;
     cursor: default;
+    padding-right: 30px;
```
jobara: Better to not use px values if possible. We tend to use rem or em instead because they will scale with text size changes.

jobara:
@NRDa27  could you also please amend your commit log to include the JIRA number at the beginning.  And the commit log should also summarize the change.

See:

* https://wiki.fluidproject.org/display/fluid/Coding+and+Commit+Standards#CodingandCommitStandards-JIRAandgithub
* https://help.github.com/articles/changing-a-commit-message/

NRDa27:
My solution is not working in IE because of the box model issue of IE, I totally forgot about that.
Let me try some other method

NRDa27:
@jobara, can you take a look at my new fix for this issue, I tested it in IE and it works fine.

src/components/uploader/css/Uploader.css
```diff
@@ -241,7 +241,7 @@ tr.fl-uploader-file-state-uploaded button:before {
     display: block;
     overflow: hidden;
     height: 2em;
-    width: 7em;
+    width: 9em;
```
jobara: Does it have to be all the to `9em`? There appears to be a lot of extra space to the right at this size.

jobara:
@NRDa27 also please correct the commit logs. We require that the commit logs start with the JIRA number (see: [Commit Logs](https://wiki.fluidproject.org/display/fluid/Coding+and+Commit+Standards#CodingandCommitStandards-CommitLogs) ). Because you have multiple commits to fix you may need to use `git reset --soft <commit hash>` to go back to the revision before your commits and then recommit your changes. To push those back up you'll need to then do a `git push origin -f`. Let me know if you have any questions about this.

src/components/uploader/css/Uploader.css
NRDa27: if width is 9em then output is [https://www.dropbox.com/s/vmkl45ftt36inf4/Screenshot%20from%202016-03-17%2023%3A44%3A40.png?dl=0]
if width is 8em then output is [https://www.dropbox.com/s/s91r5engrcdxsh7/Screenshot%20from%202016-03-17%2023%3A44%3A21.png?dl=0]

jobara: @NRDa27 is that specific to Firefox or do you get that elsewhere too?

NRDa27: I observed bit more width in Chrome when button size is 9em, because of different renderer in Chrome and Firefox 
But 9em seems to be perfect for Firefox

jobara: I've tried this in Safari, Chrome, Firefox (Mac OS and Win 8.1), and IE 11 and found similar results with the spacing being too wide on the right hand side. Can you try it out again on your end and determine if what you are getting is specific to Linux. Also centring the text of the button may make it look less odd. 
screen shot 2016-04-01 at 10 09 23 am

cindyli:
Thanks for improving the position of the "Browse Files" button. One issue is the change also affect the position of "add more" button that appear after you add some files to the queue. You can see from the image below that:

1. The "add more" button is a bit too far from the right border;
2. The re-positioning of the "add more" button causes the text to its left and the button itself are displayed in 2 separate lines instead of one line.

screen shot 2016-04-21 at 2 37 20 pm

src/components/uploader/css/Uploader.css
```diff
@@ -237,14 +237,23 @@ tr.fl-uploader-file-state-uploaded button:before {
  */
 
 /* Fake add files button wrapper ("Browse files") */
+.fl-force-right{
```
cindyli: Isn't it a bit weird to use a css class named "force-right" to center text?

NRDa27: Its the name given to the div of that button and center text is used to center the text in that div. Since the button needs to be on the right hand side so name of that div is force-right

cindyli: The "fl-force-right" is defined [here](https://github.com/fluid-project/infusion/blob/master/src/framework/core/css/fluid.css#L8) for the purpose indicated in its name. We should be careful with abusing it for other purposes. If width increase and center text are your intentions, consider defining another css class or even restructuring css classes that are already attached for this div.

cindyli:
@NRDa27, do you plan to continue working on this pull request? If not, this pull request will be closed due to its inactivity.

cindyli:
Closed as an inactive pull request.

PR #683, opened 2016-03-07
FLUID-5866: Removing implementation of "fluid.priorityHolder"

amb26:
and rewriting test case to demonstrate that it is unnecessary


amb26:
@colinbdclark  - a bit of cleanup after the last big pull

cindyli:
Merged at 6b79a758ba43ce98f6305603b2e5f73d4fbf8652

PR #685, opened 2016-03-09
FLUID-2779 and FLUID-3921: Fix links in pager

PaliwalSparsh:
Added required classes to the pager demo and fixed the pager.css in demo
and Pager.css in pager component in src.

Fixes -
1. Cursor on disabled and current link is arrow.
2. When disabled links appear gray.
3. Fixed implementation of classes in demo. On hovering link appear blue as it was intended according to Pager.css in src.
4. Fixed spaces in pager.css and Pager.css

Test -
Tested in the pager demo for which changes were intended.

### DIFF
--- a/demos/pager/css/pager.css
+++ b/demos/pager/css/pager.css
@@ -1,7 +1,7 @@
 /* pager demo styles */
 
 .demo-pager-container {
-    width:95%;
+    width: 95%;
 }
 
 .demo-pager-container input {
@@ -12,13 +12,9 @@
     display: none;
 }
 
-.demo-pager-container a:visited {
-    color: blue;
-}
-
 .demo-pager-container a:hover{
     outline: 2px #000 solid;
-    padding:0px;
+    padding: 0px;
 }
 
 /***********************************************
@@ -58,8 +54,8 @@
 }
 
 .demo-pager-bar  li  {
-    display:inline;
-    float:left;
+    display: inline;
+    float: left;
 }
 
 /***********************************************
@@ -77,13 +73,12 @@
 
 
 .demo-pager-pageLink-default a {
-    /* Using display:block and float:left, it creates helps create a horizontal list with fixed width elements.*/
-    display:block;
-    float:left;
-
-    width:3em;
-    text-align:center;
-    margin:2px;
+    /* Using display:block and float:left, it helps create a horizontal list with fixed width elements.*/
+    display: block;
+    float: left;
+    width: 3em;
+    text-align: center;
+    margin: 2px;
     border: 1px solid #afafaf;
 }
 
@@ -92,15 +87,13 @@
 }
 
 .demo-pager-links .fl-pager-currentPage {
-    color: #fff;
-    display:block;
-    width:3em;
-    float:left;
-    background: #000;
-    padding:0px;
-    margin:2px;
+    display: block;
+    width: 3em;
+    float: left;
+    padding: 0px;
+    margin: 2px;
     border: 1px solid #000;
-    text-align:center;
+    text-align: center;
 }
 
 .demo-pager-links .fl-pager-currentPage:visited {
@@ -108,8 +101,8 @@
 }
 
 .demo-pager-head a {
-    display:block;
-    width:100%;
+    display: block;
+    width: 100%;
 }
 
 /***********************************************
--- a/src/components/pager/css/Pager.css
+++ b/src/components/pager/css/Pager.css
@@ -18,41 +18,52 @@
 /* Pager UI */
 /* Class name for the list container */
 .fl-pager .fl-pager-ui {
-    margin:0; padding:0;
+    margin: 0;
+    padding: 0;
 }
 .fl-pager .fl-pager-ui li {
-    list-style-type:none;
-    display:inline;
+    list-style-type: none;
+    display: inline;
 }
 .fl-pager .fl-pager-ui a {
-    padding:0 3px;
-    margin:0 2px;
+    margin: 0 2px;
 }
 .fl-pager .fl-pager-ui .fl-pager-links a {
-    border:1px solid #ccc;
-    background-color:#fff;
-    zoom:1;
-}
-.fl-pager .fl-pager-ui .fl-pager-disabled,
-.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage {
-    color:#000;
-    border:none;
-    background-color:transparent;
-    text-decoration:none;
+    border: 1px solid #ccc;
+    background-color: #fff;
+    zoom: 1;
 }
 .fl-pager .fl-pager-ui .fl-pager-links a:hover {
-    background-color:#0287C5;
-    color:#fff;
+    background-color: #0287C5;
+    color: #fff;
     text-decoration:none;
 }
-.fl-pager .fl-pager-ui .fl-pager-disabled:hover,
-.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage:hover {
-    background-color:transparent;
-    color:#000;
-    text-decoration:none;
-    cursor:default;
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage {
+    color: #fff;
+    border: none;
+    outline: 0px;
+    background-color: #000;
+    text-decoration: none;
+    cursor: default;
+}
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage:hover,
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage:focus {
+    outline: 2px #000 solid;
+    background-color: #000;
+}
+.fl-pager .fl-pager-ui .fl-pager-disabled a,
+.fl-pager .fl-pager-ui .fl-pager-disabled a:hover {
+    color: #afafaf;
+    cursor: default;
+    border: none;
+    outline: 0px;
+    pointer-events: none;
 }
 
+/*On :focus doesnt work*/
+.fl-pager .fl-pager-ui .fl-pager-disabled:focus {
+    outline: 2px #000 solid;
+}
 
 /***********************************************************/
 /** THEMES - High Contrast **/
--- a/src/components/pager/js/Pager.js
+++ b/src/components/pager/js/Pager.js
@@ -295,8 +295,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     });
 
     fluid.pager.previousNext.update = function (that, disabledStyle, newModel) {
-        that.previous.toggleClass(disabledStyle, newModel.pageIndex === 0);
-        that.next.toggleClass(disabledStyle, newModel.pageIndex === newModel.pageCount - 1);
+        var isFirstPage = (newModel.pageIndex === 0);
+        var isLastPage = (newModel.pageIndex === newModel.pageCount - 1);
+        that.previous.toggleClass(disabledStyle, isFirstPage);
+        that.next.toggleClass(disabledStyle, isLastPage);
+        that.previous.attr("aria-disabled", isFirstPage);
+        that.next.attr("aria-disabled", isLastPage);
     };
 
     fluid.defaults("fluid.pager.pagerBar", {
--- a/demos/pager/index.html
+++ b/demos/pager/index.html
@@ -57,13 +57,13 @@
         <div class="flc-pager-summary">Summary goes here</div>
 
         <div class="flc-pager-top">
-            <ul class="demo-pager-bar fl-force-left fl-focus">
+            <ul class="demo-pager-bar fl-force-left fl-focus fl-pager-ui">
                 <li class="flc-pager-previous demo-pager-previous"><a href="#" title="Previous Page">&lt; Previous</a></li>
                 <li>
                     <ul class="flc-pager-links demo-pager-links">
-                        <li class="flc-pager-pageLink demo-pager-pageLink-default"><a href="#">1</a></li>
+                        <li class="flc-pager-pageLink demo-pager-pageLink-default fl-pager-links"><a href="#">1</a></li>
                         <li class="flc-pager-pageLink-skip demo-pager-pageLink-skip ">...</li>
-                        <li class="flc-pager-pageLink"><a href="#">2</a></li>
+                        <li class="flc-pager-pageLink demo-pager-pageLink-default fl-pager-links"><a href="#">2</a></li>
                     </ul>
                 </li>
                 <li class="demo-pager-next flc-pager-next"><a href="#" title="Next Page">Next &gt;</a></li>


jobara:
@PaliwalSparsh is the "FLUID-2279" JIRA number correct. It doesn't seem related to what you changed, but maybe I missed something. If you need to correct the commit log you can amend it. https://help.github.com/articles/changing-a-commit-message/

demos/pager/css/pager.css
```diff
@@ -78,12 +74,11 @@
 
 .demo-pager-pageLink-default a {
     /* Using display:block and float:left, it creates helps create a horizontal list with fixed width elements.*/
```
jobara: While you're making changes in this file, could you please correct the comment here by removing the extra "creates"?

PaliwalSparsh:
@jobara Sorry for the inconvenience it was a mistake on my part, I have fixed the issue name and also the "create" in the comment.

src/components/pager/css/Pager.css
```diff
@@ -18,41 +18,51 @@
...
-    cursor:default;
+    background-color: #000;
+    color: #fff;
+    text-decoration: none;
+    cursor: default;
+    pointer-events: none;
+    cursor: default;
+}
+.fl-pager .fl-pager-ui .fl-pager-disabled:hover,
+.fl-pager .fl-pager-ui .fl-pager-disabled a:hover {
```
jobara: can this be combined with what's on line 44?

PaliwalSparsh:
Thanks for the advice I have made all the changes and added a property I accidentally removed before. I have checked it through once again, seems to be working perfectly to me.

src/components/pager/css/Pager.css
```diff
@@ -18,41 +18,51 @@
...
+    color: #fff;
     text-decoration:none;
 }
-.fl-pager .fl-pager-ui .fl-pager-disabled:hover,
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage {
+    color: #fff;
+    border: none;
+    background-color: #000;
+    text-decoration: none;
+    pointer-events: none;
```
jobara: The current page should still show the tooltip on hover. Could you remove the `pointer-events: none;` declaration.

src/components/pager/css/Pager.css
```diff
@@ -18,41 +18,44 @@
...
-.fl-pager .fl-pager-ui .fl-pager-disabled:hover,
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage,
 .fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage:hover {
-    background-color:transparent;
-    color:#000;
-    text-decoration:none;
-    cursor:default;
+    color: #fff;
+    border: none;
+    outline: 2px #000 solid;
```
jobara: Maybe don't add the outline unless the page link has focus/hover. When it's there all the time on the current page it causes the borders of the adjacent page links to touch. Also it adds another visual indicator for which page link has focus/hover.

screen shot 2016-04-04 at 8 57 55 am

PaliwalSparsh:
I have splited the two classes, and made the appropriate changes in the required class.

src/components/pager/css/Pager.css
jobara: the outline should be added on hover and focus.

PaliwalSparsh:
Added aria-attributes to prev/next and ability to focus at them.

src/components/pager/css/Pager.css
```diff
@@ -18,41 +18,57 @@
...
+    cursor: default;
+}
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage:hover,
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage:focus {
+    color: #fff;
+    border: none;
+    outline: 2px #000 solid;
+    background-color: #000;
+    text-decoration: none;
+    cursor: default;
```
jobara: is it possible to just have the outline specified here. The rest seem like duplicates from the previous style declaration.

src/components/pager/js/Pager.js
```diff
@@ -295,8 +295,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
 
     fluid.pager.previousNext.update = function (that, disabledStyle, newModel) {
-        that.previous.toggleClass(disabledStyle, newModel.pageIndex === 0);
-        that.next.toggleClass(disabledStyle, newModel.pageIndex === newModel.pageCount - 1);
+        var previousToggle = (newModel.pageIndex === 0);
+        var nextToggle = (newModel.pageIndex === newModel.pageCount - 1);
+        that.previous.toggleClass(disabledStyle, previousToggle);
+        that.next.toggleClass(disabledStyle, nextToggle);
+        that.previous.children('a').attr('aria-disabled', previousToggle);
+        that.next.children('a').attr('aria-disabled', nextToggle);
```
jobara: I was reading through some spec documentation for [aria-disabled](https://www.w3.org/TR/wai-aria/states_and_properties#aria-disabled). It says 

> The state of being disabled applies to the current element and all focusable descendant elements of the element on which the aria-disabled attribute is applied.

That would lead me to believe that the `aria-disabled` attribute can be set directly on the previous and next elements.

jobara: Also we use `"` not `'` around strings. Please make sure to update this and run `grunt lint` to verify that there are no other code style issues.

src/components/pager/css/Pager.css
PaliwalSparsh: We would also require to add background-color as black or on hover we will get background color as blue that is what we have set for normal page link.

src/components/pager/js/Pager.js
PaliwalSparsh: I was using aria-attributes for the first time, so had no clue about that and really really sorry for the linting issue. This is really something I should have not done. Sorry for the amount of time I have made you invest into reviewing this.

PaliwalSparsh: I tried doing what you told I removed the that.previous.children("a").attr(...) and put that.previous.attr(...) but as a result of this when I use the keyboard to put focus on the disabled link, screen reader says "internal link disabled", but when i click with mouse on the disabled link it focuses on the link inside and says "internal link" only. Though I have pushed accordingly to what you said.

src/components/pager/js/Pager.js
```diff
@@ -295,8 +295,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     });
 
     fluid.pager.previousNext.update = function (that, disabledStyle, newModel) {
-        that.previous.toggleClass(disabledStyle, newModel.pageIndex === 0);
-        that.next.toggleClass(disabledStyle, newModel.pageIndex === newModel.pageCount - 1);
+        var previousToggle = (newModel.pageIndex === 0);
+        var nextToggle = (newModel.pageIndex === newModel.pageCount - 1);
```
jobara: I do still think you need to rename "previousToggle" and "nextToggle" as they do not refer to the toggles but are related to their states.

PaliwalSparsh:
Added the corrected variable names as isFirstPage and isLastPage.

klown:
@jobara You are correct that aria-disabled is propagated to (some of) the descendant elements of a container element.  I looked at the a11y tree produced by FF 45.0  and Safari 9.1 and: the &lt;li&gt; elements are disabled, as well their &lt;a&gt; children.

To be complete:
- FF exposes the parents with role "listitem" and the children with role "link".
- Safari exposes the parents with role "group" and the children with role "link".

However, I'm not convinced that the links are truly disabled.  In FF, the links are exposed as focusable -- is that what is intended?  What is supposed to happen when the "disabled" link acquires focus when a user clicks on it?  I couldn't tell if the focus was transferred to the containing &lt;li&gt;.  By comparison, it looks to me like that is where focus goes during TAB navigation (to the container, not the link).

I found some advice that one should use the CSS `pointer-events` selector, and set it to "`none`" to stop a link from gaining focus due to mouse clicks.  Unfortunately, `pointer-events` does not appear to be standard, even though it has fairly good browser support.

See David Walsh's [article and demo](https://davidwalsh.name/pointer-events), and Mozilla's developer page that [documents `pointer-events`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events).  Aside:  it looks like `pointer-events` is based on the [SVG attribute](http://www.w3.org/TR/SVG11/interact.html#PointerEventsProperty) of the same name.

Hope that helps.

jobara:
@klown thanks for the info. It was decided to keep the previous/next links in the tab order when they are disabled so that a user would still know they are there. (See: https://botbot.me/freenode/fluid-work/2016-04-05/?msg=63557568&page=2 ). However, it may make sense to set the `pointer-event` css declaration to `none` to knockout the mouse interactions.

jobara:
@PaliwalSparsh please see the above comments.

klown:
@jobara 
> It was decided to keep the previous/next links in the tab order ...

The links?  Or the list items containing the links?  (Or both?).

jobara:
@klown 

> The links? Or the list items containing the links? (Or both?).

Good question. I'm sure the conversation was discussed as though they were an atomic unit. And in terms of interaction they "behave" as a single entity. That is you can't tab focus both the `<li>` and `<a>` independently and clicking anywhere in the `<li>` will trigger the action.

klown:
@jobara 
> That is you can't tab focus both the &lt;li&gt; and &lt;a&gt; independently ... 

Right, but...

> ... and clicking anywhere in the &lt;li&gt; will trigger the action.

Clicking will put focus on the &lt;a&gt;.  At least, I think it will.  Remind me:  doesn't the click go to the deepest element in the tree that can accept a click?

PaliwalSparsh:
@klown You are absolutely right about this, click goes to the deepest element. I checked it.

Also as you suggested in the discussion above,
On setting pointer-events: none on <a>, when I click on the <li>(gave it the focus property of <a> like outline), focus neither comes to <a> (expected as pointer event: none) nor it focus comes to <li> where I expected, I have no clue why this is happening.

I guess as we cannot put focus on a div, as div is non-selectable element. Similarly we cannot put focus on a <li>.

PaliwalSparsh:
@jobara @klown 
Also as you suggested in the discussion above.
On setting ```pointer-events: none``` on``` <a>```, when I click on the``` <li>```(gave it the focus property of ```<a>``` like outline), focus neither comes to``` <a> ```(expected as``` pointer event: none```) nor it focus comes to ```<li> ```where I expected, I have no clue why this is happening.
I guess as we cannot put focus on a div, as div is non-selectable element. Similarly we cannot put focus on a ```<li>```.

klown:
@PaliwalSparsh @jobara 

> I guess as we cannot put focus on a div, as div is non-selectable element. Similarly we cannot put focus on a &lt;li&gt;.

You can put focus on any element, in theory.  I don't know the actual code or design here, so I'm not sure which of the following applies, but here are a few thoughts.

1. Put tabindex="0" on the &lt;li&gt;.  That puts the &lt;li&gt; in the TAB order, and it will gain focus when the user navigates to it by keyboard.
2. Call the element's `focus() `method in script.  Something in the logic must  of the app has to lead to this.  For example, add a click listener to the &lt;li&gt; that calls the &lt;li&gt;'s `focus() ` method.
3. Do both of the above.

References:
- [`tabindex`](https://www.w3.org/TR/html5/editing.html#sequential-focus-navigation-and-the-tabindex-attribute), specifically the section titled "If the value is a zero".
- [`focus()`](https://www.w3.org/TR/html5/editing.html#element-level-focus-apis)

jobara:
@PaliwalSparsh i'm not exactly sure I understand the problem you are describing. Maybe you can push up the changed code for us to try?

PaliwalSparsh:
@jobara
> I found some advice that one should use the CSS pointer-events selector, and set it to "none" to stop a link from gaining focus due to mouse clicks.

I have pushed a PR with this change. 
I have changed the `pointer-events : none` here. This works.
```
.fl-pager .fl-pager-ui .fl-pager-disabled a,
.fl-pager .fl-pager-ui .fl-pager-disabled a:hover {
    color: #afafaf;
    cursor: default;
    border: none;
    outline: 0px;
    pointer-events: none;
}
```
But, `:focus` pseudo class code won't have any effect.
```
.fl-pager .fl-pager-ui .fl-pager-disabled:focus {
    outline: 2px #000 solid;
}
```
This is required to have a black outline around the element when it gains focus.


So when you will use screen reader and click on previous or next, though you wont get the black outline  around the element but the good thing is screen reader will always read "internal link disabled".
Problem is how to get this black outline when element gains focus, because :focus is not working here.

I guess for this @klown has suggested some methods so @jobara which one should I use.

src/components/pager/css/Pager.css
```diff
@@ -18,41 +18,52 @@
...
+.fl-pager .fl-pager-ui .fl-pager-links .fl-pager-currentPage:focus {
+    outline: 2px #000 solid;
+    background-color: #000;
+}
+.fl-pager .fl-pager-ui .fl-pager-disabled a,
+.fl-pager .fl-pager-ui .fl-pager-disabled a:hover {
+    color: #afafaf;
+    cursor: default;
+    border: none;
+    outline: 0px;
```
cindyli: `:focus` pseudo class code does have effect if this line is removed. Are there other features that require this line?

cindyli: Make sure to test across browsers to make sure the use of `pointer-events: none``` works fine.

src/components/pager/js/Pager.js
```diff
@@ -295,8 +295,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     });
 
     fluid.pager.previousNext.update = function (that, disabledStyle, newModel) {
-        that.previous.toggleClass(disabledStyle, newModel.pageIndex === 0);
-        that.next.toggleClass(disabledStyle, newModel.pageIndex === newModel.pageCount - 1);
+        var isFirstPage = (newModel.pageIndex === 0);
```
cindyli: The handling for the previous and next links are pretty much same with the difference on the calculation of the flag (isFirstPage & isLastPage), can you create a generic function with the flag being the pass-in parameter to share this handling? Thanks.

cindyli: Actually, a better solution is to use `modelRelay` to calculate `isFirstPage` & `isLastPage` based on the source model values `{pager}.model.pageIndex` and `{pager}.model.pageCount`, then attach model listeners that listen to the model changes of `isFirstPage` & `isLastPage` to update disabled states.

An example:
Model Relay: https://github.com/GPII/first-discovery/blob/master/src/js/panels.js#L640-L646
The corresponding model listener: https://github.com/GPII/first-discovery/blob/master/src/js/panels.js#L671-L674

src/components/pager/css/Pager.css
PaliwalSparsh: `.fl-pager-currentPage` class is appended to the `<a>` tag which is inside a `<li>` item which are currently selected. So when we navigate with tab or click and reach the current page it gains focus. All apologies but I am not completely able to understand the first comment.

Really thanks for reminding me about the browser compatibility I checked it, it is supported on all browsers stable realeases.
http://caniuse.com/#feat=pointer-events

cindyli:
Sorry that I didn't make my first comment clear. That comment was in answer to your question in https://github.com/fluid-project/infusion/pull/685#issuecomment-210309347 where you asked: "how to get this black outline when element gains focus, because `:focus` is not working here".

I played with your css and found when the css line `outline: 0px;` (the line where my first comment is placed for) is removed, `:focus` will work fine. Of course we need to make sure the removal of this line does not break other features.

@PaliwalSparsh, thanks for checking the browser compatibility for `pointer-events`. It would be even better if this pull request can be physically tested across browsers. Sometimes, the combined use of one property with other css properties is not respected in some browsers.

PaliwalSparsh:
@cindyli Thanks for the advice I got it working and now I am able to get that outline on focus when navigating through keyboard but it is not gaining focus when I click on that element(i.e. previous or next disabled button), like when we click on current page link the outline appears and stays there as it has gained focus. I guess the focus should be gained using both keyboard and mouse interaction.

cindyli:
Yes, you are right that the focus should be gained using both keyboard and mouse interaction.

amb26:
Closed due to inactivity

PR #693, opened 2016-03-29
Fluid-5874: Text overlaps on headers when default theme is used on chrome browser.

lemetice:
Overlapping text disappears when default theme is used on chrome.


jobara:
@lemetice i was testing out this pull request, but I am unable to reproduce the issue it is referring to before applying your fix. Can you confirm that the issue is still valid?

jhung:
Tested this issue and it seems to have been resolved. It no longer seems to be an issue in Chrome. Perhaps this was addressed by another pull req or a browser update.

PR #695, opened 2016-03-31
FLUID-5294: Have removed default support for value.

kaspermarkus:


amb26:
Thanks, KASPARDNITT, this is a brilliantly useful pull request, and absolutely perfect! Please could you also update the accompanying docs that we have (that you have written) at https://wiki.gpii.net/w/Architecture_-_Available_transformation_functions in order to account for the change? Or perhaps rework them into markdown to accompany the main framework in https://github.com/fluid-project/infusion-docs since we probably don't yet want to confuse the GPII community by advertising changes that haven't reached GPII trunk yet

PR #698, opened 2016-04-05
FLUID-3258: Extra <div> in Image Reorderer demo markup not necessary

blag0jd:
FLUID-3258: Extra < div> in Image Reorderer demo markup not necessary.
The extra < div> tag removed.


jobara:
merged at 20e3757c5dbeffb202e7a216dfe8eb42ae8186c4

PR #699, opened 2016-04-05
Fluid-2797: The prev/next link moves as user clicks previous

blag0jd:
Fluid 2797: The prev next link moves as user clicks previous, bug fixed by making the List Item container's width fixed.

### DIFF
--- a/src/components/pager/css/Pager.css
+++ b/src/components/pager/css/Pager.css
@@ -53,6 +53,12 @@
     cursor:default;
 }
 
+/* Pager List Item Conatiner Width */ 
+
+.flc-pager-li-container {
+ width:550px !important; /* FLUID-2797 The prev/next link moves as user clicks previous, Next Button Click Fix */
+}
+
 
 /***********************************************************/
 /** THEMES - High Contrast **/
--- a/demos/pager/index.html
+++ b/demos/pager/index.html
@@ -59,7 +59,8 @@
         <div class="flc-pager-top">
             <ul class="demo-pager-bar fl-force-left fl-focus">
                 <li class="flc-pager-previous demo-pager-previous"><a href="#" title="Previous Page">&lt; Previous</a></li>
-                <li>
+                <li class="flc-pager-li-container"> <!-- FLUID-2797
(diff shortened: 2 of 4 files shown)


src/components/pager/css/Pager.css
```diff
@@ -53,6 +53,12 @@
     cursor:default;
 }
 
+/* Pager List Item Conatiner Width */ 
+
+.flc-pager-li-container {
```
jobara: The name of this class will need to change as a user may use something other than an `<li>` also should consider if they don't have that containing element at all.

blag0jd:
@jobara , I think you are right about the class name, but about the component, the `<li>`  was already there, and it is part of the original pager component. I think that removing the `<li>` will result in destructing the component, and also in this issue description the main point is on "fixing" the previous and next buttons. I think giving the class a better name will do the job. What do you think of this?

jobara:
@blag0jd yes, I'm not asking you to remove the `<li>` but the general philosophy of Infusion is to be DOM agnostic. That is, to not require specific markup. A user/integrator can provide their own markup and direct the component via the selectors block in the component's options. You can read more about this in the [DOM Binder documentation](http://docs.fluidproject.org/infusion/development/DOMBinder.html).

Specifically a name change should help.

src/components/pager/css/Pager.css
```diff
@@ -53,6 +53,12 @@
...
+ width:550px !important; /* FLUID-2797 The prev/next link moves as user clicks previous, Next Button Click Fix */
```
jobara: I'm not sure this fixed sized width is desirable. As the number of page links changes the gap between them and the next button is quite large. Also, you should try to avoid using `!important`.

Here are a few screen shots as an example.
screen shot 2016-04-12 at 9 49 36 am
screen shot 2016-04-12 at 9 52 54 am

jobara: @dayotte do you have any thoughts on this?

cindyli:
@blag0jd, do you plan to continue working on this pull request? If not, this pull request will be closed due to its inactivity.

cindyli:
Closed as an inactive pull request.

PR #700, opened 2016-04-11
FLUID-4991: The shadow effect around the hide/show preferences bug fix

blag0jd:
Shadow Bug fix for the bug: The shadow effect around the hide/show
preferences tab should change colour with the contrast theme.

### DIFF
--- a/src/framework/preferences/css/stylus/utils/Themes.styl
+++ b/src/framework/preferences/css/stylus/utils/Themes.styl
@@ -4,23 +4,38 @@
 themes = {
     ".fl-theme-bw": {
         foregroundColor: #000000,
-        backgroundColor: #ffffff
+        backgroundColor: #ffffff,
+        .fl-prefsEditor-buttons { //FLUID-4991: The shadow effect around the hide/show preferences bug fix.
+          box-shadow:none !important;
+          }
     }
     ".fl-theme-wb": {
         foregroundColor: #ffffff,
-        backgroundColor: #000000
+        backgroundColor: #000000,
+        .fl-prefsEditor-buttons { //FLUID-4991: The shadow effect around the hide/show preferences bug fix.
+          box-shadow:none !important;
+          }
     }
     ".fl-theme-yb": {
         foregroundColor: #ffff00,
-        backgroundColor: #000000
+        backgroundColor: #000000,
+        .fl-prefsEditor-buttons { //FLUID-4991: The shadow effect around the hide/show preferences bug fix.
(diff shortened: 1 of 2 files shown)


jobara:
@blag0jd i notice that your files have the wrong permissions. They should be set to 100644 but are set at  100755.

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -4,23 +4,38 @@
 themes = {
     ".fl-theme-bw": {
         foregroundColor: #000000,
-        backgroundColor: #ffffff
+        backgroundColor: #ffffff,
+        .fl-prefsEditor-buttons { //FLUID-4991: The shadow effect around the hide/show preferences bug fix.
```
jobara: I'm getting an error when trying to build the stylus files. I believe that the class name might need to be wrapped in `"`

blag0jd:
@jobara , I've just fixed both of the issues. The first one was because I switched to another OS, and then to edit the files I set their permissions to 777, and forget to change them back.

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -331,3 +331,9 @@ build-themes-Enactors(contrastThemes) {
         }
     }
 }
+
+.fl-theme-prefsEditor-default { //FLUID-4991: The shadow effect around the hide/show preferences bug fix.
+    .fl-prefsEditor-buttons {
+   box-shadow: 2px 2px 3px 0 #6f6f6f;
```
jobara: looks like you have a indenting issue here.

jhung:
I checked the CSS and drop shadow effect on contrast themes is fixed (i.e. gone).

jobara:
@blag0jd I've been looking over the changes a bit more closely and feel like it's in the wrong spot. I think you should keep the box-shadow setting in SeparatePanelPrefsEditor.styl. In Themes.styl, instead of adding that new line to the bottom, you should modify https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/css/stylus/utils/Themes.styl#L91-L95 to set `box-shadow: none` on `.fl-prefsEditor-buttons`

blag0jd:
@jobara, you're right, fixing the code in that way, is a better solution. I've just pushed the latest commit regarding this fix.

cindyli:
Merged at e06b8fd915e8528107d08690c9b711bd74b01257

PR #701, opened 2016-04-14
FLUID-5133/FLUID-5890/FLUID-5478: Improved support for inversion.

kaspermarkus:
* Rolled the switcheroo of inputPath and outputPath for standardTransformFunctions into the general framework inversion functions.
* Went through all transforms and checked for correctness and tests.
* Notes on invertibility will be added to the documentation (currently kept here: https://issues.fluidproject.org/browse/FLUID-5133) - will be added after discussion with antranig
* Added inRange and quantize functions (FLUID-5890)
* Added inverse function of stringToNumber (namely numberToString) - plus made both functions lenses
* Changed arrayToSetMembership and setMembershipToArray's grades to be standardTransformFunction
* Added a few tests where needed

### DIFF
--- a/src/framework/core/js/ModelTransformation.js
+++ b/src/framework/core/js/ModelTransformation.js
@@ -220,15 +220,7 @@ var fluid = fluid || fluid_2_0_0;
             expdef = fluid.defaults("fluid.standardTransformFunction");
         }
         var transformArgs = [transformSpec, transform];
-        if (fluid.hasGrade(expdef, "fluid.standardInputTransformFunction")) {
-            var expanded = fluid.model.transform.getValue(transformSpec.inputPath, transformSpec.input, transform);
-
-            transformArgs.unshift(expanded);
-            // if the function has no input, the result is considered undefined, and this is returned
-            if (expanded === undefined) {
-                return undefined;
-            }
-        } else if (fluid.hasGrade(expdef, "fluid.multiInputTransformFunction")) {
+        if (fluid.hasGrade(expdef, "fluid.multiInputTransformFunction")) {
             var inputs = {};
             fluid.each(expdef.inputVariables, function (v, k) {
                 inputs[k] = function () {
@@ -240,6 +232,15 @@ var fluid = fluid || fluid_2_0_0;
                 };
             });
             transformArgs.unshift(inputs);
+        }
+        if (fluid.hasGrade(expdef, "fluid.standardInputTransformFunction")) {
+            var expanded = fluid.model.transform.getValue(transformSpec.inputPath, transformSpec.input, transform);
+
+            transformArgs.unshift(expanded);
+            // if the function has no input, the result is considered undefined, and this is returned
+            if (expanded === undefined) {
+                return undefined;
+            }
         }
         var transformed = transformFn.apply(null, transformArgs);
         if (fluid.hasGrade(expdef, "fluid.standardOutputTransformFunction")) {
@@ -402,9 +403,11 @@ var fluid = fluid || fluid_2_0_0;
 
         if (standardInput) {
             fluid.model.transform.accumulateStandardInputPath("input", transformSpec, transform, transform.inputPaths);
-        } else if (multiInput) {
+        }
+        if (multiInput) {
             fluid.model.transform.accumulateMultiInputPaths(defaults.inputVariables, transformSpec, transform, transform.inputPaths);
-        } else {
+        }
+        if (!multiInput && !standardInput) {
             var collector = defaults.collectInputPaths;
             if (collector) {
                 var collected = fluid.makeArray(fluid.invokeGlobalFunction(collector, [transformSpec, transform]));
--- a/src/framework/core/js/ModelTransformation.js
+++ b/src/framework/core/js/ModelTransformation.js
@@ -160,13 +160,13 @@ var fluid = fluid || fluid_2_0_0;
 
     // TODO: This cut and pasted code was hoisted out of the inverse of transforms - it needs to go into the
     // the inversion mechanism itself
-    fluid.model.transform.copyInversePaths = function (transformSpec, transformer) {
-        var togo = fluid.copy(transformSpec);
+    fluid.model.transform.invertPaths = function (transformSpec, transformer) {
         // TODO: this will not behave correctly in the face of compound "input" which contains
         // further transforms
-        togo.inputPath = fluid.model.composePaths(transformer.outputPrefix, transformSpec.outputPath);
-        togo.outputPath = fluid.model.composePaths(transformer.inputPrefix, transformSpec.inputPath);
-        return togo;
+        var oldOutput = fluid.model.composePaths(transformer.outputPrefix, transformSpec.outputPath);
+        transformSpec.outputPath = fluid.model.composePaths(transformer.inputPrefix, transformSpec.inputPath);
+        transformSpec.inputPath = oldOutput;
+        return transformSpec;
     };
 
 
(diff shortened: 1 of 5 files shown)


src/framework/core/js/ModelTransformation.js
```diff
@@ -382,6 +382,10 @@ var fluid = fluid || fluid_2_0_0;
     };
     // unsupported, NON-API function
     fluid.model.transform.handleInvertStrategy = function (transformSpec, transform, transformOpts) {
+        // if we have a standardTransformFunction we can switch input and output arguments:
+        if (fluid.hasGrade(transformOpts.defaults, "fluid.standardTransformFunction")) {
```
amb26: As simple as that! Excellent

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -368,32 +384,40 @@ var fluid = fluid || fluid_2_0_0;
         }
 
         fluid.each(options, function (outPath, key) {
-            // write to output path given in options the value <presentValue> or <missingValue> depending on whether key is found in user input
+            // write to output object the value <presentValue> or <missingValue> depending on whether key is found in user input
             var outVal = (value.indexOf(key) !== -1) ? transformSpec.presentValue : transformSpec.missingValue;
-            fluid.model.transform.setValue(outPath, outVal, transformer);
+            fluid.set(output, outPath, outVal);
```
amb26: Very interesting - can you explain what the implications are of adjusting this from setValue to fluid.set?

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -409,15 +433,19 @@ var fluid = fluid || fluid_2_0_0;
         }
 
         var outputArr = [];
-        fluid.each(options, function (arrVal, inPath) {
-            var val = fluid.model.transform.getValue(inPath, undefined, transformer);
+        fluid.each(input, function (val, key) {
             if (val === transformSpec.presentValue) {
-                outputArr.push(arrVal);
+                outputArr.push(options[key]);
```
amb26: Was this simply a blunder in the original impl? This looks like a pretty significant change

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -653,12 +695,11 @@ var fluid = fluid || fluid_2_0_0;
         return parsedOffset;
     };
 
-    fluid.transforms.invertArrayIndexation = function (transformSpec, transformer) {
-        var togo = fluid.model.transform.copyInversePaths(transformSpec, transformer);
-        if (!isNaN(Number(togo.offset))) {
-            togo.offset = Number(togo.offset) * (-1);
+    fluid.transforms.invertArrayIndexation = function (transformSpec) {
```
amb26: Now I see we need to be VERY VERY CLEAR because something quite subtle and probably highly dangerous has happened. The inversion path in the framework will now provide a copy of transformSpec to an fluid.standardTransformFunction inverter, by virtue of calling "copyInversePaths", but will provide the original transformSpec to other inverters. This is probably an unacceptable subtlety, and we should make sure to copy the transformSpec once during the other path too and clearly document this fact. All of this annoyance will go away with the IMMUTABLE REVOLUTION upcoming in the next revision of the core framework.

Once we have done this, we can eliminate the copy operation that inevitably occurs at the head of all the nonstandard transforms like valueMapper.

Also, the work in removing copyInversePaths is incomplete - we should be able to treat multiInput transforms as well, since the junk is still present in the inverse, e.g. of fluid.transforms.linearScale

src/framework/core/js/ModelTransformationTransforms.js
kaspermarkus: yeah - so this is related to changing it to a standardTransformFunction. Previously, the only way of outputting anything from this transform function was by specifying an "outputPath" - and `fluid.model.transform.setValue(outPath, outVal, transformer);` was the part responsible for writing to the output model.

With the change, we now use the general framework for standardOutputTransformFunction (allowing you to either return or write to outputPath, depending on whether the latter is specified).. So we're creating an `output` object, containing the result of this transformation, and return it to let the framework do the writing to outputPath or passing on to the parent transform function (depending on whether outputPath was specified)

src/framework/core/js/ModelTransformationTransforms.js
kaspermarkus: As you can see in line 416, I changed this to a standardInputTransformFunction, so the `input`/`inputPath` is automatically parsed by the framework - so that's the reason for removing the 'getValue' portion. The change in the each function (looping through inputs instead of options) was simply because it seemed more logical to look each input key up in the options block than vice versa.

I guess looking at it now, this result in a difference in behavior... Whereas before an input that was not present in the options block would just result in something not being present in the output array, this version will push an empty entry (ie. undefined) onto the array. We didn't have before, and dont have now any tests for this, and I'm not actually sure what the desired behavior should be

kaspermarkus: So thinking through it, it seems to me that we would want the original behavior, ie., that the following transform:

```
```
{
    "detections": {
        "hasMouse": "supported",
        "hasKeyboard": "supported",
        "hasTrackpad": "not supported",
        "hasHeadtracker": "not supported",
        "hasTelephone": "supported"
    }
}
```

File: src/framework/core/js/ModelTransformationTransforms.js (hunk 4 above)
Comment (kaspermarkus) [inline src/framework/core/js/ModelTransformationTransforms.js]: Good catch antranig!! I've updated the framework to now always copy the transformSpec object before passing it on to the individual inversion function.

In terms of the multiInput transform, it is actually on purpose that I'm not calling the inversePath copying for all of them. It's being used in linearScale, because this takes as one of the inputs `input` (or `inputPath`) that mimic the behavior of the same variables in the standardInputTransformFunction. But this doesn't hold up for things like the binaryOp transform, which doesn't accept `input` and `inputPath`.

The reason linearScale is a multiInputTransformFunction and not a standardInputTransformFunction is that it uses some extra variables (scale and offset) for which we can get the path equivalent as well as default values.

Comment (amb26) [inline src/framework/core/js/ModelTransformationTransforms.js]: Hi @kaspermarkus - I think this then shows up a hole in our "type system" since hopefully it is not completely accidental that linearScale uses an input named "input" and uses it in exactly the same sense as a standardInputTransform. What I propose is that to signal this fact we then turn linearScale into something which inherits *both* from standardInputTransform and multiInputTransform (together with suitable support in the core transformer to recognise what this means) which then would hopefully mean that we can get rid of the unsightly manual call to copyInputPaths in its body.

Comment (kaspermarkus) [inline src/framework/core/js/ModelTransformationTransforms.js]: good point! I've updated the code accordingly

**Comment by kaspermarkus** [comment]
OK - should be ready for another round of review

File: src/framework/core/js/ModelTransformation.js [hunk 5]
```diff
@@ -382,6 +383,11 @@ var fluid = fluid || fluid_2_0_0;
     };
     // unsupported, NON-API function
     fluid.model.transform.handleInvertStrategy = function (transformSpec, transform, transformOpts) {
+        transformSpec = fluid.copy(transformSpec);
```
amb26: This now copies the spec twice in the case it is standard

kaspermarkus: Not sure I follow this comment... Where is it also being copied?

amb26: You copy it once here and then again in invertPaths

kaspermarkus: are you sure your diff is up to date? I've deleted the old fluid.copy from invertPaths on the old line 164

amb26: Oops, quite right, sorry! I misread the diff ...

amb26:
This pull request should never have been merged, it has failing tests

src/framework/core/js/ModelTransformationTransforms.js
amb26: This change is faulty and caused https://issues.fluidproject.org/browse/FLUID-5907

tests/framework-tests/core/js/ModelTransformationTests.js
```diff
@@ -59,29 +59,68 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
     };
 
-    fluid.tests.transforms.testOneTransform = function (message, model, transform, method, expected, transformWrap) {
-        if (transformWrap) {
-            transform = fluid.tests.transforms.wrapTransform(transform);
-        }
-        var transformed = fluid.model.transform(model, transform);
-        jqUnit[method].apply(null, [message, expected, (transformWrap ? transformed.value : transformed) ]);
+    fluid.tests.transforms.testOneTransform = function (test) {
+        var transformed = fluid.model.transform(test.model, test.transform);
```
amb26: This change is responsible for the test failures reported at http://lists.idrc.ocad.ca/pipermail/fluid-work/2016-May/009969.html

PR #703, opened 2016-04-20
FLUID-5886: Fix for failure to deduplicate model listeners by namespace

amb26:
at a single applier. This rebalances the implementation of change events and standard events, with the former now implemented in terms of the latter. This introduces a new options-based API for event.addListener and an improved fluid.isPlainObject.

### DIFF
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -1341,9 +1341,9 @@ var fluid = fluid || fluid_2_0_0;
     /** Construct an "event firer" object which can be used to register and deregister
      * listeners, to which "events" can be fired. These events consist of an arbitrary
      * function signature. General documentation on the Fluid events system is at
-     * http://wiki.fluidproject.org/display/fluid/The+Fluid+Event+System .
+     * http://docs.fluidproject.org/infusion/development/InfusionEventSystem.html .
      * @param {Object} options - A structure to configure this event firer. Supported fields:
-     *     {String} name - a name for this firer
+     *     {String} name - a readable name for this firer to be used in diagnostics and debugging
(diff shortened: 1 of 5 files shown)


src/framework/core/js/Fluid.js
```diff
@@ -1354,24 +1357,44 @@ var fluid = fluid || fluid_2_0_0;
...
                     return;
                 }
+                if (fluid.isPlainObject(listener, true)) {
+                    record = listener;
+                    listener = record.listener;
+                    namespace = record.namespace;
+                    priority = record.priority;
+                    listenerId = record.listenerId;
+                } else {
+                    record = {};
```
cindyli: Is the `else` block necessary as this situation will be handled at line 1391.

src/framework/core/js/Fluid.js
```diff
@@ -1354,24 +1357,44 @@ var fluid = fluid || fluid_2_0_0;
...
                 }
                 if (!listener) {
                     return;
                 }
+                if (fluid.isPlainObject(listener, true)) {
+                    record = listener;
+                    listener = record.listener;
+                    namespace = record.namespace;
+                    priority = record.priority;
+                    listenerId = record.listenerId;
```
cindyli: Does the `softNamespace` also need to be extracted like what you've done for other listener parts? What I mean is to add a line of ```softNamespace = record.softNamespace```.

src/framework/core/js/Fluid.js
amb26: Great catches both, @cindyli - thanks!

cindyli:
Merged at ef4324149aba47693d85f61357feecf6b127a9bf

PR #705, opened 2016-04-22
FLUID-5892: Refactor fluid.slider to be able to use native or jQueryUI-based slider

waharnum:
This PR adds the option to use the native HTML `input[type=range]` slider as well as the jQuery UI-based slider in the Preferences Framework, per recent discussion with @cindyli, @michelled & @amb26.

I believe this will allow us to implement https://issues.fluidproject.org/browse/FLUID-5698 and some other improvements.

Applying styles to the native slider appears to require some browser-specific pseudoclasses: https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/


src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -113,65 +113,65 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
-            slider: {
-                expander: {
-                    "this": "{that}.container",
-                    method: "slider",
-                    args: ["{that}.combinedSliderOptions"]
-                }
+            }
+        },
+        modelRelay: {
+            target: "value",
```
amb26: This is rather awkward - especially with respect to @kaspermarkus upcoming fix which will make this transform invertible. If you change the type of the field, you should lens it to a different path.

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -113,65 +113,65 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     fluid.defaults("fluid.slider", {
```
amb26: Rather than destroy our existing impl, we should factor it into three grades - fluid.slider containing common definitions, fluid.slider.native and fluid.slider.jQuery containing the addons necessary to construct the two different varieties of slider.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
```diff
@@ -27,45 +27,48 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 
         jqUnit.test("Test Init", function () {
             jqUnit.expect(8);
-            var textfieldSlider = fluid.tests.textfieldSlider.createTextfieldSlider({model: {value: 15}});
-            jqUnit.assertEquals("Slider value is set to input value", 15, $(".flc-textfieldSlider-slider").slider("value"));
+            var that = fluid.tests.textfieldSlider.createTextfieldSlider({model: {value: 15}});
```
amb26: After we have factored out the two varieties of slider, this can become two tests

src/components/textfieldSlider/js/TextfieldSlider.js
waharnum: Pretty easy and allows us to support backwards compatibility with browsers that don't have full support for `input[type=range]`. Thoughts on what should be the default for the reference implementation?

amb26: The "default" should be the native HTML5 widget. However, we can apply this default in a cool way, by using the "makeGradeLinkage" facility targetted at the base grade - http://docs.fluidproject.org/infusion/development/IoCAPI.html#fluid-makegradelinkage-linkagename-inputnames-outputnames- - in this way, the default can be easily overridden on a global or local basis.

waharnum: @amb26 - can you expand a little on how the default can be overriden? I assume we'd want something similar to a configuration variable like this pseudocode:

```
if(PREFERNATIVEWIDGETS) {
fluid.makeGradeLinkage("fluid.textfieldSliderNativeWidget", ["fluid.textfieldSlider"],["fluid.textfieldSliderNative"]);
} else {
fluid.makeGradeLinkage("fluid.textfieldSliderNativeWidget", ["fluid.textfieldSlider"],["fluid.textfieldSliderJQuery"]);
}
```

Then we follow that pattern for other components where we need the preference for native / jQuery to be used?

amb26: Yes, something like that. Probably "fluid.textfieldSliderVariety" or so would be more intelligible. The default, written in the framework, would read
fluid.makeGradeLinkage("fluid.textfieldSliderVariety", ["fluid.textfieldSlider"], "fluid.textfieldSliderNative");

Then a user who wants to override this can issue

fluid.makeGradeLinkage("fluid.textfieldSliderVariety", ["fluid.textfieldSlider"], "fluid.textfieldSliderJQuery");

in their own file, before they construct any widgets.
The framework can also support variants of this which are just scoped to particular areas of the component tree, but the user would have to currently "roll their own distribution" for this since the signature of fluid.makeGradeLinkage doesn't support this yet.

waharnum: Got it - the key I was missing was to have makeGradeLinkage target the same namespace, so that it's overridable by the end user. Thanks!

waharnum: @amb26 thoughts on appropriate way to manage something like a global toggle ("use native widgets" vs. "use jQuery widgets")? As I work my way through UIO and the Preferences Framework I'm finding several components that need to have a sense of the preference for the native vs. jQuery slider (primarily in terms of what HTML template they use). It doesn't seem appropriate to expect a user to override each one by one.

It seems at first glance reading the documentation on context awareness that it might be appropriate to make use of it - there are numerous components that need to respond to a context of "use native widgets" vs. "use jQuery widgets". Or is that more complexity than is necessary in this scenario? I'm conscious that whatever the approach taken is, it needs to avoid anything that seems like magic or central control fragility (hence my hesitation around the very idea of a "global toggle").

waharnum:
I've refactored fluid.textFieldSlider to also use context-awareness rather than grade linkages. This means an end user only has to make one change rather than two to switch between native or jQueryUI-based slider; this seemed appropriate, per discussion with @amb26, as grade linkage vs. context-awareness have some overlap in their purpose, and I think it makes sense to only use one in this context.

waharnum:
I think this is ready for review for incorporation into Infusion core - I've created a parallel documentation branch to update the documentation at https://github.com/waharnum/infusion-docs/tree/FLUID-5892

@jobara had offered to review (and his experience with Preferences / UIO would be very welcome), so tagging him here.

waharnum:
A quick note to say that if people want to see the native sliders in action, you can check out the implementation on my personal site at http://www.alanharnum.ca/accessibility/inclusion/one-size-fits-one/dogfooding/2016/05/17/ui-options-implemented.html - I used a custom build of Infusion from my branch for this.

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -39,6 +38,23 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
         },
+        contextAwareness: {
+            sliderVariety: {
+                checks: {
+                    nativeHTML: {
+                        contextValue: "{fluid.prefsWidgetType}",
+                        equals: "nativeHTML",
+                        gradeNames: "fluid.textfieldSlider.nativeHTML"
+
+                    },
```
jobara: Do we actually have to check for native? Could we set the native slider as the default in the component and just override it with the jQuery one if the contextValue is set to jQueryUI?

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -110,72 +152,136 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
         range: {}, // should be used to specify the min, max range e.g. {min: 0, max: 100}
-        selectors: {
-            thumb: ".ui-slider-handle"
-        },
         members: {
             combinedSliderOptions: {
                 expander: {
                     funcName: "fluid.slider.combineSliderOptions",
                     args: ["{that}.options.sliderOptions", "{that}.options.range"]
                 }
```
jobara: combinedSliderOptions seems to be something specifically needed for the jQuery cases. Is that true, and if so, does it make sense to move it to that grade?

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -110,72 +152,136 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                    }
+                }
+            },
+            invokers: {
+                setSliderValue: {
+                    "this": "{that}.slider",
+                    "method": "slider",
+                    args: ["value", "{arguments}.0"]
+                },
+                setSliderAria: {
```
jobara: Maybe this should be renamed to indicated that it is only setting the "aria-valuenow" attribute. It is a bit confusing as is, because of all the other aria attributes that are set onCreate.

src/components/textfieldSlider/js/TextfieldSlider.js
```diff
@@ -110,72 +152,136 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                    "method": "attr",
+                    args: ["aria-valuenow", "{arguments}.0"]
+                },
+                setModel: {
+                    changePath: "value",
+                    value: "{arguments}.1.value"
+                }
+            },
+            listeners: {
+                // This can be removed once the jQuery UI slider has built in ARIA
```
jobara: I'm somewhat shocked that this isn't the case already. Although I tried there [demo](http://jqueryui.com/slider/) and it still seems to be missing.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -217,7 +257,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             var iframeWindow = that.iframe[0].contentWindow;
             that.iframeDocument = iframeWindow.document;
 
-            that.jQuery = iframeWindow.jQuery;
+            // Use the iframe's own jQuery if it's loaded, otherwise use the
+            // parent document
+
+            var iframeWindowHasJQuery = iframeWindow.jQuery !== undefined;
+            that.jQuery = iframeWindowHasJQuery ? iframeWindow.jQuery : $;
```
jobara: Is there a case where iframeWindow.jQuery is legitimately falsey? Could we instead write this as 
```javascript
that.jQuery = iframeWindow.jQuery || $;
```

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
```diff
@@ -21,76 +21,160 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+        // Returns the value of either a native input[type=range] or a jQuery slider
+        fluid.tests.textfieldSlider.getSliderValue = function (slider) {
+
+            var sliderValue;
+            try {
+                sliderValue = slider.slider("value");
+            } catch (e) {
+                sliderValue = slider.val();
+            }
```
jobara: Instead of using a try catch could we do something like:

```javascript
return slider.slider ? slider.slider("value") : slider.val();
```

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
```diff
@@ -21,76 +21,160 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        };
+
+        // Sets the value of either a native input[type=range] or a jQuery slider
+        fluid.tests.textfieldSlider.setSliderValue = function (slider, newValue) {
+
+            try {
+                slider.slider("value", newValue);
+            } catch (e) {
+                fluid.changeElementValue(slider, newValue);
+            }
```
jobara: similar to above

src/components/textfieldSlider/js/TextfieldSlider.js
waharnum: We could do something like this (it still requires a separate `fluid.textfieldSlider.nativeHTML` grade but is less verbose):

```
        contextAwareness: {
            sliderVariety: {
                checks: {
                    jQueryUI: {
                        contextValue: "{fluid.prefsWidgetType}",
                        equals: "jQueryUI",
                        gradeNames: "fluid.textfieldSlider.jQueryUI"
                    }
                },
                defaultGradeNames: "fluid.textfieldSlider.nativeHTML"
            }
        },
```

That's maybe a clearer expression of the component intention than registering and using `fluid.prefsWidgetType: "nativeHTML"` - how does that sound?

src/components/textfieldSlider/js/TextfieldSlider.js
waharnum: Yeah, I think this makes sense - looking at it more closely `combinedSliderOptions` is necessary because the jQuery UI slider requires a single options block when being created. There's no need for it in the native implementation. Will refactor.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
```diff
@@ -99,8 +183,26 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
                     value: 1
```
jobara: Shouldn't this also source its value from the model?

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
amb26: The overall strategy isn't right. We should no more have implementation switches based on conditionals in code in our tests than we should in the real implementation. This should be parameterised inside a grade with an invoker whose implementation comes from a grade, just as in the component impl itself.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
amb26: Remember, testing code is just code :)

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
```diff
@@ -99,8 +183,26 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+                    min: -5,
+                    max: 5
+                },
+                model: {
+                    value: 1
+                }
+            });
+            fluid.tests.textfieldSlider.testInputField("aaa", 1, that);
+            fluid.tests.textfieldSlider.testInputField(null, 0, that);
+        });
```
jobara: These two tests seem quite duplicated with the primary difference being which version of the slider to use. Maybe parameterizing this would help.

src/components/textfieldSlider/js/TextfieldSlider.js
waharnum: The jQuery UI slider also still isn't capable of functioning in an iframe without a separately-included jQuery. 😿

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
waharnum: This is probably safer. Or, per @amb26's comments below about implementation switches based on conditionals, maybe it should be parameterised in the `fluid.prefs.separatedPanel.renderIframe` based on the same contextAwareness checks.

tests/framework-tests/preferences/js/PanelsTests.js
```diff
@@ -1286,6 +1321,58 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        },
+        modules: [{
+            name: "Test the text sizer settings panel (jQuery slider)",
+            tests: [{
+                expect: 4,
+                name: "Test the rendering of the text size panel",
+                sequence: [{
+                    func: "fluid.tests.switchToJqueryWidgets",
+                    args: ["{textSizePanelJQuery}"]
+                },  {
```
jobara: It seems like a lot of the test sequences are duplicated. I think you factor these out so that the sequences are stored as an array and passed in here. The difference I noticed between them is with the first element in the sequence which sets which context to use. I think you could rewrite the switch function to something like:

```javascript
fluid.tests.switchWidgetTo = function (widgetType, testPanel) {
    fluid.contextAware.makeChecks({
        "fluid.prefsWidgetType": {
            value: widgetType
        }
    });

    testPanel.events.onSwitchedWidget.fire();
};
```

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
waharnum: Perhaps to main API parallels a `setSliderValue` invoker should be added to the `fluid.slider.native` grade, since one already exists in `fluid.slider.jQuery`? The tests could then simply use the invoker on the components being tested.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
waharnum: Perhaps to main API parallels a `setSliderValue` invoker should be added to the `fluid.slider.native` grade, since one already exists in `fluid.slider.jQuery`? The tests could then simply use the invoker on the components being tested.

tests/manual-tests/framework/preferences/fullPage/index.html
```diff
@@ -87,6 +87,7 @@
                     </li>
                     <li>Use the "save and apply" button to apply the preview styles to Preferences Editor</li>
                     <li>Use the "reset" button to reset the styles</li>
+                    <li>Be sure to test with the <a href="index-jQueryUI.html">jQuery UI widgets version as well.</a></li>
```
jobara: I think that it's good to have the link here, but I'm also thinking that it might be hard to discover the alternate versions of these tests because of the use of index.html and the fact that testers may not read the instructions. I wonder if separating the versions out into separate sub directories would help with this and keeping the links the instructions as well.

tests/manual-tests/framework/preferences/assortedContent/index-jQueryUI.html
```diff
@@ -0,0 +1,299 @@
+<!DOCTYPE html>
```
jobara: Instead or as well as making using this manual test, maybe creating a UIO or PrefsFramework example would be good. https://github.com/fluid-project/infusion/tree/master/examples

jobara: The idea with examples is that they are illustrative of how to do something in particular. In this case it would be switching to the jQuery widgets. And the broader hope is to move or some link these to concrete documentation/tutorials in our docs site.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
jobara: Seems reasonable. I wonder if we should add the invoker to the base grade and supply the concrete implementation to the native and jQuery variants.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
waharnum: @jobara - The expected values in the `testInputField` calls when testing invalid values, you mean?

They use direct values in the current test code (I'm guessing because of the 'pass null, expect 0' test):
https://github.com/fluid-project/infusion/blob/master/tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js#L102
https://github.com/fluid-project/infusion/blob/master/tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js#L103

tests/manual-tests/framework/preferences/fullPage/index.html
waharnum: I could also rename the `index.html` in each directory to `index-nativeHTML.html`, so that someone doing the tests would see more obviously there were two testing files. This would require the least reorganization of the file structure.

tests/component-tests/textfieldSlider/js/TextfieldSliderTests.js
waharnum: I thought about this more and ended up implementing it test grades - it's not necessary for the components themselves & I think it's better to encourage JS get/set actions against the component model that's bound to the slider.

tests/manual-tests/framework/preferences/assortedContent/index-jQueryUI.html
waharnum: I've added a jQuery UI version of the prefsFramework demo.

waharnum:
@jobara I think this one is ready for another look.

jobara:
@waharnum instead of adding main.css to the demo, can you use one of the default jQuery UI theme that comes with Infusion https://github.com/waharnum/infusion/tree/bc70d21602a42be7300aa4186ef351a585b5782b/src/lib/jquery/ui/css/default-theme

jobara:
@waharnum also we should avoid duplicating all the binary and any other resources shared by the two prefsFramework demos. I'm also not sure it needs to be in a demo, i think an example is probably good enough.

waharnum:
Where would the example "live"? Would it be part of another demo? Just wondering about how to actually put that into action without it being a full demo.

jobara:
@waharnum it can live as a new example here https://github.com/fluid-project/infusion/tree/master/examples/framework/preferences

waharnum:
It looks like there are some circumstances where even with the use of native widgets (the conditional adjusters present in the `examples`), a separate include of jQuery in the iframe for a separated panel is necessary. You can see the issue in action at https://github.com/fluid-project/infusion/pull/705/commits/aed69b986c837c1acde10531861c072c7c582a55 - as far as I can tell, something about the creation being delayed by the conditional causes a context switch from the iframe to the main document, and creation of the slider components that appear only when the conditional adjustors are turned on fails with this message:

`Instantiation of view component with type fluid.textfieldSlider failed, since selector "body .mpe-increasing .mpe-increasing-magFactor .mpe-slider" did not match any markup in the document` (the markup does exist in the document, but it exists within the iframe - I believe I've confirmed this is not a sequencing issue with templates not being loaded in time, etc)

I don't think this is inherent to modern browsers (it's possible to locate elements within the iframe with the parent jQuery), but it seems bound up in some complexity I haven't been able to fully figure out between the behaviour of viewComponents, the renderer, the DOM binder, and the way the component trees and markup of the Prefs Editor are built and rebuilt. As best I can narrow it down it may be something that's happening in the `fluid.prefs.subPanel.resetDomBinder` function, but some initial mucking about in there didn't reveal an obvious fix - I can get the component creation code to stop throwing errors by commenting out lines [153](https://github.com/fluid-project/infusion/blob/63b8a53442f712d00dfbb684702ee1f7df2b5f55/src/framework/preferences/js/Panels.js#L153) and [154](https://github.com/fluid-project/infusion/blob/63b8a53442f712d00dfbb684702ee1f7df2b5f55/src/framework/preferences/js/Panels.js#L154), but while the component itself seems to get created if that's done, it doesn't seem to get affiliated/bound with the actual document markup - no errors, but the UI elements visible to the user (and the underlying markup) do not get altered when the component's model is altered, and vice-versa.

amb26:
Hi @waharnum - I tried running all the prefs framework tests both at the tip of this branch and at aed69b9 and couldn't get them to fail - how should I be able to see the failure?

waharnum:
I've merged the latest from `fluid-project/infusion` master, fixed a small merge conflict related to template names, used the new ESLint linting (really nice) and run the full test suite. Everything looks good.

colinbdclark:
@jobara Is this something you can re-review and merge if you think it's looking good?

jobara:
@colinbdclark would probably need someone to review the unit tests I've contributed.

examples/framework/preferences/conditionalAdjusters-singlePanel/jQueryUI.html
```diff
@@ -0,0 +1,99 @@
...
+
+            fluid.contextAware.makeChecks({
+                "fluid.prefsWidgetType": {
+                    value: "jQueryUI"
+                }
+            });
+
+            $(document).ready(function () {
+                example.initPrefsEditor(".flc-prefsEditor-separatedPanel");
+            })
```
jobara: Rather than have this `$(document).ready()` call, we should move these to a script tag in the appropriate position in the DOM after all it's required markup has loaded.

examples/framework/preferences/conditionalAdjusters/js/auxiliarySchema.js
```diff
@@ -19,7 +19,26 @@ var example = example || {};
...
     fluid.defaults("example.auxSchema", {
-        gradeNames: ["fluid.prefs.auxSchema"],
+        gradeNames: ["fluid.prefs.auxSchema", "fluid.contextAware"],
+        listeners: {
+            "onCreate.logInfo": {
+                this: "console",
+                method: "log",
+                args: "{that}"
+            }
+        },
```
jobara: we should remove this console log

examples/framework/preferences/conditionalAdjusters-singlePanel/js/auxiliarySchema.js
```diff
@@ -123,4 +131,31 @@ var example = example || {};
         }
     });
 
+    // Context-aware mixin for jQueryUI slider
+    fluid.defaults("example.auxSchema.nativeHTML", {
```
jobara: Is the comment wrong, or are we missing one of the mixins?

tests/framework-tests/preferences/js/CompositePanel.js
```diff
@@ -0,0 +1,188 @@
+/*
+Copyright 2013-2016 OCAD University
```
cindyli: 2016 rather than 2013-2016?

tests/framework-tests/preferences/js/CompositePanel.js
```diff
@@ -0,0 +1,188 @@
+/*
```
cindyli: Shall this file be named "CompositePanelTests.js" or "CompositePanelTestsGrades.js" to be in sync with other test files?

tests/framework-tests/preferences/js/CompositePanel.js
jobara: Much of this file was moved from BuilderTests.js

jobara:
This PR has been continued in https://github.com/fluid-project/infusion/pull/724

PR #706, opened 2016-04-25
FLUID-5887: Resolution and checks for double initialisation of model machinery

amb26:
simplification and correction of grade merging algorithm

### DIFF
--- a/src/framework/core/js/DataBinding.js
+++ b/src/framework/core/js/DataBinding.js
@@ -416,7 +416,7 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         };
         var parsed; // resolve ref into context and modelSegs
         if (typeof(ref) === "string") {
-            if (ref.charAt(0) === "{") {
+            if (fluid.isIoCReference(ref)) {
                 parsed = fluid.parseModelReference(that, ref);
                 var modelPoint = parsed.segs.indexOf("model");
                 if (modelPoint === -1) {
@@ -783,7 +783,7 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
 
     fluid.parseImplicitRelay = function (that, modelRec, segs, options) {
         var value;
-        if (typeof(modelRec) === "string" && modelRec.charAt(0) === "{") {
+        if (fluid.isIoCReference(modelRec)) {
             var parsed = fluid.parseValidModelReference(that, "model reference from model (implicit relay)", modelRec, true);
             if (parsed.nonModel) {
                 value = fluid.getForComponent(parsed.that, parsed.segs);
--- a/src/framework/core/js/DataBinding.js
+++ b/src/framework/core/js/DataBinding.js
@@ -853,6 +853,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     };
 
     fluid.establishModelRelay = function (that, optionsModel, optionsML, optionsMR, applier) {
+        var shadow = fluid.shadowForComponent(that);
+        if (!shadow.modelRelayEstablished) {
+            shadow.modelRelayEstablished = true;
+        } else {
+            fluid.fail("FLUID-5887 failure: Model relay initialised twice on component", that);
+        }
         fluid.mergeModelListeners(that, optionsML);
 
         var enlist = fluid.enlistModelComponent(that);
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -348,6 +348,10 @@ var fluid = fluid || fluid_2_0_0;
             fluid.isArrayable(totest) ? "array" : "object";
     };
 
+    fluid.isIoCReference = function (ref) {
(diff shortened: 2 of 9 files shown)


src/framework/core/js/Fluid.js
```diff
@@ -1627,35 +1635,24 @@ var fluid = fluid || fluid_2_0_0;
...
-        for (var i = gradeNames.length - 1; i >= 0; -- i) {
+    // unsupported, NON-API function
+    // Recursively builds up "gradeStructure" in first argument. 2nd arg receives gradeNames to be resolved, with stronger grades at right (defaults order)
+    // builds up gradeStructure.gradeChain pushed from strongest to weakest (reverse defaults order)
+    fluid.resolveGradesImpl = function (gs, gradeNames) {
+        gradeNames = fluid.makeArray(gradeNames);
+        for (var i = gradeNames.length - 1; i >= 0; -- i) { // from stronger to weaker
             var gradeName = gradeNames[i];
             if (gradeName && !gs.gradeHash[gradeName]) {
                 var isDynamic = gradeName.charAt(0) === "{";
```
cindyli: Shall `gradeName.charAt(0) === "{"` be replaced by `fluid.isIoCReference()`?

Other scripts such as DataBinding.js, FluidIoC.js and IoCTestUtils.js have occurrences of checking IoC references. Shall those be replaced as well?

src/framework/preferences/js/Panels.js
```diff
@@ -281,8 +281,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
      * Only use in cases where the instatiated component cannot be used.
      */
     fluid.prefs.compositePanel.prefetchComponentOptions = function (type, options) {
-        var baseOptions = fluid.getGradedDefaults(type, fluid.get(options, "gradeNames"));
-        return fluid.merge(baseOptions.mergePolicy, baseOptions, options);
+        var baseOptions = fluid.getMergedDefaults(type, fluid.get(options, "gradeNames"));
+        // TODO: awkwardly, fluid.merge is destructive on each argument!
```
cindyli: Can you elaborate on `destructive`: What does `fluid.merge` do on arguments?

src/framework/core/js/Fluid.js
```diff
@@ -808,10 +812,14 @@ var fluid = fluid || fluid_2_0_0;
      * @return The supplied argument, recursively frozen
      */
     fluid.freezeRecursive = function (tofreeze) {
-        fluid.each(tofreeze, function (value) {
-            fluid.freezeRecursive(value);
-        });
-        return fluid.isPlainObject(tofreeze) ? Object.freeze(tofreeze) : tofreeze; // IE11 crashes on freeze of non-object
+        if (fluid.isPlainObject(tofreeze)) {
```
cindyli: If you merge the master branch in, `fluid.isPlainObject()` tests both object and array. Should the second argument added with `true` if you don't wanna freeze arrays and `Object.freeze()` is also not able to.

src/framework/core/js/DataBinding.js
```diff
@@ -853,6 +853,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     };
 
     fluid.establishModelRelay = function (that, optionsModel, optionsML, optionsMR, applier) {
+        var shadow = fluid.shadowForComponent(that);
```
cindyli: Shall a test be added to cover the case that model relay is not merged/established more than once?

src/framework/core/js/Fluid.js
amb26: Object.freeze() is perfectly well able to freeze arrays - that is, plain ones. It fails to freeze "exotic" kinds of arrays such as those descended from TypedArray. Recall that typeof([]) === "object" in JS ...

src/framework/core/js/DataBinding.js
amb26: Not sure what you mean "not more than once" - it does check that it is established any more times than once, and clearly it will be established *at least* once since this method has started to execute. The existing tests cases triggering this route are sufficient since numerous of them started to fail under the old framework when this check was introduced.

src/framework/preferences/js/Panels.js
amb26: It will tend to corrupt them in various hard-to-predict ways, for example by adding properties whose value is `undefined` or replacing IoC references with their targets. As our docs say, fluid.merge should not be called by users - http://docs.fluidproject.org/infusion/development/OptionsMerging.html . In the next framework rewrite we might improve this. In any case, this update was helpful in revealing this fact - without the fluid.copy(), we now get an error triggered as fluid.merge tries to corrupt the defaults record which is now immutable.

src/framework/core/js/Fluid.js
amb26: Good idea!

amb26:
Updated with some uses of fluid.isIoCReference() - the remaining sites are too performance-critical or an awkward fit

cindyli:
Merged at 63b8a53442f712d00dfbb684702ee1f7df2b5f55

PR #709, opened 2016-05-04
FLUID-5901: Fix for jqUnit.assertDeepEq comparing all primitive values as equal

amb26:
(following QUnit.propEqual) and fixes for various tests around the framework which were faultily relying on this behaviour. Commented out tests for FLUID-5220 which this fix exposes numerous other problems with, and reopened its JIRA.


tests/test-core/testTests/js/jqUnitTests.js
```diff
@@ -0,0 +1,70 @@
...
+    fluid.registerNamespace("fluid.tests.jqUnit");
+    
+    jqUnit.module("Conventional jqUnit tests");
+    
+    fluid.tests.jqUnit.failIsPass = false;
+    
+    // We have to use this insane scheme for subverting QUnit's accounting rather than a teardown function,
+    // since we need to beat qunit-composite.js own logging function. When will the "industry" learn that
+    // integration is the first priority?
+    QUnit.log(function (data) {
```
cindyli: Is the subverting using `QUnit.log()` needed as `fluid.tests.failingTeardown()` converts assertion results too?

tests/test-core/jqUnit/js/jqUnit.js
```diff
@@ -172,11 +172,19 @@ var jqUnit = jqUnit || {};
         },
 
         assertDeepEq: function (msg, expected, actual) {
-            QUnit.propEqual(actual, expected, processMessage(msg));
+            if (fluid.isPrimitive(expected)) {
```
cindyli: Shall `actual` be verified as `isPrimitive` too? Otherwise, tests that have `expected` as an object but `actual` as a primitive would pass while they should actually fail. An example:

```
jqUnit.assertDeepEq("True is not deep equal to an object", {}, true);
```

The same comment for line 183.

tests/test-core/testTests/js/jqUnitTests.js
amb26: It's because qunit-composite.js uses a completely different strategy to the standard QUnit UI for accounting for test failures: https://github.com/fluid-project/infusion/blob/master/tests/lib/qunit/addons/composite/qunit-composite.js#L79 . If we just use failingTeardown, the main QUnit UI works fine, but the "failures" are rendered as real failures in the iframe-based all-tests.html driver.

tests/test-core/jqUnit/js/jqUnit.js
amb26: Extremely good point!

tests/test-core/testTests/js/jqUnitTests.js
cindyli: I see. Thanks for the explanation. Could you add it into the code comment too? Thanks.

amb26:
I've updated the pull with the suggested fix plus test cases. Also, amazingly, I discovered an extremely ancient "jqUnitTests.js" which had been implemented quite a while ago but neglected because it could not be run as part of the all-tests.html master runner since we had not implemented the "fail is pass" infrastructure in this pull. I copied up the tests in it into the new jqUnit-tests.js and deleted the old files.

cindyli:
Merged at 35d924eafc28ff601c2f6f69123a25d32a360362

PR #711, opened 2016-05-10
FLUID-5904: Fix for corruption if exception is thrown from expander

amb26:


jobara:
Merged at 78d75841a6d6305b11f3ac8ca1b2b66fc67e4c22

PR #710, opened 2016-05-04
FLUID-5891, FLUID-5300, FLUID-5608: Cleaned up valuemapper syntax and tests

kaspermarkus:
Note that this pull requests depends on: https://github.com/fluid-project/infusion/pull/701 (now merged)

PR #715, opened 2016-05-25
FLUID-5051: Added tests for collectInputPaths for transforms

kaspermarkus:
Based off of https://github.com/fluid-project/infusion/pull/710

Note that the "collectInputPath" function currently doesn't ensure that the result is an array set (ie. with only one copy of each path). This means that the same path can be present multiple times in the output - even from a single transform (as can be seen here: https://github.com/fluid-project/infusion/compare/master...kaspermarkus:FLUID-5051?expand=1#diff-d453881aa3f26a20a139679dba71e07eR1492). I haven't modified this behavior, as I figured it might be useful for deciding the importance of a specific path (ie. how many time it is being used). I can change this behavior if requested


amb26:
Could you close and regenerate this patch against current master now that your valueMapper update has gone in? Thanks

PR #718, opened 2016-06-12
FLUID-5810: Fix for erratic failure of tests. 

amb26:
QUnit patched to support new "testsArriving" config member, which is now used by IoC Testing Framework


colinbdclark:
This is so awesome! We can now run our Infusion tests on the continuous integration server! Thanks, @amb26.

PR #725, opened 2016-07-06
FLUID-5924: Fixed native sliders in Firefox

jobara:
Fixed an issue where the native sliders in Firefox would not initialize with the correct value.

Also fixed a linting issue.

https://issues.fluidproject.org/browse/FLUID-5924


cindyli:
Merged @ 3f148ee5baa5ba2f36195a390d2c2afd52c9e232

PR #726, opened 2016-07-07
Responsive design

styki:
I am working on the responsive design of the UI options as part of the Google Summer of Code program. This is the work that I have done for the past few weeks. I used the mock-ups (https://wiki.fluidproject.org/display/fluid/%28Floe%29+User+Interface+%28Learner%29+Options+Mobile+and+Responsive+Design) as a guideline for how it is supposed to look for small screens (max-width: 640px). 

The changes are tested in Chrome and using the UI options demo. It is the only demo in which I have added all new files. There might be quite a few bugs in the other demos, which I will resolve later.

Known bug:
The change between small and large screen version is triggered only when the page is first loaded and everything is rendered. If you resize the window is changed, it does not switch to the other model. I think the reason for that is once the components are build, they become immutable, so I can't change the gradeNames.

I will be very thankful if you can give me some feedback on how to improve the implementation and a better way to construct the components if there is one.


src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,153 @@
...
+        },
+        invokers: {
+            setModel: {
+                changePath: "stringValue",
+                value: "{arguments}.0.target.value"
+            }
+        }
+    });
+
+    fluid.defaults("fluid.button", {
```
jobara: The implementation seems to be for a specific type of button, one that increments/decrements a value. `fluid.button` seems like it should be for a generic button. This should probably be called something like `fluid.stepper.button`, `fluid.button.stepper` or something like that.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,153 @@
...
+        gradeNames: ["fluid.viewComponent"],
+        range: {}, // should be used to specify the min, max range e.g. {min: 0, max: 100}
+        invokers: {
+            setModel: {
+                changePath: "value",
+                value: {
+                    expander: {
+                        funcName: "fluid.buttonCalculateValue",
+                        args: ["{that}.options.incrementCoefficient", "{that}.options.buttonOptions.stepMultiplier", "{that}.model.value", "{that}.options.range.min", "{that}.options.range.max"]
```
jobara: The `incrementCoefficient` and `stepMultiplier` options should have default values.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,153 @@
...
+
+    fluid.defaults("fluid.button", {
+        gradeNames: ["fluid.viewComponent"],
+        range: {}, // should be used to specify the min, max range e.g. {min: 0, max: 100}
+        invokers: {
+            setModel: {
+                changePath: "value",
+                value: {
+                    expander: {
+                        funcName: "fluid.buttonCalculateValue",
```
jobara: `fluid.buttonCalculateValue` should be an invoker as well.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,153 @@
...
+        listeners: {
+            "onCreate.bindClickEvt": {
+                "this": "{that}.container",
+                "method": "click",
+                "args": "{that}.setModel"
+            }
+        }
+    });
+
+    fluid.buttonCalculateValue = function (value, stepMultiplier, modelValue, min, max) {
```
jobara: This should be in the same namespace as the grade.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,153 @@
...
+        modelValue = Math.round(modelValue * stepMultiplier);
+        min = Math.round(min * stepMultiplier);
+        max = Math.round(max * stepMultiplier);
+        if ((modelValue + value) < min) return min/stepMultiplier;
+        else if ((modelValue + value) > max) return max/stepMultiplier;
+        else return (modelValue + value)/stepMultiplier;
```
jobara: Please use "{}" for these conditionals

jobara: I'm not sure I follow the logic or reason for the logic in this function. For example why redefine the max and min values? Why the need for rounding? What if I want a step that is decimal value. I'm thinking it would be easier to have two methods:

1. A "setValue" method for setting a specific value
2. A "step" method which performs a step

Also if the button cares about a max and min value, it should also use a model relay to ensure that the value stays within range.

src/framework/preferences/js/Panels.js
```diff
@@ -767,13 +818,25 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        protoTree: {
+            label: {messagekey: "lineSpaceLabel"},
+            narrowIcon: {messagekey: "lineSpaceNarrowIcon"},
+            wideIcon: {messagekey: "lineSpaceWideIcon"},
+            multiplier: {messagekey: "multiplier"},
+            lineSpaceDescr: {messagekey: "lineSpaceDescr"}
+        }
+    });
+
+    fluid.defaults("fluid.prefs.defaultLineSpace", {
```
jobara: Probably better to name these more specifically. For example "fluid.prefs.lineSpace.textfieldSlider" and "fluid.prefs.lineSpace.textfieldButtons"

jobara:
@styki in response to your question about the version only being chosen on screen load. One issue you'll have is that there is a context check in the schema to specify which templates are fetched. The schema is parsed once and used to create a grade for instantiating the component. This means that even when you re-run the check you aren't pulling in the new templates because you aren't running the schema again. I guess you can look into the template loader and see if you can have it fetch more templates, but make sure that it is still easy and straightforward for an integrator to specify the specific templates ( and locations of them ) to be loaded.

src/components/textfieldButtons/js/TextfieldButtons.js
styki: I add the rounding because there was and issue with the floating point numbers. In my first implementation I just added 0.1 and didn't have to use a stepMultiplier. The problem was once you click a few times on the buttons rounding errors appear and the modelValue became  1.29999999. The method that I used first multiplies to a whole number and rounds it to get rid of the rounding error. After the value is changed it is then divided by the same stepMultiplier. If you have a better idea we can implement it, but that was the easiest way I could think of to not use floating point numbers.

jobara: I think we might have a utility for something like this, but I can't seem to find it. Perhaps @amb26 would know, or have some thoughts on this in general.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,166 @@
+/*
+Copyright 2013 OCAD University
```
cindyli: Fix the copyright to 2016. This applies to all new or updated js files in this branch. For new files, use 2016, for updated ones, add the year 2016, for example, changing from 2013 to 2013-2016. Thanks.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,166 @@
...
+                options: {
+                    model: "{textfieldButtons}.model",
+                    range: "{textfieldButtons}.options.range"
+                }
+            },
+            increaseButton: {
+                type: "fluid.stepper.button",
+                container: "{textfieldButtons}.dom.increaseButton",
+                options: {
+                    model: "{fluid.textfieldButtons}.model",
```
cindyli: The repetition of line 36-38 and line46-48 can be avoided using ```distributeOptions```. 

The doc: http://docs.fluidproject.org/infusion/development/IoCSS.html
An example: https://github.com/fluid-project/infusion/blob/master/src/components/reorderer/js/ModuleLayout.js#L204

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,166 @@
...
+                max: "{that}.options.range.max"
+            }
+        },
+        range: {
+            min: 0,
+            max: 100
+        }
+    });
+
+    fluid.defaults("fluid.textfieldButtons.textfield", {
```
cindyli: This component seems a copy-paste of ```fluid.textfieldSlider.textfield``` at https://github.com/fluid-project/infusion/blob/master/src/components/textfieldSlider/js/TextfieldSlider.js#L95

Please use the original component directly. Every time when you do a copy and paste, that is a warning signal. Thinking deeper in those cases to see if you can avoid the code repetition by creating a common utility function or creating a common grade component or refactoring the code to reuse common parts.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,166 @@
...
+        }
+    });
+
+    fluid.defaults("fluid.stepper.button", {
+        gradeNames: ["fluid.viewComponent"],
+        range: {}, // should be used to specify the min, max range e.g. {min: 0, max: 100}
+        invokers: {
+            setModel: {
+                changePath: "value",
+                value: {
```
cindyli: line 127-131 can be simplified to ```value: "{that}.calculateValue"```.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -0,0 +1,166 @@
...
+        buttonOptions: {
+            stepMultiplier: 10
+        }
+    });
+
+    fluid.stepper.button.buttonCalculateValue = function (value, stepMultiplier, modelValue, min, max) {
+        modelValue = Math.round(modelValue * stepMultiplier);
+        min = Math.round(min * stepMultiplier);
+        max = Math.round(max * stepMultiplier);
+        if ((modelValue + value) < min) {
```
cindyli: Isn't line 155-160 already guarded by ```fluid.transforms.limitRange``` checking at line 60-69?

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                listener: "{that}.applyMuteAudio",
+                args: ["{change}.value"]
+            }
+        }
+    });
+
+    // An array in which all the components that have been muted by this option are stored
+    var mutedComponents = [];
+
+    fluid.prefs.enactor.muteAudio.applyMuteAudio = function (value) {
```
cindyli: Change the parameter name ```value``` to a more intuitive name such as "isMuted".

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        $(document).ready(function () {
+            var audios = $("audio, video");
+            if (value) {
```
cindyli: The ```if``` check can be moved inside the loop of ```audios.each()```.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+            }
+            $(this).css("background-color", backgroundColorToSet);
+        });
+    };
+
+    fluid.prefs.enactor.blueColorFilter.addMutationObserver = function (applyFilter) {
+        // Uses document ready, because the script is added before the body
+        $(document).ready(function (){
+            var isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+            if (isSafari) {
```
cindyli: Are code in ```if``` and ```else``` blocks identical? Did I miss something? Please refactor due to subtle or even none difference.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                var backgroundColorToSet = "rgba(" + backgroundColor[0] + ", " + backgroundColor[1] + ", " + backgroundColor[2] + ", " + backgroundColor[3] + ")";
+            }
+            $(this).css("background-color", backgroundColorToSet);
+        });
+    };
+
+    fluid.prefs.enactor.blueColorFilter.addMutationObserver = function (applyFilter) {
+        // Uses document ready, because the script is added before the body
+        $(document).ready(function (){
+            var isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
```
cindyli: If the check on safari is still required, create a common utility function. See a similar checking on the web text-to-speech support: https://github.com/fluid-project/infusion/blob/master/src/components/textToSpeech/js/TextToSpeech.js#L26

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+            if (isSafari) {
+                var observer = new WebKitMutationObserver(function(mutations) {
+                    mutations.forEach(function (mutation) {
```
cindyli: Replace the use of ```forEach``` with ```fluid.each```. The same comment applies to the rest of this branch.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                        if (mutation.type === "childList") {
+                            // Changes in HTMLHtmlElement can be just hovering over elements, which should not trigger applyFilter
+                            if (changedDOMFlag === false && mutation.target != "[object HTMLHtmlElement]") {
```
cindyli: Use ```!==``` instead of ```!=```. Run ```grunt lint``` on this branch to get a linting report and fix them all. Thanks.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        if (times === undefined) {
+            times = blueColorFilterValue;
+        }
+        blueColorFilterValue = times;
+        var elements = $("body *");
+        var index;
+        elements.each(function () {
+            var element = this;
+            // Checks if the element exists in the dictionary with keys the elements and values the colors
+            for (var count in initialColorsDictionary) {
```
cindyli: Replace ```for``` with ```fluid.each```. The same comment applies to the whole branch.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        modelListeners: {
+            value: {
+                listener: "{that}.applyFilter"
+            }
+        }
+    });
+
+    /*Used in fluid.prefs.enactor.blueColorFilter.applyFilter and fluid.prefs.enactor.blueColorFilter.addMutationObserver
+     had to be defined here in order to not be create every time the model is changed */
+    var initialColorsDictionary = [];
```
cindyli: Use component options or component member options to track these 4 open variables. Component options are used for component users/integrators to config the component while component member options are more for internally tracking component changes.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+    var noMatchFlag = true;
+    var changedDOMFlag = false;
+    var blueColorFilterValue;
+
+    fluid.prefs.enactor.blueColorFilter.applyFilter = function (times) {
+        // Prevents a bug when the panel is open and the model.value becomes undefined
+        if (times === undefined) {
```
cindyli: Line 527-529 can be simplified to ```times = times || blueColorFilterValue```. ```blueColorFilterValue``` should be passed in as an argument.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+            times = blueColorFilterValue;
+        }
+        blueColorFilterValue = times;
```
cindyli: Confusing with line 528. Please refactor.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        var elements = $("body *");
+        var index;
+        elements.each(function () {
+            var element = this;
+            // Checks if the element exists in the dictionary with keys the elements and values the colors
+            for (var count in initialColorsDictionary) {
+                if (initialColorsDictionary[count].key == this) {
```
cindyli: Linting error with ```==```. Use ```===``` instead.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                    break;
+                }
+            }
+            // If it is not already in the records its color and background-color are added as arrays
+            if (noMatchFlag) {
+                var color = $(element).css("color");  // get the color as "rgb(..., ..., ...)"
+                color = color.slice(4, -1);
+                color = color.split(", ");  // color becomes and array with the 3 color components as his elements
+                var backgroundColor = $(element).css("background-color"); // the color can be is rgb or rgba format
+                if (backgroundColor.substr(0,4) === "rgba") {
```
cindyli: Line 548-553 can be simplified to:
```
var startPosition = backgroundColor.substr(0,4) === "rgba" ? 5 : 4;
backgroundColor = backgroundColor.slice(startPosition, -1);
```

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                    break;
+                }
+            }
+            var color = initialColors[index].color;  //get the color of the current component
+            color[2] = Math.round(color[2] * (1 - times)) + "";  //blue component
+            var colorToSet = "rgb(" + color[0] + ", " + color[1] + ", " + color[2] + ")";  //make the color in a "rbg(..,..,..)" format
+            $(this).css("color", colorToSet);
+            var backgroundColor = initialColors[index].backgroundColor;
+            backgroundColor[2] = Math.round(backgroundColor[2] * (1 - times)) + "";
+            if (backgroundColor[3] === undefined) {
```
cindyli: See comment above, refactor line 579-583 due to the similarity of ```if``` and ```else``` blocks.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                if (initialColors[count].key == this) {
+                    index = count;
+                    break;
+                }
+            }
+            var color = initialColors[index].color;  //get the color of the current component
+            color[2] = Math.round(color[2] * (1 - times)) + "";  //blue component
+            var colorToSet = "rgb(" + color[0] + ", " + color[1] + ", " + color[2] + ")";  //make the color in a "rbg(..,..,..)" format
+            $(this).css("color", colorToSet);
+            var backgroundColor = initialColors[index].backgroundColor;
```
cindyli: Add a comment to explain the array structure of ```backgroundColor```.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+    });
+
+    /*Used in fluid.prefs.enactor.blueColorFilter.applyFilter and fluid.prefs.enactor.blueColorFilter.addMutationObserver
+     had to be defined here in order to not be create every time the model is changed */
+    var initialColorsDictionary = [];
+    var noMatchFlag = true;
+    var changedDOMFlag = false;
+    var blueColorFilterValue;
+
+    fluid.prefs.enactor.blueColorFilter.applyFilter = function (times) {
```
cindyli: This function is rather long and unclear. It can be refactored into a few functions that achieve one step a time. For example, one function to determine no matching color, the other to handle the case of no matching etc. Also, add code comments for clarification. Thanks.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                    backgroundColor: backgroundColor
+                });
+            }
+            noMatchFlag = true;
+        });
+
+        var initialColors = $.extend(true, [], initialColorsDictionary);  // make a deep copy of the original array in order to prevent changes in the initial array
+        elements.each(function () {
+            // Finds the element in the dictionary
+            for (var count in initialColors) {
```
cindyli: fluid.each

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                if (initialColors[count].key == this) {
```
cindyli: linting problem with ```==```.

src/components/textfieldButtons/js/TextfieldButtons.js
styki: I am trying do that, but I don't how to have 2 different targets with the same record. The targets are {that > increaseButton}.options and {that > decreaseButton}.options.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                subtree : true
+            };
+
+            observer.observe(target, config);
+        });
+    };
+
+    /* Add a mutationObserver to track when the UI panel in its responsive version is opened.
+       It then adds and removes a class to the hide/show button to style it properly. */
+    $(document).ready(function (){
```
cindyli: Refactor this chunk of code into one or more infusion components. You may wanna consider to use context awareness (http://docs.fluidproject.org/infusion/development/ContextAwareness.html) to check if Safari is used and attach proper grade(s) in different scenarios.

src/components/textfieldButtons/js/TextfieldButtons.js
cindyli: The target should be ```{that > fluid.stepper.button}.options```.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        var isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+        function observerFunction() {
+            var openedResponsivePanelCheck = $(".flc-slidingPanel-panel.flc-prefsEditor-iframe").attr("aria-expanded");
```
cindyli: Use another context awareness for this check too.

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+            var panelButtons = $(".fl-prefsEditor-buttons");
+            if (openedResponsivePanelCheck == "true" && $(window).width() < 640) {
```
cindyli: linting with ```==```.

cindyli: The hardcoded ```640``` should be supplied from a component option.

cindyli: line 639-645 can be simplified to:
```
var isMobile = openedResponsivePanelCheck == "true" && $(window).width() < 640;
panelButtons[isMobile ? "addClass" : "removeClass"](...);
```

src/framework/preferences/js/Enactors.js
```diff
@@ -427,4 +427,244 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                panelButtons.addClass("flc-prefsEditor-opened-panel-buttons");
```
cindyli: 1. The class name should not be hardcoded here. It should be from a component option.
2. The infusion naming convention of when to use ```flc-``` and ```fl-``` prefixes:
- the class names starting with flc- are used to identify the elements to UI Options;
- the class names starting with fl- are used for visual styling.
The css class here is for visual styling so please use prefix ```fl-```. Also, the css styling classes are usually defined in components' options.styles block. An example: https://github.com/fluid-project/infusion/blob/master/src/components/overviewPanel/js/OverviewPanel.js#L138

src/framework/preferences/js/Panels.js
```diff
@@ -683,20 +706,33 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
                 }
             }
         },
-        protoTree: {
```
cindyli: Why the ```protoTree``` section is removed?

src/framework/preferences/js/Panels.js
```diff
@@ -662,13 +673,25 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        protoTree: {
+            label: {messagekey: "textSizeLabel"},
+            textSizeDescr: {messagekey: "textSizeDescr"},
+            smallIcon: {messagekey: "textSizeSmallIcon"},
+            largeIcon: {messagekey: "textSizeLargeIcon"},
+            multiplier: {messagekey: "multiplier"}
+        }
+    });
+
+    fluid.defaults("fluid.prefs.textSize.textfieldSlider", {
```
cindyli: 1. This component doesn't have a base grade name, add a comment to explain the reason, or even add a base grade ```fluid.component``` for the refactoring mention in (2). The same comment applies to ```fluid.prefs.textSize.textfieldButtons```.
2. Again, due to the similarity, create a base grade component to be shared by ```fluid.prefs.textSize.textfieldSlider``` and ```fluid.prefs.lineSpace.textfieldSlider``` and another grade to be shared by ```fluid.prefs.textSize.textfieldButtons``` and ```fluid.prefs.lineSpace.textfieldButtons```.

src/framework/preferences/js/Panels.js
```diff
@@ -750,7 +786,18 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
      * A sub-component of fluid.prefs that renders the "line space" panel of the user preferences interface.
      */
     fluid.defaults("fluid.prefs.panel.lineSpace", {
-        gradeNames: ["fluid.prefs.panel"],
+        gradeNames: ["fluid.prefs.panel", "fluid.contextAware"],
+        contextAwareness: {
```
cindyli: Move the ```contextAwareness``` section down below after ```preferenceMap``` to be in sync with other panel components.

cindyli: Same comment for ```fluid.prefs.panel.textSize```.

src/framework/preferences/js/Panels.js
```diff
@@ -788,12 +847,99 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+    /*********************************
+     * Preferences Editor Blue Color Filter *
+     *********************************/
+
+    /**
+     * A sub-component of fluid.prefs that renders the "blue color filter" panel of the user preferences interface.
+     */
+    fluid.defaults("fluid.prefs.panel.blueColorFilter", {
+        gradeNames: ["fluid.prefs.panel", "fluid.contextAware"],
+        contextAwareness: {
```
cindyli: move ```contextAwareness``` down below as above.

src/framework/preferences/js/Panels.js
```diff
@@ -683,20 +706,33 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
-            textSizeDescr: {messagekey: "textSizeDescr"}
-        },
         sliderOptions: {
             orientation: "horizontal",
             step: 0.1,
             range: "min"
         }
     });
 
+    fluid.defaults("fluid.prefs.textSize.textfieldButtons", {
```
cindyli: I'm not familiar with how or how well the text to speech tool works with the touch screen on mobile devices. It certainly concerns me when there is no any label or description being set in this button component. No matter how well the text to speech works on mobile devices, nothing would be announced when these buttons gain focus. Are there any label/description design for UIO mobile, @jhung? Or my concern doesn't make sense with mobile devices?

src/framework/preferences/js/Panels.js
```diff
@@ -788,12 +847,99 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        }
+    });
+
+    fluid.defaults("fluid.prefs.blueColorFilter.textfieldSlider", {
+        selectors: {
+            multiplier: ".flc-prefsEditor-multiplier",
+            blueColorFilterDescr: ".flc-prefsEditor-blue-color-filter-descr"
+        },
+        components: {
+            textfieldSlider: {
```
cindyli: Once the common grade component created for text size and line space components, it should also be reused here.

src/framework/preferences/js/Panels.js
```diff
@@ -802,6 +948,26 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         }
     });
 
+    fluid.defaults("fluid.prefs.blueColorFilter.textfieldButtons", {
```
cindyli: Use the common grade created on buttons for text size and line space panels.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -299,7 +299,29 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             var attrs = {height: height + 15}; // TODO: Configurable padding here
             var panel = separatedPanel.slidingPanel.locate("panel");
             panel.css({height: ""});
-            iframe.animate(attrs, 400);
+            if ($(window).width()< "640") {
```
cindyli: No hardcoded ```640```. Use component option instead.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -299,7 +299,29 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                iframe.removeAttr("style");
+                
+                //The rest of the if make the scroll go to the next tool
+                var jqueryDokkument = $(dokkument);
```
cindyli: Move this special handling into its own utility function.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -299,7 +299,29 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                var scrollStep = iframe.height();
+                var lastPosition = 0;
+                jqueryDokkument.on("scroll", function () {
+                    var scrollPosition = jqueryDokkument.scrollTop();
+                    //Scroll down
+                    if (lastPosition > scrollPosition) {
```
cindyli: Add code comments to clarify the purpose of these calculation.

src/components/textfieldButtons/js/TextfieldButtons.js
cindyli: Sorry, the use of ```expander``` is correct.

src/framework/preferences/js/StarterGrades.js
```diff
@@ -221,6 +243,21 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
                 }
             },
+            blueColorFilter: {
+                type: "fluid.prefs.panel.blueColorFilter",
+                container: "{prefsEditor}.dom.blueColorFilter",
+                createOnEvent: "onPrefsEditorMarkupReady",
+                options: {
+                    gradeNames: "fluid.prefs.prefsEditorConnections",
+                    model: {
+                        lineSpace: "{prefsEditor}.model.preferences.blueColorFilter"
```
cindyli: typo of "lineSpace"?

src/framework/preferences/js/StarterGrades.js
```diff
@@ -268,6 +305,21 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
                 }
             },
+            muteAudio: {
+                type: "fluid.prefs.panel.muteAudio",
+                container: "{prefsEditor}.dom.muteAudio",
+                createOnEvent: "onPrefsEditorMarkupReady",
+                options: {
+                    gradeNames: "fluid.prefs.prefsEditorConnections",
+                    model: {
+                        toc: "{prefsEditor}.model.preferences.muteAudio"
```
cindyli: typo of "toc"?

src/framework/preferences/js/StarterGrades.js
```diff
@@ -398,6 +452,7 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             textSize: "%messagePrefix/textSize.json",
             textFont: "%messagePrefix/textFont.json",
             lineSpace: "%messagePrefix/lineSpace.json",
+            blueColorFilter: "%messagePrefix/blueColorFilterjson",
```
cindyli: missing dot in from of "json". Please test to make sure messages are properly loaded and rendered. Thanks.

src/framework/preferences/js/StarterSchemas.js
```diff
@@ -21,11 +21,41 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
      * contrast, table of contents, inputs larger and emphasize links
      *******************************************************************************/
 
+    var check;
+    var widthCheck = function() {
```
cindyli: Avoid using private variables and functions. Use a global namespace starting as "fluid.prefs.xxxx" for this function. Worthwhile to use context awareness for this check too.

cindyli: Move the check related code into a separate .js file. Stick with starter schema handling in this file.

src/framework/preferences/js/StarterSchemas.js
```diff
@@ -21,11 +21,41 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+            fluid.contextAware.forgetChecks("fluid.responsiveCheck");
+            fluid.defaults("fluid.prefs.auxSchema.starter.default", {
+                gradeNames: ["fluid.contextAware"],
+                contextAwareness: {
+                    sliderVariety: {
+                        checks: {
+                            jQueryUI: {
+                                contextValue: "{fluid.prefsWidgetType}",
+                                equals: "jQueryUI",
+                                gradeNames: ["fluid.prefs.auxSchema.starter.textSize.jQueryUI", "fluid.prefs.auxSchema.starter.lineSpace.jQueryUI", "fluid.prefs.auxSchema.starter.blueColorFilter.jQueryUI"]
```
cindyli: Keep in mind UIO is flexible to mix in any panels. Having all three panel grades combined together would cause issue when some or all of these panels are not requested to compose UIO. These grades to be connected individually with their own starter components.

src/components/textfieldButtons/js/TextfieldButtons.js
```diff
@@ -20,9 +20,17 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
 
     fluid.defaults("fluid.textfieldButtons", {
         gradeNames: ["fluid.viewComponent"],
+        distributeOptions: {
+            target: "{that > fluid.stepper.button}.options",
+            record: {
+                model: "{fluid.textfieldButtons}.model",
+                range: "{textfieldButtons}.options.range",
```
cindyli: It would be nice to be consistent with other lines by using ```{fluid.textfieldButtons}```.

src/components/textfieldButtons/js/TextfieldButtons.js
styki: I am not entirely sure how the limit Range works. Does it return the max value if the supplied value is greater than the max and what is its syntax? I use stepMultipliers to avoid rounding errors. We discussed that with jobara here.

src/framework/preferences/js/Enactors.js
styki: This if statement had to fix an issue with the mutationObserver, because older versions of Safari required a webkit prefix, but it did not fix the issue. I will delete the Safari check and leave the issue open.

src/components/textfieldButtons/js/TextfieldButtons.js
cindyli: Some test cases to help you understand how limitRange works: https://github.com/fluid-project/infusion/blob/master/tests/framework-tests/core/js/ModelTransformationTests.js#L4028-L4077

The effectiveness should be tested in your test coverage for this component. 

From my understanding, this function simply needs one line:
```return modelValue + stepMultiplier * incrementCoefficient;```

BTW, using the name ```value``` in the parameter signature for ```incrementCoefficient``` is confusing.

src/framework/preferences/js/Enactors.js
styki: Grunt lint brought up that an argument is not used anywhere, but if I remove it the function does not work, so I will leave it there.

cindyli: Please don't leave linting errors behind. This needs further investigation. Thanks.

src/framework/preferences/js/Enactors.js
styki: I tried to replace all the for-s with fluid.each, but in one case it didn't work. Is there a break equivalent for the fluid.each function?

src/framework/preferences/js/Enactors.js
styki: I tried doing that, but there was a significant performance issue every time an element was added to the array with change applier. Does infusion use parse by reference for arrays, because if it does it will solve the problem?

cindyli: Models and change appliers are only used when the change on a model value triggers the change on other system parts, such as the front end display or the back end handling. ```noMatchFlag``` and ```changedDOMFlag``` seems falling in this category. But ```initialColorsDictionary``` is more for tracking the initial colors which is more proper to use component member options. An example: https://github.com/fluid-project/infusion/blob/master/src/components/pager/js/Pager.js#L74. Accessing and changing member option values are via ```that.initialColorsDictionary```.

src/framework/preferences/js/Enactors.js
cindyli: Right. Please use ```for``` or ```while``` when ```break``` is required. Thanks.

src/framework/preferences/js/Enactors.js
styki: I moved the two flags in the functions and the initialColorsDictionary and blueColorFilter are in the members option.

src/framework/preferences/js/Enactors.js
styki: Should these functions be a fluid functions or just vanilla function defined outside of the applyFilter function or they can even be inside it?

src/framework/preferences/js/Enactors.js
styki: I just have to be an expander to make the function run once. Do you have any suggestions where to put it?

src/framework/preferences/js/Panels.js
styki: I was moved to the general fluid.prefs.panel.textSize component.

src/framework/preferences/js/StarterGrades.js
styki: I fixed it, but everything worked fine before that, I don't know how and why.

src/framework/preferences/js/StarterSchemas.js
styki: There was a big problem with the contextAwareness and this was the only solution that worked.

src/framework/preferences/js/Panels.js
styki: I tried to do that, but  the options did not get pass on as they were supposed to. I will text you tomorrow to explain the problem in more details.

src/framework/preferences/js/Enactors.js
cindyli: These functions should be global namespaced fluid functions defined outside of the ```applyFilter```.

src/framework/preferences/js/Enactors.js
cindyli: Going thru the functionalities in this document ready function, please consider:
1. Whether the restyling of buttons can be done via pure css using responsive design;
2. Use context awareness for checking ```isSafari```;
3. The instantiation of the mutation service should only be presented together with the mute panel.

When all these are properly handled, this function may not be required any more.

src/framework/preferences/js/StarterSchemas.js
cindyli: Please refer the existing code in the master branch of how different versions of sliders are attached on the text size and line spacing panels - https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/StarterSchemas.js#L42, as well as line 101.

src/components/textfieldButtons/js/TextfieldButtons.js
styki: I am not sure where to put the transform. If I put it in the setModel invoker in the place of value the model ends up being a function. I am using this transformation
type: "fluid.transforms.limitRange",
input: "{that}.calculateValue",
min: "{that}.options.range.min",
max: "{that}.options.range.max"

cindyli: The transformation has already been placed in the parent component ```fluid.textfieldButtons``` - https://github.com/styki/infusion/blob/3fccf2929f4f10e2fc5f01e893ef6971c6fb1095/src/components/textfieldButtons/js/TextfieldButtons.js#L61-L69. Since the parent and children components are sharing the same model. It's enough to guard for all of them.

styki: I see I have just made some unnecessary checks.

src/framework/preferences/js/StarterSchemas.js
styki: I tried to do that, but the problem is that the we must have the responsive check first and if it fails then we have to check for jQuery or native slide. In each case we had a different template to be applied and they did not override each other. There was also a problem with nesting contextAwarenesses. I tried I lot of different versions and this is the only one that worked.

cindyli: Instead of nesting contextAwarenesses, have you tried to narrow down the context checks into 4 scenarios:
- jQuery + non-mobile
- jQuery + mobile
- nativeHtml + non-mobile
- nativeHtml + mobile

src/framework/preferences/js/Enactors.js
styki: I managed to solve the problem with CSS only.

cindyli: Awesome! Thanks, Stan.

src/framework/preferences/js/Panels.js
cindyli: Noticed you've moved the rendering of the label and description to the parent component so this comment is not valid.

src/framework/preferences/js/StarterSchemas.js
styki: I tried encapsulating the check in a function and the use a expander, but I got and error. For now I will leave it like that. If you have a better I idea you can try it, because I don't know which will be the best solution for this case.

src/framework/preferences/js/StarterSchemas.js
styki: I make a little bit more testing in the upcoming days, but there are to many things connected and will leave it like that for now.

the-t-in-rtf:
@cindyli @jhung, what is the status of this?  Is it likely to be merged?  I have an otherwise responsive design that needs to be integrated with UIOptions, and would like to use a responsive version if it's ready or heading toward being ready.

cindyli:
@the-t-in-rtf, thanks for bringing this up. 

This pull request contains three things:

1. The responsive design that replaces sliders with text field buttons on mobile devices;
2. A new feature called "Blue Color Filter";
3. Another new feature called "Mute Audio".

Before this pull request can be merged into the master, what's left to do:
1. The CSS part of the responsive design needs to be reviewed by @jhung;
2. The component for creating text field buttons and new features (item 2, 3 on the above list) need tests to be written.

@jhung, please add if I missed anything above.

I'm not sure how mature it is that new features of "Blue Color Filter" and "Mute Audio" have been decided to be included as UIO default panels. (@jhung, do you have some ideas?) If it's not yet decided, we probably should consider to split this pull request to 2 or even 3 individual pull requests so the responsive design itself can make into the master sooner and easier.

mgifford:
Would love to see this as part of a new release of this library.

waharnum:
@mgifford if you're referring to the responsive layout specifically, we've recently released a responsive version of the Preferences Framework that you can see in action in the demo at https://build.fluidproject.org/infusion/demos/prefsFramework/, among other places

waharnum:
@mgifford Yes, the latest master version of Infusion should include responsive layout for the Preferences Framework out of the box.

amb26:
@cindyli, @waharnum - is there work we need to rescue from this pull or is everything incorporated in the 2018 release with responsive design?

waharnum:
@amb26 I don't believe so, but @cindyli should confirm.

cindyli:
Most in this pull request was covered by the 2018 release with responsive design. The rest about the new "Blue Color Filter" feature has been taken off the implementation list I believe.

Closed.

PR #727, opened 2016-07-07
FLUID-5926: Lazy loading option for UIO and separatedPanel prefs editors

jobara:
Provides an option to lazy load the contents of the panel for a separatedPanel prefsEditor

https://issues.fluidproject.org/browse/FLUID-5926

[Page Load performance metrics](https://cdn.rawgit.com/jobara/prefsFramework-optimization-examples/466d6cd726fd7bfb7d703574841e0e55634e4599/index.html)

Associated Docs PR ( https://github.com/fluid-project/infusion-docs/pull/92 )


src/components/uiOptions/js/UIOptions.js
```diff
@@ -19,7 +19,8 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     });
 
     fluid.defaults("fluid.uiOptions.prefsEditor", {
-        gradeNames: ["fluid.prefs.constructed.prefsEditor"],
+        gradeNames: ["fluid.prefs.constructed.prefsEditor", "{that}.getLazyLoadGrade"],
+        lazyLoad: false,
```
amb26: This can/should be done via the contextAwareness facility. It's much more idiomatic to write a block which evaluates {fluid.uiOptions.prefsEditor}.options.lazyLoad and then outputs the fluid.uiOptions.prefsEditor.lazyLoad grade if it is found.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -333,4 +333,84 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                        }
+                    }
+                }
+            }
+        }
+    });
+
+    fluid.prefs.separatedPanel.lazyLoad.showPanel = function (separatedPanel, callback) {
+        if (separatedPanel.prefsEditor) {
+            setTimeout(callback, 1);
```
amb26: Use fluid.invokeLater(callback)

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -333,4 +333,84 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                }
+            },
+            slidingPanel: {
+                options: {
+                    invokers: {
+                        operateShow: {
+                            funcName: "fluid.prefs.separatedPanel.lazyLoad.showPanel",
+                            args: ["{separatedPanel}", "{that}.events.afterPanelShow.fire"],
+                            // override default implementation
+                            "this": null,
```
amb26: This usability bug should have been fixed by https://issues.fluidproject.org/browse/FLUID-5184 with https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/Fluid.js#L2479 - could you verify that this works fine with these "null" entries removed?

jobara: @amb26, it seems to work now without the "null" entries.

jobara:
@amb26, ready for another round of review.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -330,7 +339,84 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+            separatedPanel.events.onReady.addListener(function (that) {
+                that.events.onReady.removeListener("showPanelCallBack");
+                fluid.invokeLater(callback);
+            }, "showPanelCallBack");
+            separatedPanel.events.onLazyLoad.fire();
+        }
+
+    };
+
+    fluid.prefs.separatedPanel.lazyLoad.preloadResources = function (that, resources, toPreload) {
```
amb26: This function should be documented clearly - with the meaning and types of the arguments `resources` and `toPreload`. Also, some workflow seems to have been copypasted from the grade fluid.resourceLoader - this should at least be acknowledged and explained. Is this grade designed to be used alongside fluid.resourceLoader or does it replace it? What happens if there is a race over updating the "that.resources" member?

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -330,7 +339,84 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                            args: ["{separatedPanel}", "{that}.events.afterPanelShow.fire"]
+                        }
+                    }
+                }
+            }
+        }
+    });
+
+    fluid.prefs.separatedPanel.lazyLoad.showPanel = function (separatedPanel, callback) {
+        if (separatedPanel.prefsEditor) {
```
amb26: This if statement looks like an awkward race. Why is it here? Hopefully we can achieve this effect declaratively.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -330,7 +339,84 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+    fluid.prefs.separatedPanel.lazyLoad.preloadResources = function (that, resources, toPreload) {
+        toPreload = fluid.makeArray(toPreload);
+        var preloadResources = {};
+
+        fluid.each(toPreload, function (resourceName) {
+            preloadResources[resourceName] = resources[resourceName];
+        });
+
+        fluid.fetchResources(preloadResources, function () {
+            that.resources = preloadResources;
```
amb26: It's not obvious that there is a test verifying that this member gets set. I assume that this grade can never coincide with the "messageLoader" which is checked in the tests.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -330,7 +339,84 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     fluid.prefs.separatedPanel.showPanel = function (panel, callback) {
         // A bizarre race condition has emerged under FF where the iframe held within the panel does not
         // react synchronously to being shown
-        setTimeout(callback, 1);
+        fluid.invokeLater(callback);
+    };
+
+    fluid.defaults("fluid.prefs.separatedPanel.lazyLoad", {
```
amb26: I think this grade could do with a big docs comment explaining its strategy - and outlining the expected sequence of activities - what happens first, which events are fired, and what they then trigger off. It should clearly contrast the standard fetch workflow of separatedPanel with the modified one and reference the JIRA and use cases for this functionality.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
jobara: The purpose is to handle the lazy loading. If the `separatedPanel.prefsEditor` has been instantiated already, we merely just trigger the typical opening flow to show the panel. If it hasn't be instantiated yet, we go through the process of instantiating it. Not sure of how to do this completely declaratively, but open to suggestions.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
jobara: This is part of a function that is used to reconfigure how the messageLoader fetches resources. This is because we need to load in some of the messages, i.e the text on the tab of the sliding panel, but don't want to load in all of the strings for the panels until the prefsEditor is to be instantiated. That is why the tests verifies `prefsEditorLoader.messageLoader.resources`.

jobara:
@amb26 I've added comments where you requested. I've also left a couple to replies to your other comments. Please let me know if there is anything else needed.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -330,7 +339,121 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     fluid.prefs.separatedPanel.showPanel = function (panel, callback) {
         // A bizarre race condition has emerged under FF where the iframe held within the panel does not
         // react synchronously to being shown
-        setTimeout(callback, 1);
+        fluid.invokeLater(callback);
+    };
+
+    /**
+     * FLUID-5926: Some of our users have asked for ways to improve the initial page load
```
amb26: This is a great comment

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
amb26: Ok - I guess at least we can remove the deregistration of the listener of itself in order to avoid the obvious "listener removing itself" code smell - since the onReady event for the separatedPanel could not possibly fire more than once. Other than that I guess it's hard to think of a useful improvement. Time for a few more showers.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -330,7 +339,121 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+    fluid.prefs.separatedPanel.lazyLoad.preloadResources = function (that, resources, toPreload) {
+        toPreload = fluid.makeArray(toPreload);
+        var preloadResources = {};
+
+        fluid.each(toPreload, function (resourceName) {
+            preloadResources[resourceName] = resources[resourceName];
+        });
+
+        fluid.fetchResources(preloadResources, function () {
+            that.resources = preloadResources;
```
amb26: Ok. Could you add a comment indicating that this logic is copied from the implementation of fluid.resourceLoader.loadResources

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -330,7 +339,121 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+                fluid.invokeLater(callback);
+            }, "showPanelCallBack");
+            separatedPanel.events.onLazyLoad.fire();
+        }
+
+    };
+
+    /**
+     * Used to override the standard "onCreate.loadResources" listener for fluid.resourceLoader component,
+     * allowing for pre-loading of a subset of resources. This is requried for the lazyLoading workflow
```
amb26: Typo requried

jobara:
@amb26 fixed up the comments as suggested.

amb26:
Thanks for the great work on this feature, @jobara - as well as being a practical necessity, it has opened up new horizons and issues for the framework.

PR #728, opened 2016-07-11
FLUID-5927: Asserting that context are the same DOM element

jobara:
Doing a deep equals on the documents causes Safari to fail as there are some lower level properties
that aren't set yet and have TyperError as a value. E.g. selectionStart. However it should be
sufficient to do an assertEquals to ensure that the contexts refer to the same Document.

https://issues.fluidproject.org/browse/FLUID-5927

### DIFF
--- a/tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorTests.js
+++ b/tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorTests.js
@@ -347,7 +347,7 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
             fluid.each(memberNames, function (memberName) {
                 var comp = parentPanel[memberName];
                 jqUnit.assertNotUndefined("The " + memberName + " should be instantiated", comp);
-                jqUnit.assertDeepEq("The context for " + memberName + " should be retained", comp.origContext, comp.container[0].ownerDocument);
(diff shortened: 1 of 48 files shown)


tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorTests.js
```diff
@@ -347,7 +347,7 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
             fluid.each(memberNames, function (memberName) {
                 var comp = parentPanel[memberName];
                 jqUnit.assertNotUndefined("The " + memberName + " should be instantiated", comp);
-                jqUnit.assertDeepEq("The context for " + memberName + " should be retained", comp.origContext, comp.container[0].ownerDocument);
```
amb26: This change is fine, and the tests should never have been using assertDeepEq for DOM nodes. However, I can't test this pull myself.

colinbdclark: I ran this test in Justin's branch on my Mac and it now passes with no failure messages in Safari.

cindyli:
The fix on the same issue has been addressed and merged in [this pull request](https://github.com/fluid-project/infusion/pull/734). @jobara, can you please verify? Thanks.

jobara:
@cindyli yes thanks for catching that. I knew I felt like I had fixed that before.

jobara:
Fixed by https://github.com/fluid-project/infusion/pull/734

PR #722, opened 2016-06-21
FLUID-5922 - Upgrade to Fedora 24

gtirloni:
To test, retrieve this PR and run:

```
vagrant up
grunt tests
```

PR #729, opened 2016-07-20
FLUID-5932: Provide a fallback when grunt run outside of git repo

jobara:
Returns "unknown" when grunt attempts to get the git branch and revision from the downloaded zip version of Infusion.

https://issues.fluidproject.org/browse/FLUID-5932


gtirloni:
This is a blocker for #722 as it fails to provision the VM.

Gruntfile.js
```diff
@@ -16,13 +16,23 @@ var _ = require("lodash");
 var path = require("path");
 var execSync = require("child_process").execSync;
 
+var getFromExec = function (command) {
+    var result;
+    try {
+        result = execSync(command);
+    } catch (e) {
+        result = "unknown\n";
```
cindyli: I wonder what is captured in the argument ```e``` when an error occurs. Does it make sense to return an error message or an error code or both if there's any returned by ```e```.

jobara:
@cindyli I'm logging the error info. We expect errors to happen when a build is created from a copy of the repo that isn't a git repo.. example from a zip download from GitHub

Gruntfile.js
```diff
@@ -16,13 +16,25 @@ var _ = require("lodash");
 var path = require("path");
 var execSync = require("child_process").execSync;
 
+var getFromExec = function (command) {
+    var result;
+    try {
+        result = execSync(command);
+    } catch (e) {
+        console.log("Error executing command: " + command);
```
amb26: The error and stack trace output that appears from this task is unsightly and will probably confuse users who are in this workflow. I suggest that we add an "options" block to this function with a configuration silent: true/false which will swallow this output conditionally. We can then perhaps also support a default return value in this options block.

Gruntfile.js
```diff
@@ -16,13 +16,25 @@ var _ = require("lodash");
...
+};
+
 module.exports = function (grunt) {
 
     // Project configuration.
     grunt.initConfig({
         pkg: grunt.file.readJSON("package.json"),
-        revision: execSync("git rev-parse --verify --short HEAD"),
-        branch: execSync("git rev-parse --abbrev-ref HEAD"),
+        revision: getFromExec("git rev-parse --verify --short HEAD"),
```
amb26: With the above fixes, this could become getFromExec( ...., {silent: true, defaultValue: "Unknown revision, not within a git repository"})

Gruntfile.js
amb26: silent: true can also suppress the stderr output by passing "ignore" as the "stdio" option to execSync: 
https://nodejs.org/api/child_process.html#child_process_options_stdio

jobara:
@amb26 ready for more review

Gruntfile.js
```diff
@@ -16,13 +16,35 @@ var _ = require("lodash");
...
+ * @returns {String} - returns a string representation of the result of the command or the defaultValue.
+ */
+var getFromExec = function (command, options) {
+    var result = options.defaultValue;
+    var stdio = options.verbose ? "pipe" : "ignore";
+
+    try {
+        result = execSync(command, {stdio: stdio });
+    } catch (e) {
+        console.log("Error executing command: " + command);
```
amb26: We shouldn't output this message to the console unless verbose is set to true

jobara:
@amb26 ready for more review

avtar:
@amb26, @colinbdclark, @jobara I need to test the Infusion CI job and having these changes merged soon would make it easier to proceed.

PR #566, opened 2014-10-30
FLUID-5542: Queued DataSource

jobara:
Added a Queued DataSource that can combined with different requestQueues to synchronize, throttle, and debounce requests to a DataSource.

http://issues.fluidproject.org/browse/FLUID-5542

This is an extension of the work started in https://github.com/fluid-project/metadata/pull/43

PR #732, opened 2016-08-04
FLUID-5936: fix bug in fluid.textToSpeech.checkTTSSupport, rewrite tests to use IoC testing

waharnum:
This PR addresses the bug described at https://issues.fluidproject.org/browse/FLUID-5936

Additionally, it rewrites the existing tests (but does not change what they test) to use the IoC testing framework to make test hangs related to async timeouts or non-firing events more visible.


amb26:
These tests hang for me at "Test Pause and Resume Events":

hu Aug 04 2016 20:49:42 GMT+0100 (GMT Daylight Time):   Test case listener has not responded after 5000ms - at sequence pos 5 of 6 sequence element  Object {listener: "fluid.tests.textToSpeech.testResume", args: Array[1], event: "{tts}.events.onResume"}  of fixture Test Pause and Resume Events

Configuration: Chrome 52.0.2743.116 m (64-bit) on Windows 7 64-bit

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -34,110 +34,150 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-            var that = fluid.tests.textToSpeech.pauseResume();
-            that.queueSpeech("Testing pause and resume events");
+    fluid.tests.textToSpeech.bypassTest = function () {
+        jqUnit.test("Tests were skipped - browser does not appear to support TTS", function () {
+            jqUnit.assert("TESTS SKIPPED - browser does not support SpeechSynthesis");
         });
-    }
+    };
+
+    // fluid.setLogging(fluid.logLevel.TRACE);
```
amb26: Stray comment

waharnum:
Thanks @amb26 - booting up my Windows VM to check. @jobara has mentioned there was an issue with pause/resume in Chrome that caused that test to be skipped via browser detection previously, it's possible that's now Windows-only.

waharnum:
The best way of achieving reliable tests with the current implementation (in both senses I used it above) appears to be short delays in issuing pause/resume commands, which also more closely emulates a typical user interaction with things like pause/resume controls.

From what research I've done (and what information is easily available), I think that various browser implementations of TTS have somewhat differing asynchronous behaviours, both within the different browsers, and within the browser's interaction with the operating system APIs to perform TTS.

This leads to differences in when certain TTS events the wrapper component depends on to fire its own events occur, both generally between different browsers/OSs, and in individual instances.

A long way of saying that introducing a short delay when invoking pause/resume leads to reliably passing tests on OS X and Windows.

amb26:
This doesn't strike me as very reliable. If the component is not responsive to events on a certain timescale, this should be built into its implementation rather than left as an artefact of the test cases. We should work to absorb whatever cross-platform slop there is into our own interface rather than special-case our testing requirements.

waharnum:
I'm of two minds about this (and trying to not have one or both of them influenced by the mind that would like to move on from the pit of cross-browser/cross-OS TTS issues...):

**Argument for Leaving the Component As-Is**

I think it's less that the component itself is not responsive to events on a certain timescale than that underlying  browser TTS events the component relies upon do not appear to be responsive to events on a certain timescale. I know this may sound like developer finger-pointing, but I think it has at least some basis in reality. I believe you'd see similar issues non-asychronously issuing rapid pause/resume commands through the TTS interface - events get lost, race conditions occur, etc.

**Argument for Absorbing Cross-Platform Slop**

I think it would probably be possible to make some improvements as part of this PR that might help to deal with the cross-platform issues. In particular, I haven't traced all the scenarios fully, but I believe there may be instances where internal component state can diverge from the state of the browser TTS interface around whether or not speech is playing/paused.

So I can try refactoring to:

- make the component state more reflective of / coupled to the browser TTS state. This may introduce other issues, due to bugs such as https://bugs.chromium.org/p/chromium/issues/detail?id=425553&q=SpeechSynthesis&colspec=ID%20Pri%20M%20Week%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified

- introduce the slight delays I found improved test reliability into the component's invocation of the browser TTS. This might improve race conditions and non-firing events.

Writing this out makes me realize I should be of the second mind, as I'm probably in the best position to do that work I'll be right now, as opposed to if I step away from it for weeks until I "have more time". 😛

waharnum:
I've moved the approach of issuing asynchronous pause/resume commands to the `speechSynthesis` object from the tests to the main component, and refactored the invokers to use this approach (the `fluid.textToSpeech.asyncSpeechSynthesisControl` function) rather than invoking the `speechSynthesis` commands directly.

A functionally imperceptible delay via `setTimeout` (10ms) before issuing commands results in consistently passing tests in TTS-supporting browsers on both Mac and Windows.

waharnum:
@amb26, is there anything else that you think should be done on this one? It would be nice to get it in for the 2.0 release.

waharnum:
update to say that I'm also looking at https://issues.fluidproject.org/projects/FLUID/issues/FLUID-5735 as part of this - tests are still failing when run as part of all-tests, but I believe I've pinned down why and should have a fix shortly.

waharnum:
I believe FLUID-5735 should now be fixed as part of this PR by commit 5ca02a86e3d6b8520f0e4623726cea0d68c57743 - the issue was with the approach taken to deciding whether or not to execute the test. Wrapping it in the `fluid.textToSpeech.checkTTSSupport` promise caused the all-tests runner to believe it was OK to move on from this test to the next one, which caused this test to fail (0 assertions run) and then causing the next test in line to fail with an incorrect number of assertions, as the tests that were being run as a result of the promise resolving were then run as part of the next test context.

I resolved this by wrapping the promise-based structure in an `asyncTest` structure; this results in the intended outcome of the tests being run or skipped based on the results of the promise ("does the test environment truly support TTS?"), while also playing nicely with the all-tests environment.

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -142,6 +142,16 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         }
     });
 
+    // Issue functions to the speechSynthesis interface
```
amb26: Very short line wrapped comments - perhaps an IDE setting run amok?

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -31,113 +31,171 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
     });
 
     fluid.tests.textToSpeech.cleanUp = function () {
-        speechSynthesis.cancel();
+        if(fluid.textToSpeech.isSupported()) {
```
amb26: Presumably spacing that fails linting - or at least I hope it does!

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -31,113 +31,171 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            fluid.tests.textToSpeech.bypassTest();
         });
-    }
 
+        var ttsTest = function () {
+            jqUnit.start();
+            jqUnit.assert(ttsStatus);
+        };
+
+        setTimeout(ttsTest, 1500);
```
amb26: Sorry, looks like I managed to comment on the commit for this rather than the pull

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
waharnum: What really fails is my failure to install my precommit hooks to make sure i can't commit unlinted code :(

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -31,113 +31,171 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        ttsSupport.then(function (){
+            ttsStatus = "TTS is available";
+            fluid.tests.textToSpeech.ttsTestEnvironment();
+        }, function () {
+            ttsStatus = "TTS is unavailable";          
+            fluid.tests.textToSpeech.bypassTest();
         });
-    }
 
+        var ttsTest = function () {
```
amb26: I don't see a reason for enclosing these definitions in a function and executing them asynchronously?

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
waharnum: It's .5 seconds longer than the default timeout of `fluid.textToSpeech.checkTTSSupport`, so we can be assured that promise has resolved and the real TTS tests are either running or have been skipped. It may be a little convoluted, I'll see if I can be simplified.

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
waharnum: This is to ensure the actual asynchronous TTS tests have either started to run or been skipped because TTS is not supported in the all-tests context. 

If there's not an asynchronous test wrapper around the promise at lines 184-191, the all-tests runner moves on to the next test (it won't wait for the promise to resolve); this seems to have been the cause of https://issues.fluidproject.org/projects/FLUID/issues/FLUID-5735.

I tried experimenting at first with a `fluid.tests.textToSpeech.ttsTestEnvironment` grade that delayed creation of the `ttsTester` component until the TTS support had been tested for, but couldn't make this work in a way that prevented the all-tests runner from moving on.

amb26: Ok, I think we are mixing up here the requirement to have executed the asynchronous fluid.textToSpeech.checkTTSSupport before beginning the test and the asynchronous tests themselves. The "test scheduling a test" workflow is ok (since this happens a lot in practice with QUnit) but can be cleaned up.

All that we need is for the body of ttsTest to execute after the ttsSupport promise resolution, and the reject case is already fully self-sufficient, all we are adding to it is an extra assertion message. The most compact way of setting this up is probably to insert a ttsSupport.then(jqUnit.start, jqUnit.start) at line 185, and eliminate all the other material.

waharnum: Understood - I simplified it and also extracted it as a general function pattern for choosing test execution / bypass based on a promise result in a way that behaves safely with the all-tests runner.

waharnum:
Do we feel okay with merging this one now @jobara / @amb26?

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -31,113 +31,175 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+    };
 
-    // Chrome doesn't properly support pause which causes this test to break.
-    // see: https://code.google.com/p/chromium/issues/detail?id=425553&q=SpeechSynthesis&colspec=ID%20Pri%20M%20Week%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified
-    if (!window.chrome) {
-        fluid.tests.textToSpeech.issueTest("Pause and Resume Events", function () {
-            jqUnit.expect(8);
-            var that = fluid.tests.textToSpeech.pauseResume();
-            that.queueSpeech("Testing pause and resume events");
+    // Chooses which test function to execute based on the results of a
```
amb26: Just be a little clearer with this comment. The primary purpose isn't entirely to choose which test function to execute - but to cause the QUnit test runner to suspend whilst the decision is being made asynchronously. The "all-tests runner context" should really be described as QUnit's asynchronous test execution model. Without this wrapper, we'd get failures even just queuing several tests up in the same file without all-tests being involved.

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -31,113 +31,175 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+    // promise; wraps the promise in an asyncTest for good behaviour
+    // within the all-tests runner context
+    //
+    // wrapperMessage, promise, resolveTestFunc, rejectTestFunc: required
+    // resolveMessage, rejectMessage: optional string, passed to test function
+    fluid.tests.textToSpeech.chooseTestByPromiseResult = function (wrapperMessage, promise, resolveTestFunc, rejectTestFunc, resolveMessage, rejectMessage) {
+        resolveMessage = resolveMessage || null;
```
amb26: I don't think forcing these from undefined to null adds much - perhaps you can add a semi-meaningful default instead.

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -31,113 +31,175 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-            jqUnit.expect(8);
-            var that = fluid.tests.textToSpeech.pauseResume();
-            that.queueSpeech("Testing pause and resume events");
+    // Chooses which test function to execute based on the results of a
+    // promise; wraps the promise in an asyncTest for good behaviour
+    // within the all-tests runner context
+    //
+    // wrapperMessage, promise, resolveTestFunc, rejectTestFunc: required
+    // resolveMessage, rejectMessage: optional string, passed to test function
+    fluid.tests.textToSpeech.chooseTestByPromiseResult = function (wrapperMessage, promise, resolveTestFunc, rejectTestFunc, resolveMessage, rejectMessage) {
```
amb26: Proper name for "function returning a promise" is "task": See prior art at https://github.com/cujojs/when/blob/master/docs/api.md#task-execution

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
waharnum: I've added some hopefully non-pleonasmic text to make the purpose clearer.

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
waharnum: I may have misunderstood this comment, but have renamed the function accordingly.

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -142,6 +142,14 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         }
     });
 
+    // Issue functions to the speechSynthesis interface after a delay; this
+    // makes the wrapper behave better when issuing commands in rapid succession
+    fluid.textToSpeech.asyncSpeechSynthesisControl = function (control, delay) {
```
amb26: I *really* wanted to approve this and had written a comment declaring that it was probably "good enough" but I think on balance that it probably isn't. There is no guarantee that timeouts scheduled at closely spaced instants in time will be honoured in the order that they are issued - here is some art: http://stackoverflow.com/questions/1776239/are-equal-timeouts-executed-in-order-in-javascript
I don't think there's an alternative to explicitly managing a queue of activities in the component. It will also make the body of this method a little more slightly.

waharnum: I agree with your assessment - closely spaced timeouts are notoriously unreliable. I was hoping this might be a sufficient patch for the moment, but a managed queue of pause/resume events won't take too long to implement.

waharnum:
I've implemented a basic queue scheme for pause/resume events, along with making all the control issuance asynchronous. The two together have led to a greater increased reliability. I also updated the tests to quite aggressively alternate pause/resume commands, which the queueing scheme seems to handle well. I need to do some further refactoring and add some comments.

I am still seeing occasional failures in Safari, which I believe may be related to the issue described at https://bugs.chromium.org/p/chromium/issues/detail?id=509488#c11 due to the common WebKit heritage. I tested the proposed "store the SpeechSynthesisUtterance on a global variable" solution in the comment and found that it caused Safari to become as reliable as Chrome and Firefox for speech synthesis. I'm not certain if I should implement this, or if there's a more elegant way to do it then that approach.

amb26:
Horrifying issue report there. The commenter does mention "or other scope" which suggests that we could put it essentially anywhere. This then suggests that we should be able to adopt the relatively principled solution of storing it in the queue structure that you just created.

waharnum:
I added a `currentUtterance` member to the component to store the utterance outside of the function scope - this appears to work just as well and suggests that the thesis about premature garbage collection before the speech synthesis has actually completed (causing the `onend` event to never fire) is correct.

I ran 25 test runs manually each in Safari with and without the storage on the member with the following results:

```
WITH:
Success: 25
Fail: 0

WITHOUT:
Success: 19
Fail (SpeechSynthesisUtterance.onend callback never fires) : 6
```

While anecdatal, my sense was Safari failed prior to that change about 25% of the time. That change results in it being as reliable as the other browsers.

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -132,25 +130,55 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
     };
 
-    fluid.textToSpeech.pause = function (that, paused) {
-        if (paused) {
-            that.events.onPause.fire();
-        } else if (that.model.speaking) {
-            that.events.onResume.fire();
+    fluid.textToSpeech.handlePause = function (that) {
+        that.applier.change("paused", true);
+        that.events.onPause.fire();
```
amb26: No need to duplicate the function of the model events with a regular event

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -132,25 +130,55 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+
+    fluid.textToSpeech.handleResume = function (that) {
+        that.applier.change("paused", false);
+        that.events.onResume.fire();
+        // Clear to issue any waiting pause command
+        fluid.textToSpeech.clearControlRequest(that, "pauseRequested", "pause");
+    };
+
+    fluid.textToSpeech.clearControlRequest = function (that, modelBoolPath, controlName) {
+        if(fluid.get(that.model, modelBoolPath)) {
```
amb26: Remember to lint!

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -132,25 +130,55 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+            that.applier.change(modelBoolPath, false);
+            fluid.textToSpeech.asyncSpeechSynthesisControl(controlName, 10);
+        }
+    };
+
+    fluid.textToSpeech.issueControlRequest = function (that, modelBoolPath, controlName, invert) {
```
amb26: This can be eliminated and the invokers replaced with a "changePath" record - http://docs.fluidproject.org/infusion/development/ChangeApplierAPI.html#declarative-style-for-triggering-a-change

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -132,25 +130,55 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
-        } else if (that.model.speaking) {
-            that.events.onResume.fire();
+    fluid.textToSpeech.handlePause = function (that) {
+        that.applier.change("paused", true);
+        that.events.onPause.fire();
+        // Clear to issue any waiting resume command
+        fluid.textToSpeech.clearControlRequest(that, "resumeRequested", "resume");
+    };
+
+    fluid.textToSpeech.handleResume = function (that) {
```
amb26: These two handleXxxx can be eliminated

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -132,25 +130,55 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        that.applier.change("paused", false);
+        that.events.onResume.fire();
+        // Clear to issue any waiting pause command
+        fluid.textToSpeech.clearControlRequest(that, "pauseRequested", "pause");
+    };
+
+    fluid.textToSpeech.clearControlRequest = function (that, modelBoolPath, controlName) {
```
amb26: This can be replaced with a straightforward modelListener which performs a "debounce" on the model state. Consult @jobara who has a mothballed pull request containing this algorithm. https://github.com/fluid-project/infusion/pull/566 - give we have also ripped it off here - https://github.com/fluid-project/node-jqunit/blob/master/lib/jqUnit-node.js#L158 - our "rule of thumb" suggests that this now needs to go into the core framework

amb26: Useful guide: https://css-tricks.com/the-difference-between-throttling-and-debouncing/#article-header-id-1

amb26: Ok - the guide is helpful to remind me too - the algorithm we want is actually "throttle" (with trailing) rather than "debounce" :)

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -186,6 +214,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         };
 
         var toSpeak = new SpeechSynthesisUtterance(text);
+        // Store toSpeak additional on the currentUtterance member to help deal with the issue
+        // with premature garbage collection described at https://bugs.chromium.org/p/chromium/issues/detail?id=509488#c11
+        // this makes the speech synthesis behave much better in Safari in
+        // particular
+        // that.currentUtterance = toSpeak;
```
amb26: Currently commented out

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -72,9 +72,11 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             onSpeechQueued: null
         },
         members: {
-            queue: []
+            queue: [],
+            //
+            currentUtterance: {}
```
amb26: I think my suggestion of sticking this in the queue is better, since otherwise you have to make extra effort to clean it up when the utterance is done.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: One (possible) argument to make for `onPause` and `onResume` events when they could be handled by model events is that it keeps a parallel non-model event structure in the component to the events in the wrapped WebSpeechSynthesis API.

Second possible point: it's a somewhat fine-grained (and possibly redundant) distinction but the model change represents a change to the component-tracked state of the speech, while the non-model events represent callbacks from the wrapped browser API about the state of the utterance.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: The `queue` member contains the queue of texts yet to be spoken; it has nothing to do with the queueing behaviour I implemented, which (as you state above) is probably more correctly described as a throttle to prevent more pause/resume commands from being sent via the wrapper while outstanding pause/resume requests are still resolving.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: I've got my pre-commit linter running now on my Infusion project, so it runs (and aborts the commit on any linting errors) before every commit. Whatever the issue the `.eslintrc.json` config in Infusion isn't catching it. (made sure to update to latest upstream master before saying this...)

amb26: Ok, looks like we need to add in this rule to our mix: http://eslint.org/docs/rules/keyword-spacing

src/components/textToSpeech/js/TextToSpeech.js
amb26: Yes, that's right - the queue has nothing to do with your throttling, but it's still an appropriate place for the utterance. Note that it is created one for one with each queue element: https://github.com/fluid-project/infusion/blob/master/src/components/textToSpeech/js/TextToSpeech.js#L188

src/components/textToSpeech/js/TextToSpeech.js
waharnum: Does it make sense per comment above about Rule of Thumb to simply move `fluid.debounce` from the jqUnit implementation and borrow the Underscore implementation of throttle for `fluid.throttle`? Since I have need of them here, I can do this as part of the PR.

src/components/textToSpeech/js/TextToSpeech.js
amb26: I can't go along with either of these arguments :) The whole purpose of having a model idiom is to avoid the proliferation of "uninterpreted instants in time". You have 2 events and 4 invokers, when you could have a simple flag in a model and a call to a standard algorithm. Ideally we never have "non-model event structure", parallel or otherwise.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: I've made a small NOJIRA PR for this (and fixed two files that had the same style issue as part of it): https://github.com/fluid-project/infusion/pull/741

src/components/textToSpeech/js/TextToSpeech.js
waharnum: The perils of running 50 differential manual tests at the end of the day is that you do things like this and comment out the actual fix for the problem.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: Point taken - but in this case, we do need to establish some level of communication and sequencing between the wrapper component and the WebSpeechSynthesis API (so that the model doesn't enter a state that doesn't actually reflect what's happening with the browser API, which is where some of the problems have centred working on the issue of irregularly failing tests). But I'll look for ways to simplify.

waharnum: A further note on the non-model events: the public API as documented at http://docs.fluidproject.org/infusion/development/TextToSpeechAPI.html includes the `onPause` and `onResume` events.

I'm uncertain how much removing them might impact any downstream projects, if at all.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: I've gone ahead and adapted Underscore's implementation of throttle (with credit and license) in the component itself.

waharnum:
@amb26 - I've implemented throttle-based control, as you recommended, and reduced the dependence on functions for managing control vs. model paths & model listeners. At this point we should decide if the changes could go in their current form for Infusion 2.0 or if further work is needed - I don't think I have more cycles to spend on this.

waharnum:
I also want to flag the comment at https://github.com/fluid-project/infusion/pull/732#discussion_r76832166 - if we remove the non-model events for pause/resume, we'll need to update the documentation as well to say those events have been removed and that the alternative is to use model listeners.

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -170,7 +242,11 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             paused: false
         };
 
-        if (!that.queue.length) {
+        if (that.queue.currentUtterance) {
+            that.queue.currentUtterance = undefined;
```
amb26: I actually meant to put it in the queue item, rather than to rebase the entire queue structure.
So, on line 280 we would have that.queue.push({text: text, utterance: toSpeak})
queue can then just go back to being a [] and the rest of the changes reverted

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -142,23 +156,81 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
 
-    fluid.textToSpeech.pause = function (that, paused) {
-        if (paused) {
-            that.events.onPause.fire();
-        } else if (that.model.speaking) {
-            that.events.onResume.fire();
+    fluid.textToSpeech.requestControl = function (that, control, change) {
+        // If there's a control request (value change to true), clear and
+        // execute it
+        if(change.value) {
```
amb26: I assume still fails linting :)

amb26:
Thanks, @waharnum - changes look great, with a minor tweak. Thanks for flagging the required docs update - this change is ok at present since we haven't yet made a release with this component included.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: What fails (the sequel) is actually my ability to install my precommit hooks on my laptop as well as my work desktop to make sure I can't commit unlinted code.

waharnum:
Thanks  @amb26 - I'll make that change to the queue structure (and fix the lint), and do a separate docs pull request to remove the onPause/onResume events.

src/components/textToSpeech/js/TextToSpeech.js
waharnum: Agree this is better; I had to refactor the queue behaviour a little so `that.queue.shift` happens at `handleEnd` rather than `handleStart` or we run into the same premature garbage collection behaviour in Safari that we were seeing previously. I don't think this is a problematic change (it also lets `handleStart` simply be a changePath record rather than a function), but please let me know if you foresee an issue with that behavioural change.

waharnum:
@amb26 / @jobara:

- queue updated as suggested
- parallel documentation pull at https://github.com/fluid-project/infusion-docs/pull/99/files (JIRA at https://issues.fluidproject.org/projects/FLUID/issues/FLUID-5936)

amb26:
Hey there @waharnum - thanks so much for the marathon work on this, and pushing it the extra mile to the end!

PR #733, opened 2016-08-12
FLUID-5192: Removed strings for slider icons

jobara:
It was determined in FLUID-5680 that the icons for the sliders are presentational.
The labels were not being used either so I have removed all of the strings and associated
code to insert them into the DOM.

https://issues.fluidproject.org/browse/FLUID-5192


cindyli:
Merged at 8c675a5a4a90a1e8d50074ae08541e067bcb64fe

PR #735, opened 2016-08-12
FLUID-5940: Checking if we were successful in resolving Infusion

michelled:


src/module/fluid.js
```diff
@@ -38,9 +38,11 @@ var upPath = path.resolve(__dirname, "../../../../..");
 var upInfusionPath = resolveModuleSync("infusion", upPath);
 if (upInfusionPath) {
     upInfusion = require(upInfusionPath);
-    upInfusion.log("Resolved infusion from path " + __dirname + " to " + upInfusion.module.modules.infusion.baseDir);
-    module.exports = upInfusion;
-    return;
+    if (upInfusion && Object.keys(upInfusion).length != 0) {
```
colinbdclark: ESLint should catch a minor typo here.

src/module/fluid.js
```diff
@@ -38,9 +38,11 @@ var upPath = path.resolve(__dirname, "../../../../..");
...
+        upInfusion.log("Resolved infusion from path " + __dirname + " to " + upInfusion.module.modules.infusion.baseDir);
+        module.exports = upInfusion;
+        return;
+    } else { console.log(upInfusionPath + " resolved to current Infusion"); }
```
colinbdclark: I'm guessing this might also fail ESLinting because of a lack of line breaks around the block?

src/module/fluid.js
michelled: Thx! I fixed the issue.

src/module/fluid.js
michelled: It didn't, but I added line breaks.

src/module/fluid.js
```diff
@@ -38,9 +38,13 @@ var upPath = path.resolve(__dirname, "../../../../..");
 var upInfusionPath = resolveModuleSync("infusion", upPath);
 if (upInfusionPath) {
     upInfusion = require(upInfusionPath);
-    upInfusion.log("Resolved infusion from path " + __dirname + " to " + upInfusion.module.modules.infusion.baseDir);
-    module.exports = upInfusion;
-    return;
+    if (upInfusion && Object.keys(upInfusion).length !== 0) {
```
amb26: i) This check should have a substantial comment including a) the JIRA number, b) a rough description of the situation the problem occurs in and c) its mechanism of action
ii) A check for whether upInfusion.module is defined or not is a more conservative check

src/module/fluid.js
```diff
@@ -38,9 +38,13 @@ var upPath = path.resolve(__dirname, "../../../../..");
...
+        upInfusion.log("Resolved infusion from path " + __dirname + " to " + upInfusion.module.modules.infusion.baseDir);
+        module.exports = upInfusion;
+        return;
+    } else {
+        console.log(upInfusionPath + " resolved to current Infusion");
```
amb26: This condition is really the same as the one where Infusion is at top level, so we should just fall through here and display the message on line 49

michelled:
@amb26 I addressed your comments above, however, something strange has happened when I merged master into my branch. Let me know if you'd like me to make a new branch with this fix. It will probably mean a new pull request but at least it wouldn't show @jobara's changes mixed in.

amb26:
Closed in favour of #738

PR #736, opened 2016-08-15
FLUID-5872: Setting the iframe source programmatically

jobara:
When the iframe source was set in the document, it could load before the load event is bound.
This prevented the test from working as it was waiting for the load event.

https://issues.fluidproject.org/browse/FLUID-5872

### DIFF
--- a/tests/component-tests/tooltip/js/TooltipTests.js
+++ b/tests/component-tests/tooltip/js/TooltipTests.js
@@ -212,7 +212,6 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
     });
 
     fluid.tests.tooltip.FLUID5846.setupIframe = function (that, iframeSrc, iframe) {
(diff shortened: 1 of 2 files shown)


tests/component-tests/tooltip/js/TooltipTests.js
```diff
@@ -211,7 +211,8 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 
     });
 
-    fluid.tests.tooltip.FLUID5846.setupIframe = function (that, iframe) {
+    fluid.tests.tooltip.FLUID5846.setupIframe = function (that, iframeSrc, iframe) {
+        $(iframe).attr("src", iframeSrc);
```
cindyli: Please add a comment with why this is needed, with a link to the JIRA. Thanks.

cindyli:
Merged at da88fd7b88f4a60ab9a4f352fe1edbe2973afe10

PR #738, opened 2016-08-22
FLUID-5940: Correct problems with auto-deduping Infusion on Windows

amb26:
This pull request replaces #735 - in addition to the check made there, we also need to fix an underlying issue in the "resolve" module - see https://github.com/substack/node-resolve/issues/106 for details.

I've run the following test - firstly, the reproduction steps listed on #735 of checking out, for example, kettle in "C:\kettle", running npm install, and then npm test from the kettle directory itself. Also, I have run   https://github.com/GPII/universal with its infusion patched in-place to this version to check it starts up ok. Perhaps you can think of other tests. In general this issue is ludicrously hard to test because of its large number of environmental assumptions and dependencies, down to the use of a particular shell on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "bugs": "http://issues.fluidproject.org/browse/FLUID",
     "homepage": "http://www.fluidproject.org/",
     "dependencies": {
-        "resolve": "1.1.6"
+        "fluid-resolve": "1.2.0"
     },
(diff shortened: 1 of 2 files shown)


jobara:
merged at d82e14339fe975b9c2426f4138659fa45da16884

PR #744, opened 2016-09-01
FLUID-5370: Removed old tokens

jobara:
Removed old tokens that were preserved for backwards compatibility.

Docs PR: https://github.com/fluid-project/infusion-docs/pull/98

https://issues.fluidproject.org/browse/FLUID-5370

### DIFF
--- a/src/components/uploader/js/FileQueueView.js
+++ b/src/components/uploader/js/FileQueueView.js
@@ -394,7 +394,6 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
 
         strings: {
             progress: {
-                toUploadLabel: "To upload: %fileCount %fileLabel (%totalBytes)",
                 singleFile: "file",
                 pluralFiles: "files"
             },
--- a/src/components/uploader/js/FileQueueView.js
+++ b/src/components/uploader/js/FileQueueView.js
@@ -404,8 +404,8 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
                 remove: "Press Delete key to remove file"
             },
             errors: {
-                HTTP_ERROR: "File upload error: a network error occured or the file was rejected (reason unknown).",
-                IO_ERROR: "File upload error: a network error occured.",
+                HTTP_ERROR: "File upload error: a network error occurred or the file was rejected (reason unknown).",
+                IO_ERROR: "File upload error: a network error occurred.",
                 UPLOAD_LIMIT_EXCEEDED: "File upload error: you have uploaded as many files as you are allowed during this session",
                 UPLOAD_FAILED: "File upload error: the upload failed for an unknown reason.",
                 QUEUE_LIMIT_EXCEEDED: "You have as many files in the queue as can be added at one time. Removing files from the queue may allow you to add different files.",
--- a/src/components/uploader/js/Uploader.js
+++ b/src/components/uploader/js/Uploader.js
@@ -139,18 +139,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     };
 
     fluid.uploader.renderUploadTotalMessage = function (that) {
-        // Preservered for backwards compatibility, should be refactored post v1.5
-        var numReadyFiles = that.queue.getReadyFiles().length;
-        var bytesReadyFiles = that.queue.sizeOfReadyFiles();
-        var fileLabelStr = fluid.uploader.fileOrFiles(that, numReadyFiles);
-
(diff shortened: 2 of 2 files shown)


src/components/uploader/js/Uploader.js
```diff
@@ -139,18 +139,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
     fluid.uploader.renderUploadTotalMessage = function (that) {
-        // Preservered for backwards compatibility, should be refactored post v1.5
-        var numReadyFiles = that.queue.getReadyFiles().length;
-        var bytesReadyFiles = that.queue.sizeOfReadyFiles();
-        var fileLabelStr = fluid.uploader.fileOrFiles(that, numReadyFiles);
-
         var totalCount = that.queue.files.length;
         var noFilesMsg = that.options.strings.progress.noFiles;
-
         var totalStateStr = fluid.stringTemplate(that.options.strings.progress.toUploadLabel, {
```
cindyli: Please verify if this string definition of progress.toUploadLabel is still needed: https://github.com/fluid-project/infusion/blob/master/src/components/uploader/js/FileQueueView.js#L397. Thanks.

cindyli:
Remove the use of these old tokens from tests as well.

jobara:
@cindyli i didn't see any tokens in the tests.. could you please point me at something in particular that needs to be updated.

cindyli:
@jobara a token in the tests: https://github.com/fluid-project/infusion/blob/master/tests/component-tests/uploader/js/FileQueueTests.js#L315

src/components/uploader/js/Uploader.js
```diff
@@ -200,20 +191,11 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
-        var numFilesInBatch = batch.files.length;
-        var fileLabelStr = fluid.uploader.fileOrFiles(that, numFilesInBatch);
-
         var uploadingSize = batch.totalBytesUploaded + that.queue.sizeOfUploadedFiles();
 
         var totalProgressStr = fluid.stringTemplate(that.options.strings.progress.totalProgressLabel, {
-            curFileN: batch.fileIdx,
-            totalFilesN: numFilesInBatch,
-            fileLabel: fileLabelStr,
-            currBytes: fluid.uploader.formatFileSize(batch.totalBytesUploaded),
```
amb26: I'm concerned by these changes which appear to be removing functionality from the uploader as well as compatibility information. I'm also concerned that this information could be removed without breaking any tests...

src/components/uploader/js/FileQueueView.js
```diff
@@ -394,7 +394,6 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
 
         strings: {
             progress: {
-                toUploadLabel: "To upload: %fileCount %fileLabel (%totalBytes)",
```
amb26: This string was removed but the use of it wasn't. I think a better role for this pull request would be to add tests for the functionality that it has rather than removing the functionality

jobara: @amb26 could you point me at the location where it is used?

src/components/uploader/js/Uploader.js
jobara: @amb26 as of Infusion 1.5 these values were no longer used in any of the default strings, but they were left in place for backwards compatibility incase a user had custom strings here. 

See the strings section from:
http://docs.fluidproject.org/infusion/development/UploaderAPI.html#uploader-options

amb26: Yes - I don't see why we should remove them?

jobara:
@cindyli i believe those are arguments past into the test for the file queue itself and aren't related to the tokens removed for the strings in the uploader.

src/components/uploader/js/Uploader.js
jobara: My thought is that we don't use them so it becomes an added maintenance burden, plus a higher chance of breaking without us catching it. Also, I'd like to reduce the file size and javascript processing as much as possible. The more we keep building up the larger these inflate over time. I totally agree we should keep this between major releases, but I'd like to take the opportunity at a major release trim what we can.

jobara:
This was discussed at the Community Meeting on Sept 14, 2016. It was determined that the old values were still useful to integrators and that all relevant information that could be tokenized and used within a string template should be modelled and passed into the string template function. This has been captured in [FLUID-5957](https://issues.fluidproject.org/browse/FLUID-5957)

PR #751, opened 2016-09-14
FLUID-5719: update CDNs for rich text editor code

waharnum:
Partially deals with https://issues.fluidproject.org/browse/FLUID-5719.

Due to version changes in what InlineEditIntegrations is assuming vs. the latest versions:
- TinyMCE still throws errors, but partially works
- CKEditor seems fine

So it's still broken, but somewhat less broken (per discussion with @jobara)


jobara:
@waharnum I did some digging into the TinyMCE issues. 

The Tiny MCE 3.4.9 focus issue doesn't seem to be a problem anymore, and this code needs to be removed. https://github.com/waharnum/infusion/blob/FLUID-5719/src/components/inlineEdit/js/InlineEditIntegrations.js#L181-L194

Also the fluid.setCaretToEnd function can be removed and its use ( https://github.com/waharnum/infusion/blob/FLUID-5719/src/components/inlineEdit/js/InlineEdit.js#L181 ) replaced with `that.editField.focus()`. 

I think that should make everything work, I've tested it on Firefox, Safari, and Chrome in Mac OS but we'd also need to try it with IE 11 and Edge.

amb26:
Thanks, @jobara - that's a really helpful investigation. It will be great to be able to ditch that code.

waharnum:
In Edge & IE11 the changes seems to work fine for the manual tests (both for the Rich Text editor and the dropdown one). No more errors.

cindyli:
Merged at 68adc5f70fd520acb33e0d847fb6e567c86485fa

PR #752, opened 2016-09-21
FLUID-5825, FLUID-5966: Upgrade jQuery and jQuery UI 

jobara:
Upgraded to:
* jQuery v3.1
* jQuery UI v1.12.1

https://issues.fluidproject.org/browse/FLUID-5825
https://issues.fluidproject.org/browse/FLUID-5966


src/framework/core/js/FluidView.js
```diff
@@ -99,6 +99,7 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
      * @return a single-element jQuery of container
      */
     fluid.container = function (containerSpec, fallible, userJQuery) {
+        var selector = fluid.get(containerSpec, "selector") || containerSpec;
```
amb26: It's impossible that containerSpec is null - containerSpec.selector || containerSpec is ok here

src/framework/core/js/FluidView.js
```diff
@@ -119,6 +120,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             fluid.fail("fluid.container was supplied a non-jQueryable element");
         }
 
+        container.selector = selector;
```
amb26: We should add some explanation here of what effect this logic is meant to achieve - perhaps in the form of a JIRA. It should explain that the fidelity of this approach in recovering the selector is limited (especially if the node comes from outside the system) and that this can only be depended on in limited contexts. Should explicitly mention the code in the prefs framework which relies on this as it tries to cross the iframe boundary in separatedPanel.

src/framework/preferences/js/Panels.js
```diff
@@ -150,8 +150,11 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         // in order to reset the dom binder when panels are in an iframe.
         // It can be be eliminated once we have the new renderer.
         var userJQuery = that.container.constructor;
-        var oldContainer = that.container;
-        that.container = userJQuery(that.container.selector, oldContainer[0].ownerDocument);
+        var context = that.container[0].ownerDocument;
+        var selector = that.container.selector;
+        that.container = userJQuery(selector, context);
```
amb26: See above - it should be made clear that this code is related to that within the DOM binder and that the user can go to a single place, perhaps a JIRA, in order to understand the workflow. This JIRA should link to https://issues.fluidproject.org/browse/FLUID-4536 which is our previous work in this area.

ReleaseNotes.md
```diff
@@ -103,11 +103,11 @@ categorized by license:
 ### MIT License ###
 * [Foundation v5.4.5](http://foundation.zurb.com/index.html)
 * [HTML5 Boilerplate v4.3](http://html5boilerplate.com/)
-* [jQuery javascript library v1.11.0](http://jquery.com/)
-* [jQuery UI button, dialog, draggable, mouse, position, resizable, slider, tabs, and tooltip javascript widgets v1.10.4](http://ui.jquery.com/)
+* [jQuery v3.1.0](http://jquery.com/)
+* [jQuery UI (Core; Interactions: draggable, resizable; Widgets: button, checkboxradio, controlgroup, dialog, mouse, slider, tabs, and tooltip) v1.12.1](http://ui.jquery.com/)
```
amb26: Is this a particular build of jQuery UI that we depend on now? Could you leave instructions on how to produce it?

jobara: It's basically the same custom build we relied on before, except now I've explicitly mentioned all of the checkboxes that need to be marked off on their custom builder.

jobara: with the exception of core, because we take everything except the jQuery 1.7 compatibility file.

src/framework/preferences/js/Panels.js
jobara: This specific JIRA doesn't exist. I suppose we could file something after the fact if that's what you mean. Also, in the end I'm not sure this was specific to the iframe boundary but rather a product of recreating the container inside resetDomBinder method. Which is why the changes were needed here as well. I left the changes in `fluid.container` mostly to guard against the case where someone does something like `fluid.someComponent($(".someContainer"), {});`

amb26: Yes, we file it after the fact. There needs to be a clear paper trail leading from this code that someone can follow in order to gain an understanding of it.

jobara:
@amb26 I've updated the comments and addressed the assignment of the selector variable. Ready for more review.

PR #753, opened 2016-09-23
FLUID-5963: Using eslint config from eslint-config-fluid module

jobara:
Making use of the sharable eslint config via the eslint-config-fluid module

https://issues.fluidproject.org/browse/FLUID-5963

Depends on https://github.com/fluid-project/eslint-config-fluid/pull/1
Also needs to be updated to point at the npm package instead of my branch for the eslint-config-fluid dependency.


jobara:
@amb26 could you please review this along with https://github.com/fluid-project/eslint-config-fluid/pull/1 

I've opened up commits from maintainers to this branch, so you can push up a change to point at the npm package for the eslint-config-fluid dependency if you publish that one.

PR #757, opened 2016-09-29
FLUID-5973: conditional testing utilities based on context awareness or promise results

waharnum:
@jobara - an initial pass at being able to skip tests (or run alternate tests) based on platform.

I also took the opportunity to move some conditional test features that were in the `fluid.tests.textToSpeech` namespace into a generalized `fluid.tests.conditionalTestUtils` namespace.


tests/test-core/utils/js/ConditionalTestUtils.js
```diff
@@ -0,0 +1,87 @@
...
+    "use strict";
+
+    fluid.registerNamespace("fluid.tests.conditionalTestUtils");
+
+    // "platformArg" is a string or array of strings of platform names to detect
+    // This returns a boolean based on whether or not the string is
+    // present in the navigator.platform string
+
+    fluid.tests.conditionalTestUtils.isPlatform = function (platformArg) {
+        var detectedPlatform = navigator.platform;
```
amb26: This implementation should be rewritten so it can bind to any material configured by the Infusion ContextAwareness scheme rather than being hardwired to check for strings within navigator.platform (and this particular case of it bundled as ContextAwareness checks)

amb26: This will also aid you in writing test cases for this facility itself - otherwise it will be rather hard to mock!

waharnum: @amb26 - I think the latest push is probably more like what you're talking about, but I don't think it's quite right - not sure I understand what is meant by "bind to any material configured by the Infusion ContextAwareness scheme". Can you clarify?

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -224,38 +242,50 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-        jqUnit.asyncTest(wrapperMessage, function () {
-            jqUnit.expect(0);
-            task().then(function () {
-                jqUnit.start();
-                resolveFunc(resolveMessage);
-            }, function () {
-                jqUnit.start();
-                rejectFunc(rejectMessage);
-            });
+    fluid.tests.textToSpeech.runTests = function (that) {
```
amb26: This is generic and can go into your testUtils - factor off a grade from fluid.tests.textToSpeech.testRunner to account for the fact that it has certain material in the option named "tests" and document its structure

tests/test-core/utils/js/ConditionalTestUtils.js
```diff
@@ -0,0 +1,81 @@
...
+
+(function () {
+    "use strict";
+
+    fluid.registerNamespace("fluid.test.conditionalTestUtils");
+
+    // "platformArg" is a string to check against navigator.platform
+
+    fluid.test.conditionalTestUtils.isPlatform = function (platformArg) {
+        var detectedPlatform = navigator.platform ? navigator.platform : undefined;
```
amb26: Bind to the string you registered in ContextAwareness rather than navigator.platform literally so there's the potential to mock this out

tests/test-core/utils/js/ConditionalTestUtils.js
```diff
@@ -0,0 +1,81 @@
...
+    // the promise. Without this, QUnit will merrily proceed along to the
+    // next test set, which can cause errors various contexts including the
+    // all-tests runner.
+    //
+    // wrapperMessage, task, resolveFunc, rejectFunc: required
+    // "task" must be a function returning a promise
+    // resolveMessage, rejectMessage: optional strings, passed to the test
+    // functions
+
+    fluid.test.conditionalTestUtils.chooseTestByPromiseResult = function (wrapperMessage, task, resolveFunc, rejectFunc, resolveMessage, rejectMessage) {
```
amb26: I think that this function is sufficiently meaty that we should have a test case for it

tests/component-tests/textToSpeech/js/TextToSpeechTests.js
```diff
@@ -224,38 +242,50 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
         jqUnit.assertFalse("Shouldn't be paused", that.model.paused);
     };
 
-    fluid.tests.textToSpeech.bypassTest = function (bypassMessage) {
-        jqUnit.test("Tests were skipped.", function () {
-            jqUnit.assert(bypassMessage);
-        });
+    fluid.tests.textToSpeech.baseTests = function () {
+        fluid.test.conditionalTestUtils.chooseTestByPromiseResult("Confirming if TTS is available for initialization and start/stop tests", fluid.textToSpeech.checkTTSSupport, fluid.tests.textToSpeech.ttsTestEnvironment, fluid.test.conditionalTestUtils.bypassTest, "Browser appears to support TTS", "Browser does not appear to support TTS");
```
amb26: Outrageously long line, even by my standards : P

tests/test-core/utils/js/ConditionalTestUtils.js
waharnum: What's the appropriate approach to referring to a value held at a ContextAwareness path in a global function in that manner?

tests/test-core/utils/js/ConditionalTestUtils.js
amb26: I think you have it very nearly right. Given you are now governed by a contextAwareness block this is now ready to bind to anything. I've just proposed a few tweaks and refactoring based on modularity/using the context scheme a bit more, and suggested some tests

amb26:
Can we have a nice title for the pull request - cheers!

tests/test-core/utils/js/ConditionalTestUtils.js
amb26: You could call this function: https://github.com/fluid-project/infusion/blob/master/src/framework/enhancement/js/ContextAwareness.js#L134

amb26: Sorry that its docs status is so grey - I see it is not marked as NON-API but unfortunately does not appear in http://docs.fluidproject.org/infusion/development/ContextAwareness.html

waharnum: @amb26 thanks - should that function be documented as part of the public API? I can make a separate JIRA for that so that we can address it at some point.

waharnum: @amb26 that function still seems to require a `that-type` reference; I think what I'm looking for (possibly incorrectly) is a way to have something like the following:

```
fluid.contextAware.getContextAwareCheckValue("fluid.platform");
```

Does that make sense, or is that an inadvisable approach?

amb26: https://issues.fluidproject.org/browse/FLUID-5975

amb26: A location is the component tree is always required in order to make a context check. If you are only interested in globally resolvable contexts, you can supply fluid.rootComponent

waharnum:
@amb26 I've added test coverage for both the `fluid.test.conditionalTestUtils.chooseTestByPromiseResult` utility function and the `fluid.test.conditionalTestUtils.contextAwareTestRunner` grade. I think this is ready for another look.

michelled:
@waharnum can you please address the conflict? Also, once that is done, is this ready for more review?

waharnum:
@michelled conflict addressed & tests for the conditional test utilities restructured into their own files - this should be ready for review by @amb26

PR #720, opened 2016-06-16
FLUID-5917: Makes Testem generate a TAP report

avtar:
These changes make Testem generate a TAP report that can be used by Jenkins.

PR #760, opened 2016-10-06
FLUID-5978: Infusion build distributions

jobara:
Adding a grunt task and prepublish script for building a dist directory.

https://issues.fluidproject.org/browse/FLUID-5978


Gruntfile.js
```diff
@@ -259,7 +308,49 @@ module.exports = function (grunt) {
...
+
+        if (options.exclude || options.include) {
+            options.target = "custom";
+        }
+
+        setBuildSettings(options);
+
+        var concatTask = options.compress ? "uglify:" : "concat:";
+        var tasks = [
+            "clean:build",
```
amb26: Thanks, @jobara - this pull looks generally good. Whilst we are touching the build scripts in this area, I thought it might be appropriate to fix up our clean:build strategy a little more:

Since we merged in https://github.com/fluid-project/infusion/pull/720 I feel we've gone from deleting too much build junk to deleting too little. We now accumulate an entire copy of the all the frameworks's JS source in build/ after a build, even though it is all included within one of the build files themselves.
Could we perhaps have a clean:postBuild task which, say, deletes something like build/src/**.js - these redundant files are very confusing in an IDE, since as well as having one copy of all the framework's source per project that depends on it, we are now likely to double this by each of those copies now having an extra copy within build/src - Cheers

amb26: The third option would be nice, if it didn't seem to involve too much work - otherwise we could just leave things are they are

jobara:
@amb26 I believe I have option 3 working and also did some cleanup after distributions are built. This is ready for more review.

amb26:
Thanks, @jobara - this looks great! Now that we are using the prepublish lifecycle script, we should now make sure to update your pull https://github.com/fluid-project/fluid-publish/pull/5 so that it executes this step (if it exists in package.json) before publishing artefacts to npm.

PR #761, opened 2016-10-10
FLUID-5971: Migrate gpii-binder transforms to infusion... 

the-t-in-rtf:
As [requested](https://github.com/GPII/gpii-binder/pull/4#issuecomment-249723223), this PR moves various [`gpii-binder`](https://github.com/GPII/gpii-binder) transformation functions into infusion.

See [FLUID-5971](https://issues.fluidproject.org/browse/FLUID-5971) for details.


the-t-in-rtf:
Please note, there are failing tests for other parts of infusion in Firefox, but the same tests fail with master and both OS X and Linux, so I assume these are known to the group.  The tests related to the changes made here definitely pass.

the-t-in-rtf:
The other half of this is to remove the transforms from gpii-binder:

https://github.com/GPII/gpii-binder/pull/5

That should be merged after this one, so that I can replace the git/hash reference in that package.json with a real version.

jobara:
@the-t-in-rtf which tests are failing in Firefox? I know there is one regarding text-to-speech in Linux, but am not aware of anything failing in macOS.

the-t-in-rtf:
When I run `testem ci --file tests/testem.json --launch firefox`,  I get the following failures with Firefox 49.0 on OS X:

```
not ok 6 Firefox 49.0 - Framework Tests: ./framework-tests/core/html/Keyboard-a11y-test.html
    ---
        stack: >
            runLoggingCallbacks@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:1609:4
            assert.ok@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:544:3
            pok@http://localhost:7357/1261/tests/test-core/jqUnit/js/jqUnit.js:110:9
            jsUnitCompat.assertTrue@http://localhost:7357/1261/tests/test-core/jqUnit/js/jqUnit.js:143:13
            @http://localhost:7357/1261/tests/framework-tests/core/js/Keyboard-a11yTests.js:567:9
            Test.prototype.run@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:207:4
            run/<@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:365:5
            process@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:1457:4
            QUnit.start/<@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:483:5
            
        message: >
            keyboard-a11y: Leaving container: onLeaveContainer: On leave is called when tabbing out of the selectables container.
        Log: |
    ...

...

not ok 9 Firefox 49.0 - Framework Tests: ./framework-tests/core/html/FluidView-test.html
    ---
        stack: >
            runLoggingCallbacks@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:1609:4
            assert.ok@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:544:3
            pok@http://localhost:7357/1261/tests/test-core/jqUnit/js/jqUnit.js:110:9
            jsUnitCompat.assertTrue@http://localhost:7357/1261/tests/test-core/jqUnit/js/jqUnit.js:143:13
            blurTest/</blurOutwaiter@http://localhost:7357/1261/tests/framework-tests/core/js/FluidViewTests.js:244:21
            
        message: >
            Fluid View Tests: Dead man's blur test - nonExcluded component one: nonExcluded component one - Blur handler has not executed
        Log: |
    ...

...

not ok 39 Firefox 49.0 - Component Tests: ./component-tests/pager/html/PagedTable-test.html
    ---
        actual: >
            
        expected: >
            Max depth.
        stack: >
            runLoggingCallbacks@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:1609:4
            .push@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:958:3
            assert.propEqual@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:579:3
            jsUnitCompat.assertDeepEq@http://localhost:7357/1261/tests/test-core/jqUnit/js/jqUnit.js:178:17
            fluid.tests.tooltip.assertVisible@http://localhost:7357/1261/tests/component-tests/tooltip/js/TooltipTests.js:38:9
            fluid.tests.assertVisibleTips@http://localhost:7357/1261/tests/component-tests/pager/js/PagedTableTests.js:263:9
            fluid.tests.verifyPagerToolTip@http://localhost:7357/1261/tests/component-tests/pager/js/PagedTableTests.js:281:9
            fluid.tests.verifyToolTipsOfPagerBar@http://localhost:7357/1261/tests/component-tests/pager/js/PagedTableTests.js:271:13
            fluid.test.decoders.func/<.execute@http://localhost:7357/1261/tests/test-core/utils/js/IoCTestUtils.js:275:17
            fluid.test.execExecutor@http://localhost:7357/1261/tests/test-core/utils/js/IoCTestUtils.js:434:9
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/1261/tests/test-core/utils/js/IoCTestUtils.js:525:21
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/1261/tests/test-core/utils/js/IoCTestUtils.js:526:21
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/1261/tests/test-core/utils/js/IoCTestUtils.js:526:21
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/1261/tests/test-core/utils/js/IoCTestUtils.js:526:21
            fluid.test.processTestCase/</testFunc@http://localhost:7357/1261/tests/test-core/utils/js/IoCTestUtils.js:605:21
            Test.prototype.run@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:207:4
            run/<@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:365:5
            process@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:1457:4
            QUnit.start/<@http://localhost:7357/1261/tests/lib/qunit/js/qunit.js:483:5
            
        message: >
            Pager tooltip tests: Tooltip visibility and contents: The tooltip of page link 1, in top page bar is visible - at sequence position 4 of 15
        Log: |
```

In addition, it hangs outright on "Test Including Pause and Resume Events".

the-t-in-rtf:
If I create the standard Vagrant VM for the package there is an outright failure:

```
not ok 39 Firefox 49.0 - Component Tests: ./component-tests/pager/html/PagedTable-test.html
    ---
        actual: >
            
        expected: >
            Max depth.
        stack: >
            runLoggingCallbacks@http://localhost:7357/6586/tests/lib/qunit/js/qunit.js:1609:4
            .push@http://localhost:7357/6586/tests/lib/qunit/js/qunit.js:958:3
            assert.propEqual@http://localhost:7357/6586/tests/lib/qunit/js/qunit.js:579:3
            jsUnitCompat.assertDeepEq@http://localhost:7357/6586/tests/test-core/jqUnit/js/jqUnit.js:178:17
            fluid.tests.tooltip.assertVisible@http://localhost:7357/6586/tests/component-tests/tooltip/js/TooltipTests.js:38:9
            fluid.tests.assertVisibleTips@http://localhost:7357/6586/tests/component-tests/pager/js/PagedTableTests.js:263:9
            fluid.tests.verifyPagerToolTip@http://localhost:7357/6586/tests/component-tests/pager/js/PagedTableTests.js:281:9
            fluid.tests.verifyToolTipsOfPagerBar@http://localhost:7357/6586/tests/component-tests/pager/js/PagedTableTests.js:271:13
            fluid.test.decoders.func/<.execute@http://localhost:7357/6586/tests/test-core/utils/js/IoCTestUtils.js:275:17
            fluid.test.execExecutor@http://localhost:7357/6586/tests/test-core/utils/js/IoCTestUtils.js:434:9
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/6586/tests/test-core/utils/js/IoCTestUtils.js:525:21
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/6586/tests/test-core/utils/js/IoCTestUtils.js:526:21
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/6586/tests/test-core/utils/js/IoCTestUtils.js:526:21
            fluid.test.sequenceExecutor/that.execute@http://localhost:7357/6586/tests/test-core/utils/js/IoCTestUtils.js:526:21
            fluid.test.processTestCase/</testFunc@http://localhost:7357/6586/tests/test-core/utils/js/IoCTestUtils.js:605:21
            Test.prototype.run@http://localhost:7357/6586/tests/lib/qunit/js/qunit.js:207:4
            run/<@http://localhost:7357/6586/tests/lib/qunit/js/qunit.js:365:5
            process@http://localhost:7357/6586/tests/lib/qunit/js/qunit.js:1457:4
            QUnit.start/<@http://localhost:7357/6586/tests/lib/qunit/js/qunit.js:483:5
            
        message: >
            Pager tooltip tests: Tooltip visibility and contents: The tooltip of page link 1, in top page bar is visible - at sequence position 4 of 15
        Log: |
    ...
```

In addition, the tests hang in the standard linux VM on "Image Reorderer Tests: ItemFocusBlur: with click".

the-t-in-rtf:
@jobara, see my two previous comments for the test failures.  You should at least be able to reproduce the problems with Vagrant.

jobara:
@the-t-in-rtf i'm a bit unclear if the first example is being run in a vagrant VM or directly on OSX. Could you clarify that. Would you also be able to write these up as Infusion JIRAs?

the-t-in-rtf:
@jhung, as mentioned in the comments, the first one is OS X, and the second one is in Vagrant.

the-t-in-rtf:
As long as one other human can reproduce this, I am happy to write it up.

jobara:
@the-t-in-rtf did you get your hanging results between the OSX run and Vagrant run mixed up? If this is the case I believe the issue is that with the testem launch command it launches Firefox without giving it focus. However some of the tests require the browser to have focus. I manually placed focus on firefox when it launched and did not receive and failures. Could you verify this on your end as well?

Regarding the Vagrant test run. I wasn't able to reproduce the failure for test 39. I started with a new Vagrant build though, having destroyed the previous one. The hanging on the text to speech pause and resume tests should only occur in linux and will be addressed by PR https://github.com/fluid-project/infusion/pull/757

the-t-in-rtf:
It hung on both, even in Vagrant, where FF spawns in the foreground.  Are the testem tests run against master on a CI instance?

the-t-in-rtf:
With the focus on FF in OS X, I get all the way to the TTS tests, and then my browser eventually dies.  The last test fails, but only the first 51 were run.

the-t-in-rtf:
I think the TTS problems on my machine may be related to the [ModelTalker](modeltalker.org) engine I use, I'll turn that off and try again.

the-t-in-rtf:
The tests pass on OS X with focus on the FF window and with the system TTS.

the-t-in-rtf:
The PagedTable failures seem intermittent, I don't get them on every run.  I am running multiple passes to see if I can get an idea of how often or what (if any) differences there are.

I always have to hit "rerun" to get past the linux problems with TTS, but that's expected.

the-t-in-rtf:
OK, after a few consistent runs, I'm willing to chalk up the PagedTable failures to lack of FF focus.  Anyway, to confirm, the changes proposed here do not result in new test failures or brittleness, and should be ready to review.

jobara:
@the-t-in-rtf thanks for all the testing and investigation into the tests.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -1,7 +1,11 @@
 /*
 Copyright 2010 University of Toronto
 Copyright 2010-2011 OCAD University
+<<<<<<< HEAD
+Copyright 2013-2016 Raising the Floor - International
+=======
 Copyright 2013, 2016 Raising the Floor - International
+>>>>>>> upstream/master
```
jobara: Looks like you accidentally committed the merge conflict diff.

jobara:
@the-t-in-rtf I think we'll also need to coordinate with this docs PR https://github.com/fluid-project/infusion-docs/pull/86 to document these transforms.

the-t-in-rtf:
@jobara, sounds good.  How close is the other to merging?  If it's likely to be merged this week, I'll wait, otherwise I can submit a PR against that branch.

jobara:
@the-t-in-rtf regarding https://github.com/fluid-project/infusion-docs/pull/86 I'm not sure when it will be ready exactly. It seems that it is waiting on some changes from @kaspermarkus.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
...
+     * returned by HTML5 date and dateTime inputs.
+     *
+     * A string that cannot be parsed will be treated as `undefined`.
+     *
+     * @param value - The String value to be transformed into a Date object.
+     * @returns {Date} - A date object, or `undefined`.
+     */
+    fluid.transforms.stringToDate = function (value) {
+        var date = new Date(value);
+        return isNaN(date.getTime()) ? undefined : date;
```
jobara: @amb26 I would agree with avoiding coercion, but I don't think this is type coercion. Rather `valueOf` method on the date object returns a numbered value. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf

amb26: @jobara - of course this is coercion - what do you imagine the implicit invocation of a "valueOf" method means? : P. Consult the section "Confusion special-case behaviour" on this page:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN
Once ECMAScript 6 is widespread, we should adopt the use of Number.isNaN as a best practice, which skips the coercion. In the meantime, we should just be careful to only call such builtin functions with their native types.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
...
+     *
+     * Convert any value into a stringified boolean, i. e. either "true" or "false".  Anything that evaluates to
+     * `true` (1, `true`, "non empty string", {}, et. cetera) returns "true".  Anything else (0, `false`, `null`)
+     * returns "false".
+     *
+     * @param value - The value to be converted to a stringified Boolean.
+     * @returns {string} - A stringified boolean representation of the value.
+     */
+    fluid.transforms.booleanToString = function (value) {
+        return value ? "true": "false";
```
the-t-in-rtf: Fixed.  I also found someone else's linting errors in the IoC tests and fixed those.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
...
+     * }
+     * ```
+     *
+     * The default value for `space` is 0, which disabled spacing and line breaks.
+     *
+     * @param value {Object} - An object to be converted to stringified JSON.
+     *
+     */
+    fluid.transforms.objectToJSONString = function (value, transformSpec) {
+        var space   = transformSpec.space || 0;
```
the-t-in-rtf: Fixed.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
             (transformSpec.max === undefined ||  transformSpec.max >= value) ? true : false;
     };
 
+    /**
+     *
+     * Convert a string to a Boolean.
+     *
+     * The following are all `false`: `undefined`, `null`, "", "0", "false"
```
the-t-in-rtf: Hi, @jobara.  Agreed on using JSON syntax rather than GfM to distinguish strings from non-strings.

As to the use case, this function was designed to help in relaying HTML form element values, which are always strings.

However, I agree that it's easy to imagine people stuffing other "falsy" material in.  I can see two ways to clarify.  One is to throw an error if something other than a string (or undefined/null) is passed.  The other (which seems less harsh) is to handle `false` and `0`, which I have done in the latest commit.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
...
+    };
+
+    fluid.defaults("fluid.transforms.stringToDate", {
+        gradeNames: ["fluid.standardTransformFunction"]
+    });
+
+    /**
+     *
+     * Transform a Date object into a date string using its toISOString method.  Results in date strings that are suitable
+     * for use with both HTML5 "date" inputs and JSON Schema "date" format string validation.
```
the-t-in-rtf: Done.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
...
+     *
+     * A non-date object will be treated as `undefined`.
+     *
+     * @param value - The Date object to be transformed into an ISO 8601 string.
+     * @returns {String} - A {String} value representing the date, or `undefined` if the date is invalid.
```
the-t-in-rtf: @jobara, the `instanceOf` check should be enough to ensure that no error is thrown.  I have added tests that confirm this for both date to string transforms.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
...
+
+    /**
+     *
+     * Transform a Date object into a date/time string using its toISOString method.  Results in date strings that are
+     * suitable for use with both HTML5 "dateTime" inputs and JSON Schema "date-time" format string validation.
+     *
+     * A non-date object will be treated as `undefined`.
+     *
+     * @param value - The Date object to be transformed into an ISO 8601 string.
+     * @returns {String} - A {String} value representing the date and time, or `undefined` if the date/time are invalid.
```
the-t-in-rtf: Done.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -739,4 +739,150 @@ var fluid = fluid || fluid_2_0_0;
...
+    };
+
+    fluid.defaults("fluid.transforms.stringToBoolean", {
+        gradeNames: ["fluid.standardTransformFunction"]
+    });
+
+    /**
+     *
+     * Convert any value into a stringified boolean, i. e. either "true" or "false".  Anything that evaluates to
+     * `true` (1, `true`, "non empty string", {}, et. cetera) returns "true".  Anything else (0, `false`, `null`)
```
the-t-in-rtf: Done.

the-t-in-rtf:
Sorry for the delay in responding to your feedback, @jobara.  I have addressed all the outstanding feedback.  As [the docs PR](https://github.com/fluid-project/infusion-docs/pull/86) has been merged, I will fork that and submit a PR to cover the new functions.

jobara:
@the-t-in-rtf it seems that there is a merge conflict with this PR. Could you please have a look at that?

jobara:
@the-t-in-rtf i wonder if you also forgot to push up some changes. I'm not noticing some of the fixes you've mentioned and it looks like the last commit was from Oct 17.

the-t-in-rtf:
Sigh.  I had to create a branch to combine this branch with the fix for FLUID-5585, I suspect the updates are stored there. I will untangle things shortly.

the-t-in-rtf:
All conflicts are fixed, but this now includes Antranig's fix for FLUID-5585.  @jobara, to keep our sanity, let's review once [the PR for FLUID-5585](https://github.com/fluid-project/infusion/pull/577) is merged.

jobara:
@the-t-in-rtf thanks for sorting things out and letting me know about the dependency. Please ping me again on this once FLUID-5585 is in, if I haven't looked at it already.

the-t-in-rtf:
@jobara, Cindy [just merged the work done for FLUID-5585](https://github.com/fluid-project/infusion/pull/577).  We should be ready to continue here.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -740,4 +740,159 @@ var fluid = fluid || fluid_2_0_0;
...
+     * The following are all false: undefined, null, "", "0", "false", false, 0
+     *
+     * Everything else is true.
+     *
+     * @param value {String} The value to be interpreted.
+     * @returns {Boolean} The interpreted value.
+     */
+    fluid.transforms.stringToBoolean = function (value) {
+        if (value) {
+            return !Boolean(value === "" || value === "0" || value === "false");
```
amb26: The empty string is already falsy, so the first branch of this condition can never be hit https://developer.mozilla.org/en-US/docs/Glossary/Falsy

amb26: Also !Boolean is superfluous - it's sufficient to write return value !== "0" && value !== "false" which I think is rather clearer

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -740,4 +740,159 @@ var fluid = fluid || fluid_2_0_0;
...
+        }
+        else {
+            return false;
+        }
+    };
+
+    fluid.defaults("fluid.transforms.stringToBoolean", {
```
amb26: You should register these two transforms as inverses of each other, and supply tests that the inversion is (sufficiently) faithful, as with the other ModelTransformationTests later in the file

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -740,4 +740,159 @@ var fluid = fluid || fluid_2_0_0;
...
+     * "": {
+     *   transform: {
+     *     funcName: "fluid.transforms.objectToJSONString",
+     *     inputPath: "",
+     *     space: 2
+     *   }
+     * }
+     * ```
+     *
+     * The default value for `space` is 0, which disabled spacing and line breaks.
```
amb26: disabled -> disables

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -740,4 +740,159 @@ var fluid = fluid || fluid_2_0_0;
...
+     *
+     * @param value {Object} - An object to be converted to stringified JSON.
+     *
+     */
+    fluid.transforms.objectToJSONString = function (value, transformSpec) {
+        var space = transformSpec.space || 0;
+        return JSON.stringify(value, null, space);
+    };
+
+    fluid.defaults("fluid.transforms.objectToJSONString", {
```
amb26: Similarly, these two should be inverse pairs, as well as the Date transform below

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -740,4 +740,159 @@ var fluid = fluid || fluid_2_0_0;
...
+        gradeNames: ["fluid.standardTransformFunction"]
+    });
+
+    /**
+     *
+     * Transform a string to a date using the Date constructor.  Accepts (among other things) the date and dateTime values
```
amb26: There is a kind of "moral direction" in the framework which suggests that only JSON-serialisable material should be stored in a model. However, I'm loth to impose this as a condition, since there may be important use cases or servings of convenience where users should be allowed to store what they want there, assuming they understand the implications and are happy with the consequences. Right now the consequences are relatively harmless in that the framework (should) treat any non-plain object as immutable. This may change over time when we start to produce more mature authoring environments that expect to serialize everything in a model to disk or network and restore it faithfully on the other side. This time is some distance away, and we may well find ourselves putting schemas on models in order to announce their level of faithfulness, in which case the use of transforms like this could be done on a more formally advertised basis.
You can see the evolution of our thinking on this rather profound issue in the following references: http://docs.fluidproject.org/infusion/development/FrameworkConcepts.html#model-objects which was itself migrated from the rather more laissez-faire and primitive https://wiki.fluidproject.org/display/Infusion13/Model+Objects

So in general I'm happy for these transforms to go in, but I would like a kind of "health warning" comment added both to this comment and to the infusion-docs pull that use of these transforms implies the use of a non-JSON serializable model, which the user does at their own risk, since future versions of the framework may require extra formality in order to allow, or even forbid completely in certain contexts.

the-t-in-rtf: We should discuss further, as long-term I think we need a separate transformation API that is not tightly coupled to what we consider "acceptable" model material.  In this case, we only expand our thinking very slightly, to include transforms that assist in moving material in and out of the model.

I would propose adding text in the comment and to the docs that makes it clear that this is the intended use for these functions.   They are intended to serialise dates as strings before placing them in the model, and to reconstitute date strings stored in the model when they need to be used as date objects elsewhere.

src/framework/core/js/ModelTransformationTransforms.js
the-t-in-rtf: Done.

src/framework/core/js/ModelTransformationTransforms.js
the-t-in-rtf: Fixed.

src/framework/core/js/ModelTransformationTransforms.js
the-t-in-rtf: Fixed.

src/framework/core/js/ModelTransformationTransforms.js
the-t-in-rtf: I have made a first pass at this.  I used fairly strong wording to discourage storing Date objects in the model, and linked to the docs.  I hope that will be enough to address the concerns.

the-t-in-rtf:
I have made the requested changes and updated the docs pull as well.  @amb26, this is ready for further review.

tests/framework-tests/core/js/ModelTransformationTests.js
```diff
@@ -5127,4 +5127,311 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            method: "assertDeepEq",
+            model: { stringifiedObject: "{\"foo\":\"bar\"}" },
+            invertedRules: {
+                transform: [{
+                    type: "fluid.transforms.objectToJSONString",
+                    outputPath: "stringifiedObject",
+                    inputPath: "outie"
+                }]
+            },
+            fullyInvertible: true
```
amb26: Excellent

src/framework/core/js/ModelTransformationTransforms.js
amb26: Cheers - I think this is very nearly right. I agree that there needs to be some extension of the model idiom that assists more in dealing with "the outside world" - Clayton's wording for this has been "materialisation" which we've fretted about a bit since it implies that things inside the model are immaterial and things outside it aren't - which we don't quite agree with. I think we should just be a bit clearer about the status quo - it's not that the framework doesn't guarantee that Dates are not handled correctly now - in fact it does - it's just that they will probably not be handled correctly in future without some further formality. The docs you reference bear this out - "a tree of objects has the status of being a model if it can successfully be copied by the framework utility function fluid.copy()" - which *can* include Date objects if the user is happy to treat these as immutable. But we need to suggest that i) this is just the kind of subtlety that we need to indicate that the user will need to take care of for themselves, and ii) that this practice should be considered in some sense "already deprecated" in that it might cause their code to break in the face of future framework changes.
I guess what you're getting at with your "outside the model" is that we plan to licence this use via the "fluid.model.transformWithRules" API for static documents - which is certainly one possible usage. But all the same provisos apply there too - if the user puts Dates into their static documents, they will need to be prepared to tangle the fact that they will end up aliased to any equivalents in the output, which is just the same issue with models. So I'm not sure that the "outside the model" hedge really makes it clear to the users what they are letting themselves in for.

the-t-in-rtf: @amb26, I have committed a revised wording for all three, please review.  Once we agree on acceptable wording, I will make the same change in the docs pull.

PR #762, opened 2016-10-12
FLUID-5042: Enabling drag&drop for reorderer with touch

jobara:
Using touchPunch to map touch events to mouse events.

https://issues.fluidproject.org/browse/FLUID-5042


cindyli:
Shall touchPunch also be included into [this reorderer example](https://github.com/fluid-project/infusion/blob/master/examples/components/reorderer/tableRow/index.html)?

cindyli:
Merged at f36167b2fb5481a405643cc79c67ee76d67b3eab

PR #764, opened 2016-10-16
FLUID-5982: Allow injected components to be designated via expanders

amb26:


michelled:
@amb26: @jobara is waiting on either more tests or an explanation on why they aren't needed for this pull request

amb26:
Hi @jobara - I'm not quite sure what you mean. The targetComponent isn't instantiated here - the expander designates that the already instantiated component "sourceComponent" should be injected in its place. So as far as I can see, we test the functionality that is implemented.

jobara:
@amb26 I guess I mean is it a real component versus just having the options copied over, but I think i see that it access it off of that. So perhaps that is enough. However I have some more questions about how it works and things that could be tested.

- Does targetComponent act as an alias for sourceComponent? If i make a change to a member in one, does it change the other? 
- Do they have the same component ID's?

amb26:
@jobara - http://docs.fluidproject.org/infusion/development/SubcomponentDeclaration.html#injected-subcomponent-declaration

jobara:
merged at de9be1200c491fed7d951298ad866de54ef7f7e7

PR #765, opened 2016-10-18
FLUID-5980: Adjusting the height calculation.

jobara:
From the jQuery height docs (http://api.jquery.com/height/)
"As of jQuery 1.8, this may require retrieving the CSS height plus box-sizing property and then subtracting any potential border and padding on each element when the element has box-sizing: border-box. To avoid this penalty, use .css( "height" ) rather than .height()."

https://issues.fluidproject.org/browse/FLUID-5980

### DIFF
--- a/src/components/inlineEdit/js/InlineEdit.js
+++ b/src/components/inlineEdit/js/InlineEdit.js
@@ -221,8 +221,10 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             content: that.options.tooltipText,
             position: {
                 my: "left top",
-                at: "left bottom",
-                offset: "0 5"
+                at: "left bottom+25%", // add a 25% offset to keep the tooltip from overlapping the element it is for
+                // setting the "of" property to ensure that the tooltip is positioned relative to that.viewEl
+                // even when keyboard focus is on that.textEditButton
+                of: that.viewEl
             },
             target: "*",
             delay: that.options.tooltipDelay,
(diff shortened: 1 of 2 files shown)


src/components/inlineEdit/js/InlineEdit.js
```diff
@@ -353,8 +353,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             editModeInstruction.show();
 
             var editFieldPosition = editField.offset();
+            var editFieldHeight = parseInt(editField.css("height"), 10);
```
cindyli: Can you add a code comment with the jQuery height docs to explain why `css("height")` is preferred.

cindyli:
This pull request does fix the issue that the inline edit instruction field overlaps the input field, but doesn't improve the same issue with the tooltip overlapping on the input field, which is also stated in the https://issues.fluidproject.org/browse/FLUID-5980.

jobara:
@cindyli i added the comment and fixed the tooltip position. Ready for more review.

cindyli:
Thanks for the fix, @jobara. One thing I noticed is that, when hovering over or tapping onto an inline edit input field, the tooltip shows up after a second, which is fine. However, after the tooltip shows up, if hitting enter or clicking on the input field to go inside the field, the instruction box shows up right away while the tooltip stays and goes away after a couple of seconds. This causes the tooltip overlaps the instruction box during these a few seconds. We probably should talk to designers to figure out if this is the desired behavior. If it is not desired, a separate JIRA should be created. 

This pull request does fix the overlapping issue over the input field. Thanks.

cindyli:
Merged at 12acc5724694d703fddb48b441d4649a27c76190

PR #766, opened 2016-10-20
FLUID-5985: Fix trunk-breaking bug introduced by FLUID-5973 in non-browser contexts

waharnum:
This fixes the bug introduced by FLUID-5973 when running Infusion in non-browser contexts, as mentioned by @amb26 at https://github.com/fluid-project/infusion/pull/577#issuecomment-255065212 & encountered by @the-t-in-rtf 

The `fluid.contextAware.getBrowserPlatformName` function now makes sure a `navigator` object actually exists before attempting to access any properties.

I've also updated various variable and function names to make it clear this functionality only currently extends to the OS platform as reported in the browser context.

Confirmed both the in-browser tests and `basic-node-tests.js` are now passing.

### DIFF
--- a/src/framework/enhancement/js/ContextAwareness.js
+++ b/src/framework/enhancement/js/ContextAwareness.js
@@ -207,7 +207,13 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     // Context awareness for the platform (operating system)
 
     fluid.contextAware.getPlatformName = function () {
-        return navigator.platform ? navigator.platform : undefined;
+        // Gets platform name in browser contexts
+        if (navigator) {
+            return navigator.platform ? navigator.platform : undefined;
+        // Gets platform name in node contexts    
+        } else if (os.platform) {
+            return os.platform();
+        }
     };
 
     fluid.contextAware.makeChecks({
--- a/src/framework/enhancement/js/ContextAwareness.js
+++ b/src/framework/enhancement/js/ContextAwareness.js
@@ -204,14 +204,19 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
         }
     });
 
-    // Context awareness for the platform (operating system)
+    // Context awareness for the reported platform name (operating system)
+    // Note that navigator.platform (in-browser) and os.platform()
+    // have different return values - this simply registers the
+    // reported value as a string rather than doing any more 
+    // sophisticated detection
 
     fluid.contextAware.getPlatformName = function () {
         // Gets platform name in browser contexts
-        if (navigator) {
+        if (typeof(navigator) !== "undefined" && navigator) {
             return navigator.platform ? navigator.platform : undefined;
-        // Gets platform name in node contexts    
-        } else if (os.platform) {
+        // Gets platform name in node contexts
+        } else {
+            var os = require('os');
             return os.platform();
         }
     };
--- a/src/framework/enhancement/js/ContextAwareness.js
+++ b/src/framework/enhancement/js/ContextAwareness.js
@@ -206,16 +206,15 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
 
     // Context awareness for the reported browser platform name (operating system)
 
-    fluid.contextAware.getBrowserPlatformName = function () {
-        // Gets platform name in browser contexts
-        if (typeof(navigator) !== "undefined" && navigator) {
-            return navigator.platform ? navigator.platform : undefined;
-        }
+    fluid.registerNamespace("fluid.contextAware.browser");
+
+    fluid.contextAware.browser.getPlatformName = function () {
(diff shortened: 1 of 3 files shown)


src/framework/enhancement/js/ContextAwareness.js
```diff
@@ -204,15 +204,18 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
 
-    // Context awareness for the platform (operating system)
+    // Context awareness for the reported browser platform name (operating system)
 
-    fluid.contextAware.getPlatformName = function () {
-        return navigator.platform ? navigator.platform : undefined;
+    fluid.contextAware.getBrowserPlatformName = function () {
+        // Gets platform name in browser contexts
+        if (typeof(navigator) !== "undefined" && navigator) {
+            return navigator.platform ? navigator.platform : undefined;
```
jobara: Is it necessary to do the conditional check on `navigator.platform`? Can you just do `return navigator.platform` or could it have a value that is falsey and not undefined that you want to catch?

src/framework/enhancement/js/ContextAwareness.js
```diff
@@ -204,15 +204,18 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
...
+        }
     };
 
     fluid.contextAware.makeChecks({
-        "fluid.platformName": {
-            funcName: "fluid.contextAware.getPlatformName"
+        "fluid.browserPlatformName": {
+            funcName: "fluid.contextAware.getBrowserPlatformName"
```
jobara: Can we make use of the `fluid.browser` namespace instead? So you'd end up with with `fluid.browser.platformName` and `fluid.contextAware.browser.getPlatformName` and similar throughout.

tests/test-core/utils/js/ConditionalTestUtils.js
```diff
@@ -83,43 +83,30 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-        "fluid.platform": "fluid.test.conditionalTestUtils.getPlatformName"
-    });
-
-    fluid.contextAware.makeChecks({
-        "fluid.platform.isLinux": "fluid.test.conditionalTestUtils.isLinux",
-        "fluid.platform.isMac": "fluid.test.conditionalTestUtils.isMac",
-        "fluid.platform.isWindows": "fluid.test.conditionalTestUtils.isWindows"
+        "fluid.platform.isBrowserLinux": "fluid.test.conditionalTestUtils.isBrowserLinux",
+        "fluid.platform.isBrowserMac": "fluid.test.conditionalTestUtils.isBrowserMac",
+        "fluid.platform.isBrowserWindows": "fluid.test.conditionalTestUtils.isBrowserWindows"
```
jobara: Similar to above, can we use `fluid.browser` namespace instead. e.g. `fluid.browser.platform.isLinux`

src/framework/enhancement/js/ContextAwareness.js
amb26: +1

src/framework/enhancement/js/ContextAwareness.js
amb26: Idiomatically this can just be 
return typeof(navigator) !== "undefined" && navigator.platform ? navigator.platform : undefined;

waharnum:
@jobara I've made the suggested changes to namespace usage & the more idiomatic check for the navigator global, this can be review again for merge.

tests/test-core/utils/js/ConditionalTestUtils.js
```diff
@@ -83,43 +83,30 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 
-    fluid.test.conditionalTestUtils.isMac = function () {
-        return fluid.test.conditionalTestUtils.contextValueContains("Mac", "{fluid.platform}");
+    fluid.test.conditionalTestUtils.isBrowserMac = function () {
+        return fluid.test.conditionalTestUtils.contextValueContains("Mac", "{fluid.browser.platformName}");
     };
 
-    fluid.test.conditionalTestUtils.isWindows = function () {
-        return fluid.test.conditionalTestUtils.contextValueContains("Windows", "{fluid.platform}");
+    fluid.test.conditionalTestUtils.isBrowserWindows = function () {
```
jobara: @waharnum i'm struggling a bit with this in my head, but I think that it might be better to call these something like "isBrowserOnWindows". I think that's more specific, and in this particular case less ambiguous as it might be confused with the browsers window or window object. What do you think?

waharnum: Specific and reduced ambiguity > variable name length, always.

jobara:
Merged at 68351aadf504e6aee4827468934a1f2e341ffb9c

PR #767, opened 2016-10-20
FLUID-5986: Integration of basic-node-tests.js with Testem runner

waharnum:
* Modifies the `testem.json` configuration to add a custom "Node Module Basic Packaging Tests" launcher to run `basic-node-tests.js` as part of the overall test run
* Modifies `basic-node-tests.js` to accept a command-line flag of `--tap`; if supplied, it will output a TAP-format report (see https://github.com/testem/testem#processes-with-tap-output) that can be consumed and reported by Testem
* Modifies `basic-node-tests.js` to exit with a non-zero exit code if any tests fail, to better support usage in CI

The node packaging tests will now be run as part of the `grunt tests` tasks in the VM, or locally with `testem ci --file tests/testem.json`

You can run the node packaging tests as part of Testem separately with this command: `testem ci --file tests/testem.json --launch "Node Module Basic Packaging Tests"` (assuming you have Testem installed)

### DIFF
--- a/tests/node-tests/basic-node-tests.js
+++ b/tests/node-tests/basic-node-tests.js
@@ -15,32 +15,10 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 var fluid = require("../../src/module/fluid.js"),
     path = require("path");
 
-// For accumulating TAP messages for parsing by testem custom launcher
-var tapOutput = "";
 // Number of expected assertions
 var expectedAsserts = 20;
 // Number of expected test cases
 var expectedTestCases = 10;
-
-// Set boolean for the  presence of the --tap command line option to output a
-// TAP report for consumption by testem when used as a custom launcher
-var shouldOutputTAP = process.argv.findIndex(function (argument) {
-    return argument.indexOf("--tap") > -1;
-}) > -1;
-
-// Function to output a TAP  failure report
-// Assumed to be used as below with fluid.onUncaughtException.addListener
-var outputTAPFailure = function (message) {
-    if (!fluid.tests.expectFailure) {
-        console.log("not ok -  " + message);
-    }
-};
-
-// Register listener to output TAP failure report
-if (shouldOutputTAP) {
-    fluid.onUncaughtException.addListener(outputTAPFailure, "outputTAPFailure",
-        fluid.handlerPriorities.uncaughtException.log);
-}
 
 fluid.loadTestingSupport();
 
@@ -48,6 +26,8 @@ var QUnit = fluid.registerNamespace("QUnit");
 var jqUnit = fluid.registerNamespace("jqUnit");
 
 fluid.registerNamespace("fluid.tests");
+
+fluid.registerNamespace("fluid.tests.tapOutput");
 
 fluid.loadInContext("../../tests/test-core/testTests/js/IoCTestingTests.js");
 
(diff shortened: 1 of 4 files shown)


waharnum:
Tagging @jobara / @colinbdclark / @amb26, and @avtar because this modifies the CI configuration.

tests/node-tests/basic-node-tests.js
```diff
@@ -39,8 +39,23 @@ fluid.require("test-module", require, "test-module");
...
+}) > -1;
+
+// For accumulating TAP messages for parsing by testem custom launcher
+var tapOutput = "";
+
 QUnit.testDone(function (data) {
-    fluid.log("Test concluded - " + data.name + ": " + data.passed + " passed, " + data.failed + " failed");
+    var testMessage = "Test concluded - " + data.name + ": " + data.passed + " passed, " + data.failed + " failed";
+    fluid.log(testMessage);
+    if (shouldOutputTAP) {
```
colinbdclark: This is pretty interesting stuff. I know others have been interested in outputting TAP-formatted results from node-jqUnit. Should we consider moving this upstream to there? @amb26 or @the-t-in-rtf, thoughts?

the-t-in-rtf: Yes, please.  I tried briefly integrating with a third-party plugin but was stymied by our older forked version of QUnit.  I have an approach to get TAP output from QUnit browser tests, I would love to see TAP support available for node jqUnit tests as well.

waharnum: @the-t-in-rtf, I looked at this briefly after talking to @colinbdclark & we observed that `node-jqUnit` is actually not used in running these tests currently. So I think we could consider moving this feature upstream there, but it wouldn't address the immediate issue here of wanting to run the node packaging tests as part of the Infusion CI process.

amb26: I think this is ok. We have long-standing modularisation issues with respect to node-jqUnit, and clearly a lot of this code belongs there - but this probably isn't the time to deal with them - we should wait until we can get dug in to do a full "monorepo" style reorganisation of the Infusion repository and then fold node-jqUnit into this structure as a further "internal module". 

I think the factoring could be improved slightly to stick the TAP output into a separate callback applied to QUnit.testDone, but in general I think it's ok to have such stuff duplicated between basic-node-tests and node-jqUnit until we can have a more thorough refactoring. Please could you add a comment here referencing https://issues.fluidproject.org/browse/FLUID-5744 as a pointer to this work.

waharnum: @amb26 agree that factoring makes the code cleaner & have implemented - the TAP generation is now in separate `QUnit.testDone` / `QUnit.done` callbacks.

I've linked FLUID-5744 and FLUID-5986 as related issues, and added a comment to FLUID-5744 pointing to this PR discussion.

amb26:
Cheers @waharnum - I tested this out and it mostly works fine. However, introducing a top-level error into  basic-node-tests.js (for example, "string".fail() ) outside any function just causes the tests to be silently skipped with no failure reported via TAP. It doesn't seem to matter whether this is done either before or after the QUnit.load() directive at the end of the file (which should probably be moved up higher).

This is a quite tricky case but we should probably do at least something about it. In jqUnit-node I use the following approach - https://github.com/fluid-project/node-jqunit/blob/master/lib/jqUnit-node.js#L184 - but another possibility is the use of an "uncaught exception handler".

waharnum:
@amb26 I don't think https://github.com/fluid-project/infusion/pull/767/commits/11395cc4df21b076a110bda74d80b58f7ff34071 is appropriate final form, but it achieves the desired outcome. How does `fluid.onUnCaughtException` work? Can I use this to register a similar "TAP failure report" listener to what I've done with the `try... catch` wrapper approach?

amb26:
@waharnum - yes, what you've done so far doesn't represent an uncaught exception handler, but just a standard exception block. Yes, you can and should use fluid.onUncaughtException. It's just a standard Fluid event firer to which you can add a listener which produces a TAP failure report. I'm not sure what you were trying to achieve with the "expectedTestCases" variable. In general if there is an uncaught exception it will occur outside any fixture, so this kind of diagnostic won't be helpful.

waharnum:
@amb26 regarding `expectedTestCases` TAP expects to report on the number of test cases run, which is different from the number of assertions run (what the `expected` variable that I renamed `expectedAssertions` represents) - I think we should probably track both. This also lets us add a proper TAP header, which is a further diagnostic.

I'll rewrite using `fluid.onUncaughtException`.

amb26:
Thanks, @waharnum - I understand the strange expectedTestCases loop a little better now. I assume that TAP throws a fit if there isn't output for each test?
See if you can get all the tap-related functions to group together in the file, preferably in their own namespace. Extra credit if you can manage to factor them off into their own file :)

waharnum:
@amb26 TAP is actually fine with accumulating the number of tests itself automatically - it's just an extra guard about making sure the right number of tests run (similar to `expected`).

I've got things working with `fluid.onUncaughtException`; let me see about factoring them out further if I can do so quickly.

waharnum:
@amb26 aside from the points of integration within the main tests, I split all the TAP functionality out into a separate file.

amb26:
Cheers @waharnum - the new structure looks great! Unfortunately the newly added material fails linting

waharnum:
@amb26 sorry about that, linting issues fixed now

amb26:
@waharnum - thanks so much, this is some really helpful work to shore up our infrastructure!

PR #768, opened 2016-10-24
FLUID-5987: consolidated events into a single handler

jobara:
There had been two separate handlers bound to the same event. Combined all of the
pieces into a single event handler to ensure that they opperate in the correct order.

https://issues.fluidproject.org/browse/FLUID-5987


amb26:
This was better expressed the way it was before. Instead of fusing code into longer functions, please use the listener priority system to arbitrate order between notifying listeners: http://docs.fluidproject.org/infusion/development/Priorities.html

jobara:
@amb26 and @cindyli I've updated the PR based on the suggestion, also added in some additional namespaces for listeners.

cindyli:
Shall we take the chance to declaratively define these listeners?

amb26:
@cindyli - nice idea - do you think we have time to fit it in before the release? Whilst the ginger world itself hasn't improved since the warning comment at the head of this function was written, I think we could now do this quite easily with a distributeOptions block and sticking all these definitions into an "unbound grade"

cindyli:
It's decided to address above comments after the infusion 2.0 release. The JIRA ticket to track this issue: https://issues.fluidproject.org/browse/FLUID-5988

cindyli:
Merged at a74754234ccf34e43465063bbbdf8971bb4075b0

PR #771, opened 2016-10-28
FLUID-6011: Using IDs instead of empty links for ToC

jobara:
The ToC now points at the ID of a header item instead of creating an empty anchor to point at. 

https://issues.fluidproject.org/browse/FLUID-6011

### DIFF
--- a/src/components/tableOfContents/js/TableOfContents.js
+++ b/src/components/tableOfContents/js/TableOfContents.js
@@ -20,25 +20,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     *******/
     fluid.registerNamespace("fluid.tableOfContents");
 
-
-    fluid.tableOfContents.insertAnchor = function (name, element, anchorClass) {
-       // In order to resolve FLUID-4453, we need to make sure that the owner document is correctly
-       // taken from the target element (the preview may be in an iframe)
-        var anchor = $("<a></a>", element.ownerDocument);
-        anchor.prop({
-            "class": anchorClass,
-            name: name,
-            id: name
-        });
-        anchor.insertBefore(element);
-    };
-
-    fluid.tableOfContents.headingTextToAnchorInfo = function (heading, guidFunc) {
-        var guid = guidFunc();
+    fluid.tableOfContents.headingTextToAnchorInfo = function (heading) {
+        var id = fluid.allocateSimpleId(heading);
 
         var anchorInfo = {
-            id: guid,
-            url: "#" + guid
+            id: id,
+            url: "#" + id
         };
 
         return anchorInfo;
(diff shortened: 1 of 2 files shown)


src/components/tableOfContents/js/TableOfContents.js
```diff
@@ -20,25 +20,12 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
     *******/
     fluid.registerNamespace("fluid.tableOfContents");
 
-
-    fluid.tableOfContents.insertAnchor = function (name, element, anchorClass) {
-       // In order to resolve FLUID-4453, we need to make sure that the owner document is correctly
-       // taken from the target element (the preview may be in an iframe)
-        var anchor = $("<a></a>", element.ownerDocument);
```
simonbates: Is this comment above regarding iframe something that we need to be concerned with? I'm thinking not, as we are adding an id attribute to an existing element rather than adding a new element but wanted to check with you.

tests/component-tests/tableOfContents/js/TableOfContentsTests.js
```diff
@@ -276,7 +276,7 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
             // To address IE7 problem, http://bugs.jquery.com/ticket/7117
             // To fix, strip it URI if the windows.location is in href. Otherwise, do nothing.
             var eleHref = elm.attr("href").replace($(location).attr("href"), "");
-            jqUnit.assertTrue("ToC anchor set correctly", hInfo.url ? eleHref === hInfo.url : eleHref.indexOf(fluid.fluidInstance) === 1);
+            jqUnit.assertTrue("ToC anchor set correctly", hInfo.url ? eleHref === hInfo.url : eleHref.indexOf("fluid-id-" + fluid.fluidInstance) === 1);
```
simonbates: Am I correct in reading this as: if a header had an existing id, verify that the table of contents entry href is using it, otherwise verify that the table of contents entry href has the right form (starts with "fluid-id-" + fluid.fluidInstance) ?

It looks to me like we aren't verifying the generated ids for headers that didn't already have one. Is this right?

src/components/tableOfContents/js/TableOfContents.js
jobara: @simonbates I believe it was only an issue because of the injection, but you could test it out by running the full page manual-tests for this branch ( e.g http://build.fluidproject.org/infusion/tests/manual-tests/framework/preferences/fullPage-schema/index-nativeHTML.html ). 

Also, looking at the [FLUID-4453](https://issues.fluidproject.org/browse/FLUID-4453), and we've also mostly removed the jQuery requirements in the iframes as well.

tests/component-tests/tableOfContents/js/TableOfContentsTests.js
```diff
@@ -260,7 +260,7 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
         return set;
     };
 
-    var renderTOCTest = function (that, testHeadings) {
+    var renderTOCTest = function (that, testHeadings, anchorInfo) {
```
simonbates: Please rename the "that" parameter to "levels".

jobara:
@simonbates ready for more review.

simonbates:
Merged at 131466ee67323db4c01442f38a2b76363ac24ab4

PR #773, opened 2016-11-02
FLUID-6027: Switching to fl-hidden-accessible from fl-uploader-hidden

jobara:
Switched from fl-uploader-hidden to fl-hidden-accessible so that the label of the
button would not be removed from the accessibility tree

https://issues.fluidproject.org/browse/FLUID-6027


cindyli:
Merged at 0c483584d4ee60f0596aa18350de0a119b6a48bc

PR #774, opened 2016-11-02
FLUID-6025: Removing delete instructions from the row labels

jobara:
The aria-label had include the remove information that was only relevant before
the file was uploaded. This has been removed. The state of the file row is still
updated in the title attribute, which contains the removal instructions before uploading
and says "file uploaded" after.

https://issues.fluidproject.org/browse/FLUID-6025


cindyli:
Merged at b7d67b0211a919939cdbc1adaf5d9bb6c315147f

PR #775, opened 2016-11-03
FLUID-6017 + FLUID-6018: style native HTML sliders with CSS, make them responsive to prefs framework enactors

waharnum:
This addresses both https://issues.fluidproject.org/browse/FLUID-6017 and https://issues.fluidproject.org/browse/FLUID-6018 by styling the native sliders with CSS, using the techniques described at https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/

Additionally, since it's rather painful to run `grunt buildStylus` each time when doing Stylus development, I've added a `grunt watch:buildStylus` task using `grunt-contrib-watch` that automatically recompiles Stylus -> CSS as changes are made. I ended up doing this as a local customization at first and then concluded it might be generally useful for the project.

Needs to be tested with IE and Edge, will do so when in the office later this morning. Opening PR now for review of the basic approach.

Tagging @jobara

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,8 +181,8 @@ CSS files for the Preferences Framework have been re-written in Stylus. Only Sty
 
 For developing the Preferences Framework, run the following from the project root to compile Stylus files to CSS:
 
-    `grunt buildStylus`
+    grunt buildStylus
 
 A `watch` task using [grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch) is also supplied to ease Stylus development. This task launches a process that watches all Stylus files in the `src` directory and recompiles them when they are changed. This task can be run using the following command:
 
-    `grunt watch:buildStylus`
+    grunt watch:buildStylus
--- a/README.md
+++ b/README.md
@@ -178,6 +178,11 @@ All of these libraries are already bundled within the Infusion image.
 
 Infusion is in the process of switching to use [Stylus](http://learnboost.github.io/stylus/) for CSS pre-processing.
 CSS files for the Preferences Framework have been re-written in Stylus. Only Stylus files are pushed into the github repository.
+
 For developing the Preferences Framework, run the following from the project root to compile Stylus files to CSS:
 
-    grunt buildStylus
+    `grunt buildStylus`
+
+A `watch` task using [grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch) is also supplied to ease Stylus development. This task launches a process that watches all Stylus files in the `src` directory and recompiles them when they are changed. This task can be run using the following command:
+
+    `grunt watch:buildStylus`
--- a/src/framework/preferences/css/stylus/PrefsEditor.styl
+++ b/src/framework/preferences/css/stylus/PrefsEditor.styl
@@ -124,7 +124,7 @@
             left:0;
             margin-top: -0.4em;
             margin-left: -0.4em;
-            slider-thumb-common()
+            slider-thumb-common();
 
             &:active {
                 outline: none;
@@ -139,15 +139,15 @@
         // Native slider thumb appearance
         &::-webkit-slider-thumb {
             -webkit-appearance: none;
-            slider-thumb-common()
+            slider-thumb-common();
         }
 
         &::-moz-range-thumb {
-            slider-thumb-common()
+            slider-thumb-common();
         }
 
         &::-ms-thumb {
-            slider-thumb-common()
+            slider-thumb-common();
         }
         // Native slider track appearance
         // We hide any inherent track styling and let the styles
@@ -155,15 +155,15 @@
         // from native widget styling
 
         &::-webkit-slider-runnable-track {
-            slider-native-track-common()
+            slider-native-track-common();
         }
 
         &::-moz-range-track {
-            slider-native-track-common()
+            slider-native-track-common();
         }
 
         &::-ms-track {
-            slider-native-track-common()
+            slider-native-track-common();
             color: transparent;
         }
 
--- a/src/framework/preferences/css/stylus/PrefsEditor.styl
+++ b/src/framework/preferences/css/stylus/PrefsEditor.styl
@@ -152,7 +152,7 @@
         slider-native-track-common()
     }
 
-    .fl-prefsEditor input[type=range].fl-slider::-ms-track {
+    input[type=range].fl-slider::-ms-track {
         slider-native-track-common()
     }
 
--- a/src/framework/preferences/css/stylus/PrefsEditor.styl
+++ b/src/framework/preferences/css/stylus/PrefsEditor.styl
@@ -9,6 +9,21 @@
          url('../fonts/InfusionIcons-PrefsEditor.eot');
 }
 
+// Mixin for common styles of the "thumb" (slider control)
+// used by both the jQuery and native HTML sliders
+slider-thumb-common()
+    height: 1.5em;
+    width: 1.5em;
+    background-color: #fff;
+    border: 1px solid #b3b3b3;
+    border-radius: 2em;
+    box-shadow: 4px 2px 3px rgba(0,0,0,0.3);
+    background-image: -webkit-linear-gradient(right top, #fff 46%, #e9eaea 73%);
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -178,6 +178,11 @@ All of these libraries are already bundled within the Infusion image.
 
 Infusion is in the process of switching to use [Stylus](http://learnboost.github.io/stylus/) for CSS pre-processing.
 CSS files for the Preferences Framework have been re-written in Stylus. Only Stylus files are pushed into the github repository.
+
 For developing the Preferences Framework, run the following from the project root to compile Stylus files to CSS:
 
-    grunt buildStylus
+    `grunt buildStylus`
```
jobara: The indentation had already put this in a code block. Perhaps you want to change it to the explicit code block with  "```"

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -108,21 +108,54 @@
...
+        slider-thumb-common()
+    }
+
+    input[type=range].fl-slider::-moz-range-thumb {
+        slider-thumb-common()
+    }
+
+    .fl-prefsEditor input[type=range].fl-slider::-ms-thumb {
+        slider-thumb-common()
+    }
```
jobara: I'm having a bit of trouble reading the indentation, but I think the `.fl-prefsEditor` portion is unnecessary and maybe incorrect in this position as it's already grouped from line 12. You could probably also adjust the use the [parent reference](http://stylus-lang.com/docs/selectors.html#parent-reference) to move up the `input[type=range].fl-slider` declarations. And you may be able to simplify the vendor specific thumb styles in a similar fashion.

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -108,21 +108,54 @@
...
+
+    input[type=range].fl-slider::-moz-range-track {
+        slider-native-track-common()
+    }
+
+    .fl-prefsEditor input[type=range].fl-slider::-ms-track {
+        slider-native-track-common()
+    }
+
```
jobara: Similar to previous comment

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -24,11 +24,17 @@ themes = {
     }
 }
 
+slider-thumb-common-contrastTheme()
+    background: none;
+    background-color: fColor;
```
jobara: I'm assuming this is defining the stylus function. We still use the {} to keep things readable.

src/framework/preferences/css/stylus/PrefsEditor.styl
waharnum: Good tip on the parent references, thanks.

waharnum: Parent references can make it a little more concise, but each vendor-specific pseudoclass still needs to be a separate declaration, or the browser CSS parsers will choke on the unrecognized pseudoclasses.

jobara: @waharnum could you write it like?

```
&::-webkit-slider-thumb,
&::-moz-range-thumb,
&::ms-thumb {
    slider-thumb-common()
}
```

waharnum: That causes Stylus to produce the following: 

```
.fl-prefsEditor input[type=range].fl-slider::-webkit-slider-thumb,
.fl-prefsEditor input[type=range].fl-slider::-moz-range-thumb,
.fl-prefsEditor input[type=range].fl-slider::-mms-thumb {
  /* ...css directives /*
}
```

It's valid CSS, but some browsers will throw out the entire directive because of a vendor-specific pseudoclass they don't recognize. The issue described at http://stackoverflow.com/questions/16982449/why-isnt-it-possible-to-combine-vendor-specific-pseudo-elements-classes-into-on, basically.

waharnum:
@jobara, this is ready for review again - the one outstanding graphical issue is not a blocker (cropping of thumb in IE11 / Edge), but has been documented at https://issues.fluidproject.org/browse/FLUID-6030

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -122,21 +122,65 @@
...
-            height: 1.5em;
-            width: 1.5em;
             margin-top: -0.4em;
             margin-left: -0.4em;
-            background-color: #fff;
-            border: 1px solid #b3b3b3;
-            border-radius: 2em;
-            box-shadow: 4px 2px 3px rgba(0,0,0,0.3);
-            background-image: linear-gradient(right top, rgb(255,255,255) 46%, rgb(233,234,234) 73%);
+            slider-thumb-common()
```
waharnum: Will do - I'm uncertain how stable / active it is, but is it worth considering the use of [grunt-stylint](https://github.com/xdissent/grunt-stylint) in the future?

jobara: @waharnum definitely something to think about. We should also have something similar for our css and html files too.

jobara:
@jhung could you please review the stylus changes in this PR

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -122,21 +122,65 @@
...
+            slider-thumb-common();
 
             &:active {
                 outline: none;
             }
         }
     }
+
+    // Native slider appearance
+
```
jhung: Recommend putting a link to the CSS-Tricks guide so we can revisit that article for information / updates as time progresses.

src/framework/preferences/css/stylus/utils/Helpers.styl
```diff
@@ -21,3 +21,23 @@ linear-gradient() {
...
+
+// Mixin for common styles of the "thumb" (slider control)
+// used by both the jQuery and native HTML sliders
+slider-thumb-common() {
+    height: 1.5em;
+    width: 1.5em;
+    background-color: #fff;
+    border: 1px solid #b3b3b3;
+    border-radius: 2em;
+    box-shadow: 4px 2px 3px rgba(0,0,0,0.3);
```
jhung: Is it possible to use `em` instead of `px`?

jhung: Edit: I see now that this styling was written elsewhere and moved to this function. I guess if it's been working we should leave it. We should make sure this gets fixed in the redesign.

waharnum: @jhung yes, this is just a straight migration of the existing styles for the jQuery-based sliders, with the common characteristics between styling native & jQuery pulled into the helper

I assume we'd want the border and box-shadow to scale via `em`?

jhung: Yes - it would be ideal if this scaled with the font-size, but I don't see it as something that would block this from getting merged.

Ideally we'd use `rem` instead of `em` even - but that's another task all together.

waharnum: @jhung along the lines of adaptive components, I wonder as well about whether or not the `box-shadow` shouldn't be removed entirely from the thumb in the various UIO contrast themes. Does it potentially interfere the accessibility of the contrast theme?

It's quite visible in some contrast themes and nearly invisible in others - I see looking at `Themes.styl` that we remove `box-shadow` for some of the other components. Two screenshots for comparison (the black-on-white and yellow-on-black themes). 

image

image

jobara: @waharnum, I agree that it would be better to remove the shadows.

waharnum: @jobara done

waharnum:
@jobara I think with the above two changes (links to css-tricks as suggested by @jhung, and the removal of the `box-shadow` in the contrast themes generation) this is good to go in unless there are further comments.

jobara:
Merged at 783779e872feb3bfe7110529a189de361de54f8c

PR #779, opened 2016-11-04
FLUID-6032: Fixing progress label

jobara:
- Added the progress label text update back to the demo
- Updated the progress component to optionally interpolate the percentage into a string template

https://issues.fluidproject.org/browse/FLUID-6032

### DIFF
--- a/demos/progress/js/progressDemo.js
+++ b/demos/progress/js/progressDemo.js
@@ -84,7 +84,7 @@ var demo = demo || {};
         strings: {
             confirmStatus: "Confirm and submit your order.",
             orderSubmitted: "Order Submitted. Demo finished.",
-            percentCompleted: "%percent% Complete",
+            percentCompleted: "%percentComplete% Complete",
             progressTitle: "Checking inventory, please wait."
         },
         events: {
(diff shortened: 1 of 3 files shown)


demos/progress/js/progressDemo.js
```diff
@@ -84,7 +84,7 @@ var demo = demo || {};
         strings: {
             confirmStatus: "Confirm and submit your order.",
             orderSubmitted: "Order Submitted. Demo finished.",
-            percentCompleted: "%percent% Complete",
+            percentCompleted: "%percentComplete% Complete",
```
cindyli: In the string text, might be less error-pron to use `%percentCompleted%` to match up with the key.

src/components/progress/js/Progress.js
```diff
@@ -132,8 +132,9 @@ var fluid_2_0_0 = fluid_2_0_0 || {};
             fluid.progress.updateWidth(that, pixels);
         }
 
-        if (labelText !== null) {
-            fluid.progress.updateText(that.label, labelText);
+        if (fluid.isValue(labelText)) {
+            var text = fluid.stringTemplate(labelText, {percentComplete: percent});
+            fluid.progress.updateText(that.label, text);
```
cindyli: Each update in this `updateProgress` function, such as `updateWidth`, `updateText`, `updateARIA` should be a namespaced model listener. But it's probably ok for now since bringing component code up to date with new infusion features is worth a separate ticket.

demos/progress/js/progressDemo.js
cindyli: `%percentComplete%` as a string template is also used in progress.js. So let's keep it this way.

cindyli:
Merged at 0afec08bca1ecd5455645f950077d46e3b7efa91

PR #782, opened 2016-11-09
FLUID-6053: reduce fixed width of sliders slightly to deal with issue…

waharnum:
This shaves 0.2em off the fixed width of the sliders so that the textfield doesn't linebreak when using the `Enlarge buttons, menus, text-fields, and other inputs` UIO option.

@jobara, the issue we found with the jQuery slider not respecting the "enlarge inputs" option has been added to an existing JIRA at https://issues.fluidproject.org/browse/FLUID-4606


jobara:
merged at 8fbd9a1ef3ee94b2dea8d9384348ac92b5a103d0

PR #789, opened 2016-11-22
FLUID-5943: Minimal Footprint Prefs Framework Example

jobara:
Example code for https://github.com/fluid-project/infusion-docs/pull/95

https://issues.fluidproject.org/browse/FLUID-5943

### DIFF
--- a/examples/framework/preferences/minimalFootprint/fullPage.html
+++ b/examples/framework/preferences/minimalFootprint/fullPage.html
@@ -1,0 +1,131 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+
+        <link rel="stylesheet" type="text/css" href="../../../../src/lib/normalize/css/normalize.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/framework/core/css/fluid.css" />
+
+        <link rel="stylesheet" type="text/css" href="../../../../src/lib/jquery/ui/css/fl-theme-by/by.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/lib/jquery/ui/css/fl-theme-yb/yb.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/lib/jquery/ui/css/fl-theme-bw/bw.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/lib/jquery/ui/css/fl-theme-wb/wb.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/lib/jquery/ui/css/fl-theme-lgdg/lgdg.css" />
+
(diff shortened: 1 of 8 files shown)


examples/framework/preferences/minimalFootprint/js/loadScripts.js
```diff
@@ -0,0 +1,66 @@
...
+https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
+*/
+
+var fluid_2_0_0 = fluid_2_0_0 || {};
+var fluid = fluid || fluid_2_0_0;
+
+(function (fluid) {
+    "use strict";
+
+    fluid.load = {};
```
amb26: Should read fluid.load = fluid.load || {} since there is the barest chance that some day such material will end up in core and then loaded in the wrong order

examples/framework/preferences/minimalFootprint/index.html
```diff
@@ -0,0 +1,72 @@
...
+        <div class="demo-goTo-link">
+            <a href="fullPage.html">Go To Display Preferences</a>
+        </div>
+        <main class="fl-centered">
+            <div class="flc-toc-tocContainer toc"> </div>
+
+            <script>
+
+                // Will load the array of scripts if the "fluid-ui-settings" cookie is found
+                fluid.load.lazyLoadScripts("fluid-ui-settings", [
```
amb26: Could you add a comment here explaining how/why this code differs from the one in the docs - that is, that it is written against the broken-out dev version of the files. 
What would be very helpful is if the comment also included the exact path that the relevant build file would be generated at if the user were to run the "grunt buildDists" task - that is, the path they will find this at if they checked out Infusion as an npm module

examples/framework/preferences/minimalFootprint/fullPage.html
```diff
@@ -0,0 +1,131 @@
...
+        <link rel="stylesheet" type="text/css" href="../../../../src/framework/preferences/css/Enactors.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/framework/preferences/css/PrefsEditor.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/framework/preferences/css/FullPrefsEditor.css" />
+        <link rel="stylesheet" type="text/css" href="../../../../src/framework/preferences/css/FullPreviewPrefsEditor.css" />
+
+        <!-- Demo style -->
+        <link rel="stylesheet" type="text/css" href="css/prefsEditorFullPage.css" />
+
+        <title>Preferences Editor: Full Page</title>
+        <script type="text/javascript" src="../../../../src/lib/jquery/core/js/jquery.js"></script>
```
amb26: Similarly a comment here would be very helpful listing the path of the built file that all these includes could be replaced with in the "dist" directory if it were there

jobara:
@amb26 ready for more review.

PR #790, opened 2016-11-24
FLUID-6086: Fixing stdio option to allow stdout

jobara:
The verbose setting now only affects stderr, as the stdout needs to be returned by the function

https://issues.fluidproject.org/browse/FLUID-6086


cindyli:
Merged at 17a345b96f19c11fed4bef5cd45b2be01a5cf03c

PR #794, opened 2016-12-05
FLUID-6091: Compact syntax with empty argument list should supply no arguments

amb26:


tests/framework-tests/core/js/FluidIoCTests.js
```diff
@@ -3578,11 +3584,15 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
         jqUnit.assertEquals("Number", 3.5, that.number);
         jqUnit.assertEquals("true", true, that["true"]);
         jqUnit.assertEquals("false", false, that["false"]);
+        jqUnit.assertEquals("false", false, that["false"]);
```
cindyli: This line is identical with line 3586 and can be removed or enhanced with another use case.

This another test case is to make sure the fix also works well with the compact syntax of `@expand`,
1. add an no args expander into the `members` block at line 3546: 
`noArgsExpander: "@expand:fluid.identity()"`
2. The 3587 line can be modified to assert:
`jqUnit.assertEquals("No arguments", undefined, that.noArgsExpander);`

The execution of the noArgsExpander should return `0` before the fix and now returns `undefined` with the fix.

cindyli:
Merged at c152ca46b197d46fdfe1fca7979d981f9dace78b

PR #797, opened 2017-01-06
FLUID-6090:  Provide a state tracking component

klown:
State tracker component:
https://issues.fluidproject.org/browse/FLUID-6090


src/components/stateTracker/js/StateTracker.js
```diff
@@ -0,0 +1,100 @@
...
+    //                          has changed.  It must have an evaluateChange()
+    //                          function.
+    // @param changeListener  - a listener to handle the onStateChange event.
+    // @param interval        - optional delay between calls to check the state's
+    //                          current value (msec).
+    // @return                - the intervalID.
+    fluid.stateTracker.startTracking = function (that, changeEvaluator, changeListener, interval) {
+        var monitor = that.initMonitorInfo(changeEvaluator);
+        that.events.onStateChange.addListener(changeListener);
+        if (!!interval) {
```
amb26: For every JavaScript value which is truthy, !! of the value is also truthy, and similarly for falsy values, so this double negation isn't necessary - https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/components/stateTracker/js/StateTracker.js
```diff
@@ -0,0 +1,100 @@
...
+            },
+            initMonitorInfo: {
+                funcName: "fluid.stateTracker.initMonitorInfo",
+                args: ["{arguments}.0"]
+                       // changeEvaluator object or function
+            }
+        }
+    });
+
+    // Initiate polling.
```
amb26: Could you write these comments in the JavaDoc-like style which we have standardised on in new work (lots of the core framework still lags), e.g. as seen in https://github.com/GPII/universal/blob/master/gpii/node_modules/gpii-oauth2/gpii-oauth2-authz-server/src/AuthorizationService.js#L182

src/components/stateTracker/js/StateTracker.js
```diff
@@ -0,0 +1,100 @@
...
+        }
+    };
+
+    // Initialize a monitor object for passing to a function that periodically
+    // checks for chanages in the state of a process.
+    // @param changeEvaluator - an object to evaluate if the state has changed
+    //                          that has a function named evaluateChange().
+    //                          Provided by client.
+    // @return                - the monitor object.
+    fluid.stateTracker.initMonitorInfo = function (changeEvaluator) {
```
amb26: A clearer pattern would have the changeEvaluator defined as an invoker on the component itself, with the initial value "fluid.notImplemented" in order to signal that the user needs to override to make the component usable - see https://issues.fluidproject.org/browse/FLUID-5733 for description

src/components/stateTracker/js/StateTracker.js
```diff
@@ -0,0 +1,100 @@
...
+var fluid_2_0_0 = fluid_2_0_0 || {};
+
+(function ($, fluid) {
+    "use strict";
+
+    fluid.registerNamespace("fluid.stateTracker");
+
+    fluid.defaults("fluid.stateTracker", {
+        gradeNames: ["fluid.component"],
+        events: {
```
amb26: The component would be more usable if this field were modelised - as well as widening the contract of the "changeEvaluator" function to simply return any value which is then applied to the model. A change in that value in the opinion of the data binding system would then notify any modelListeners bound to the fie.d

amb26:
Cheers, @klown, this is a great useful start on this interesting implementation. It would be helpful before we carry on to have an inventory taken of all potential users of this feature to ensure that we are meeting their needs - and to write up the results of the inventory on the JIRA. I know that both @simonbates and @colinbdclark might have potential use cases, as well as that originally found in the "Process Reporter". We will need at least 2 distinct use cases occurring in the wild in order for this component to make its way into the core framework.

src/components/stateTracker/js/StateTracker.js
klown: Okay.  I had thought the double negative was a sanctioned Crockford-ism, but I can't recall where it's from at the moment.  Certainly `grunt lint` didn't object ;-)

src/components/stateTracker/js/StateTracker.js
klown: Right-o.

src/components/stateTracker/js/StateTracker.js
klown: I had considered making the evaluator a function even when this code was part of the ProcessReporter.  It was just easier to make it an object that contained whatever other pieces of data or functions the client wanted to stick in there.  The unit tests go so far use a fluid component - see https://github.com/klown/infusion/blob/FLUID-6090/tests/component-tests/stateTracker/js/StateTrackerTests.js#L28

Another way to go is to make the state tracker component a base abstract grade with an evaluateChange invoker defined as `fluid.notImplemented`, as you suggest.  Then, clients would derive a concrete grade and add any data/functionality they want.

Is that what you are suggesting here?  A base abstract grade?

src/components/stateTracker/js/StateTracker.js
klown: All right, I've made it a fluid.modelCompnent.

klown:
ok to test

gtirloni:
ok to test

amb26:
Mothballed since this facility will not be developed further within the GPII but might be of further interest if the requirement arises again

klown:
Mothballing is fine.  This was effectively mothballed anyway.  The last time I looked, the tests were failing because they used setInterval() and/or setTimeout() to simulate changing the state of a checkbox -- time to come up with another approach, but not right now.

Thanks.

PR #798, opened 2017-01-16
FLUID-6111: Test case and fix for faulty consideration that model with numeric property "length" is an array

amb26:


amb26:
@dinukadesilva - would you like to test this out?

cindyli:
Merged at

PR #801, opened 2017-01-17
FLUID-6114: Generate minified versions of JS and CSS in /dist directory in NPM

waharnum:
JIRA: https://issues.fluidproject.org/browse/FLUID-6114

This came up during my work to add Infusion to CDNJS - the `/dist` directory produced when publishing to NPM currently generates only nonminified JS and CSS files.

This adds a set of definitions to the `distributions` task to produce minified versions of each distribution, and modifies the `dist`-type targets of the `stylus` and `copy` tasks to apply the `.min` filename convention to indicate minified files.

Combined with the feedback from the CDNJS maintainers to configure autoupdate via NPM, this will allow newly-published versions to NPM to automatically be updated on CDNJS.


waharnum:
@jobara, is it possible you could review this? It's not urgent but would be a nice-to-have for the next time we publish a version release to NPM.

waharnum:
@jobara as we discussed, I added a `verifyDist` task that (currently) checks that all the expected JS files and their source maps are available; it runs as the last step of `buildDists`, and fails the task if there's any expected files missing.

Interesting read as well here on when `prepublish` scripts actually run (I learned things from it): https://blog.greenkeeper.io/what-is-npm-s-prepublish-and-why-is-it-so-confusing-a948373e6be1#.36z71qioq

Gruntfile.js
```diff
@@ -40,6 +40,24 @@ var getFromExec = function (command, options) {
...
+*/
+var addMinifyToFilename = function (dest, src, extension) {
+    var minifiedExtension = ".min." + extension;
+    // Don't operate on files that already have a .min convention
+    if (src.indexOf(minifiedExtension) > -1) {
+        return dest + src;
+    } else {
+        return dest + src.replace("." + extension, minifiedExtension);
+    }
+};
```
jobara: I wonder if it would be safer and easier to use node's [path](https://nodejs.org/dist/latest-v6.x/docs/api/path.html) module? I don't think you'd need to take an extension as an argument in that case, as you could source it from the filename itself. You could also avoid the replace call at the end but just reconstructing the file name with it's name, "min" and extension.

Gruntfile.js
```diff
@@ -373,11 +421,40 @@ module.exports = function (grunt) {
...
             "distributions" + ( target ? ":" + target : "" ),
-            "cleanForDist"
+            "cleanForDist",
+            "verifyDists"
         ];
         grunt.task.run(tasks);
     });
 
+    grunt.registerTask("verifyDists", "Verifies that the expected /dist/*.js files and their source maps were created", function () {
+        console.log("Verifying that expected distribution files are present in /dist directory");
```
jobara: would be better to make use of [`grunt.log`](http://gruntjs.com/api/grunt.log).

jobara: Similar for below, and you may want to use the various types of logs they have available.

Gruntfile.js
waharnum: Definitely safer - will implement.

waharnum: It looks like `path.extname(path)` "returns the extension of the path, from the last occurrence of the . (period) character to end of string in the last portion of the path".

This doesn't work with the conventions of generating source maps in a structure like ` infusion-all.js.map`, where the expected output from that would be `infusion-all.min.js.map` to match the source file `infusion-all.min.js`. Using `path.extname(path)` results in improperly placing the `.min` before the `.map`.

I'll noodle on this further, as I think this might be resolveable without a renaming function entirely if `allBuildName` was aware of the `compress` setting. Looking further at the logic of the code, it only currently gets called for dealing with the `all.min` build files to rename them to the minified convention, because the others already have `min` in their filename from their distribution keys. However, that's a side effect rather than intentional, which is dangerous.

It might make more sense to assume all distributions should produce both a non-minified and a minified build, and restructure the logic accordingly to remove the need to define minified distributions in the `distributions` object accordingly.

jobara:
@waharnum could you please update the ReleaseNotes.md file as well to mention the CDN in the obtaining Infusion section?

waharnum:
@jobara I was going to make a separate PR for that, but can do that as part of this one.

waharnum:
@jobara per my comments about `path.extname`'s limitations when dealing with multiple period-separate file extensions, I've done the following:

- created a function for extracting 'full' filename extensions, `getFullFilenameExtension`
- used this function in `addMinifyToFilename` instead of requiring the extension to be specified

I've also implemented the "parallel" logic for generating matching minified distributions.

ReleaseNotes.md
```diff
@@ -49,6 +49,8 @@ For a complete list of Fixes and Improvements see the [Version 2.0](https://issu
 * [Fork on GitHub](https://github.com/fluid-project/infusion)
 * [Download a Build](https://github.com/fluid-project/infusion/releases)
 * [NPM](https://www.npmjs.com/package/infusion)
+* [Use from the CDNJS Content Distribution Network](https://cdnjs.com/libraries/infusion)
```
jobara: Could you do some quick clean up here to make the ReleaseNotes match the Readme. That is adjust the ordering and link text for the various ways of getting Infusion.

Gruntfile.js
```diff
@@ -40,6 +40,37 @@ var getFromExec = function (command, options) {
...
 
+/**
+* Returns the "full extension" of a filename containing
+* multiple periods; common in minified/mapped JS file name
+* conventions, such as "infusion-all.min.js.map"
+* @param {String} filename - the filename to parse
+* @returns {String} the extracted extension such as ".min.js.map"
+*/
+
+var getFullFilenameExtension = function (filename) {
```
jobara: seems to be an extra new line between the comment and the implementation.

Gruntfile.js
```diff
@@ -40,6 +40,36 @@ var getFromExec = function (command, options) {
...
+var addMinifyToFilename = function (dest, src) {
+    var fullExtension = getFullFilenameExtension(src);
+    var minifiedExtension = ".min" + fullExtension;
+    // Don't operate on files that already have a .min extension
+    if (fullExtension.indexOf(".min.") > -1) {
+        return dest + src;
+    } else {
+        return dest + src.replace(fullExtension, minifiedExtension);
+    }
+};
```
jobara: What do you think about performing the string operation using arrays instead? See this jsfiddle example I quickly threw together https://jsfiddle.net/p7xdnbbz/.

waharnum: Agree this is more elegant - using this implementation.

Gruntfile.js
```diff
@@ -50,6 +80,44 @@ module.exports = function (grunt) {
         });
     };
 
+    // Base distribution configuration
+    // This should specify only the options for include/exclude;
+    // a parallel "minify" set with options.compress: true
+    // will be generated for the final configuration
+    var baseDistributions =
```
jobara: I prefer to have the configuration as part of the gruntInitConfig as this is where you'd expect to be modifying configuration if you wanted to add or change the distributions. I think the distributions task should be responsible for creating the minified distributions as well as the source ones.

Gruntfile.js
```diff
@@ -240,7 +314,7 @@ module.exports = function (grunt) {
                 files: [{
                     expand: true,
                     src: ["src/**/css/stylus/*.styl"],
-                    ext: ".css",
+                    ext: "<% buildSettings.compress ? print('.min.css') : print('.css') %>",
```
jobara: what does the `print` method do in this case? for example would it be possible to just write `"<% buildSettings.compress ? '.min.css' : '.css' %>"`

Gruntfile.js
waharnum: OK - this is a reasonable preference. I'll dig more into Grunt to figure out how to accomplish this in a more Grunt-like way.

Gruntfile.js
waharnum: From my work before and reading the documentation at https://lodash.com/docs/4.17.4#template, I believe the `print` method is necessary when inside a lodash `evaluate` template delimiter to actually output a result.

Removing the print statements results in the `ext` parameter being blank (no file extensions); it seems necessary to use when we want to output based on evaluating JS in this style.

waharnum:
Since it seemed straightforward, I've added a verification for the CSS files as well. I'll do some refactoring to extract common functions between the two of them shortly.

Gruntfile.js
```diff
@@ -40,6 +40,35 @@ var getFromExec = function (command, options) {
...
+
+/**
+ * Rename function for grunt file tasks for  adding ".min" convention
+ * to filename string; won't do anything to strings that already
+ * include ".min"
+ * @param {String} dest - supplied by Grunt task, see http://gruntjs.com/configuring-tasks#the-rename-property
+ * @param {String} src - supplied by Grunt task, see http://gruntjs.com/configuring-tasks#the-rename-property
+*/
+var addMinifyToFilename = function (dest, src) {
+    return dest + addMin(src);
```
jobara: Does this work if this is run from Windows? If not, maybe use [path.join](https://nodejs.org/dist/latest-v6.x/docs/api/path.html#path_path_join_paths) instead?

ReleaseNotes.md
```diff
@@ -44,13 +44,15 @@ For a complete list of Fixes and Improvements see the [Version 2.0](https://issu
...
+* [Download a Release](https://github.com/fluid-project/infusion/releases)
+* [Install from NPM](https://www.npmjs.com/package/infusion)
 * [Fork on GitHub](https://github.com/fluid-project/infusion)
-* [Download a Build](https://github.com/fluid-project/infusion/releases)
-* [NPM](https://www.npmjs.com/package/infusion)
+* [Use from the CDNJS Content Distribution Network](https://cdnjs.com/libraries/infusion)
+  * To try out Infusion quickly you can use the following `script` tag to include the full framework from the CDN: `<script src='https://cdnjs.cloudflare.com/ajax/libs/infusion/2.0.0/infusion-all.min.js'></script>`
 
-You can create your own custom build of Infusion using the [grunt build script](README.md#how-do-i-create-an-infusion-package).
+See [How Do I Create an Infusion Package?](#how-do-i-create-an-infusion-package), for details on creating complete or custom packages of Infusion.
```
jobara: I think this link needs to point at the anchor in the README file.

Gruntfile.js
waharnum: Tested on Windows 10 and works (dest & src are supplied by the Grunt task, so presumably take advantage of Grunt's cross-platform capability).

jobara:
merged at 5764b3e8c7e71ca4dfa3f1c850bc1737ba36424a

PR #804, opened 2017-01-23
FLUID-6042

KhushilMistry:
Now aria-checked is true at the starting for every image.

### DIFF
--- a/demos/keyboard-a11y/js/five-star.js
+++ b/demos/keyboard-a11y/js/five-star.js
@@ -45,6 +45,7 @@ var demo = demo || {};
             "role": "radio",
             "aria-checked": false
         });
+        demo.fiveStar.updateARIA(stars,1);
     };
 
     /**


KhushilMistry:
Just added call to demo.fiveStar.updateARIA in the function demo.fiveStar.setARIA which will set aria-checked true for first start.
screen shot 2017-01-23 at 7 36 48 pm

demos/keyboard-a11y/js/five-star.js
```diff
@@ -45,6 +45,7 @@ var demo = demo || {};
             "role": "radio",
             "aria-checked": false
         });
+        demo.fiveStar.updateARIA(stars,1);
```
jobara: What if the model is set to something other than 1? It's probably better to investigate why the modelListener is not setting the value correctly. Additionally, you'll want to ensure that the `tabindex` is set to 0 for the initial star as well. You may need to look into how the items are selected when setting a different rank.

KhushilMistry: I will look at it and try to solve it.

KhushilMistry:
I have figured out that ModelListeners calls first ,not Listeners by debugging. What should we do to call Listeners first rather than ModelListeners ?

cindyli:
@KhushilMistry, when you define this model listener, you can use ```excludeSource: "init"``` to prevent it from being triggered by the initial model data.

The documentation about `excludeSource` can be found at http://docs.fluidproject.org/infusion/development/ChangeApplierAPI.html

KhushilMistry:
So I am thinking that I should prevent model listener for Rank at the start and add call function in the OnCreate listener after SetARIA function call. Is this right way of coding ? 
Its working. I have done in the local repository.

amb26:
Closed since lack of activity for a year

PR #806, opened 2017-01-31
FLUID-6122: Better guarding and lower priority for some log messages

amb26:


src/framework/core/js/DataBinding.js
```diff
@@ -578,8 +578,10 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                 segs: sourceSegs,
                 transactional: options.transactional
             }, sourceListener);
-            fluid.log(fluid.logLevel.TRACE, "Adding relay listener with listenerId " + spec.listenerId + " to source applier with id " +
-                sourceApplier.applierId + " from target applier with id " + applierId + " for target component with id " + target.id);
+            if (fluid.passLogLevel(fluid.logLevel.TRACE)) {
```
cindyli: It seems ```fluid.log()``` function already checks the condition of ```fluid.passLogLevel``` if the first argument is a logLevel: https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/Fluid.js#L301

Is it still necessary to add the ```if``` condition around ```fluid.log()```?

Same question applies to other changes.

amb26: Hi there cindyli - these guards are for performance reasons. Given that typically, "TRACE" level logging statements are in highly sensitive regions of the code, we need to be particularly careful about their impact. By JS (and most traditional, "strict" language) semantics, all arguments to a function call are fully evaluated before the function call occurs - this means that, even if the logic inside fluid.log eventually discards the logging statement, the work has already been done of concatenating the string arguments, etc., resulting in garbage pressure on the VM. This "external and internal" guard logic is common in other environments, e.g. https://commons.apache.org/proper/commons-logging/guide.html#Code_Guards

cindyli:
Merged at 6aab53afdc5cf14d8cf40fb0b2df00d54acaf330

PR #807, opened 2017-02-08
FLUID-6125: Updates the numberToString transform to provide a scale option for rounding to decimal places.

jobara:
https://issues.fluidproject.org/browse/FLUID-6125

Related Docs PR: https://github.com/fluid-project/infusion-docs/pull/116

### DIFF
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -829,9 +829,8 @@ var fluid = fluid || fluid_3_0_0;
     /**
      * Derived from AGK's stack overflow answer ( http://stackoverflow.com/a/12830454 )
      *
-     * Rounds the supplied number to at most the number of digits in the scale.
-     * If the scale is invalid (i.e falsey, not a number, negative value), it is treated as 0 and will
-     * round to the integer.
+     * Rounds the supplied number to at most the number of decimal places indicated by the scale, omiting any trailing 0s.
+     * If the scale is invalid (i.e falsey, not a number, negative value), it is treated as 0.
      * If the scale is a floating point number, it is rounded to an integer.
      *
      * @param {number} num - the number to be rounded
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -840,9 +840,10 @@ var fluid = fluid || fluid_3_0_0;
     fluid.roundToDecimal = function (num, scale) {
         // treat invalid scales as 0
         scale = scale && scale >= 0 ? Math.round(scale) : 0;
-        var number = Math.round(num * Math.pow(10, scale)) / Math.pow(10, scale);
+        var scalep = Math.pow(10, scale);
+        var number = Math.round(num * scalep) / scalep;
         if (num - number > 0) {
-            return (number + Math.floor(2 * Math.round((num - number) * Math.pow(10, (scale + 1))) / 10) / Math.pow(10, scale));
+            return (number + Math.floor(2 * Math.round((num - number) * scalep * 10) / 10) / scalep);
         } else {
             return number;
         }
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -827,9 +827,11 @@ var fluid = fluid || fluid_3_0_0;
     };
 
     /**
-     * Derived from AGK's stack overflow answer ( http://stackoverflow.com/a/12830454 )
+     * Derived from Sindre Sorhus's round-to node module ( https://github.com/sindresorhus/round-to ).
+     * License: MIT
      *
      * Rounds the supplied number to at most the number of decimal places indicated by the scale, omiting any trailing 0s.
+     * Numbers are rounded away from 0 (i.e 0.5 -> 1, -0.5 -> -1).
      * If the scale is invalid (i.e falsey, not a number, negative value), it is treated as 0.
      * If the scale is a floating point number, it is rounded to an integer.
      *
@@ -840,13 +842,10 @@ var fluid = fluid || fluid_3_0_0;
     fluid.roundToDecimal = function (num, scale) {
         // treat invalid scales as 0
         scale = scale && scale >= 0 ? Math.round(scale) : 0;
-        var scalep = Math.pow(10, scale);
-        var number = Math.round(num * scalep) / scalep;
-        if (num - number > 0) {
-            return (number + Math.floor(2 * Math.round((num - number) * scalep * 10) / 10) / scalep);
-        } else {
-            return number;
-        }
+
+        // The following is derived from https://github.com/sindresorhus/round-to/blob/v2.0.0/index.js#L17
+        var sign = num >= 0 ? 1 : -1; // manually calculating the sign because Math.sign is not supported in IE
+        return Number(sign * (Math.round(Math.abs(num) + "e" + scale) + "e-" + scale));
     };
 
     /** Calls Object.freeze at each level of containment of the supplied object
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -841,7 +841,7 @@ var fluid = fluid || fluid_3_0_0;
      * @param {Number} num - the number to be rounded
      * @param {Number} scale - the maximum number of decimal places to round to.
      * @param {String} method - (optional) Request a rounding method to use ("round", "ceil", "floor").
-     *                          if nothing or an invalid method is provided, it will default to "round".
+     *                          If nothing or an invalid method is provided, it will default to "round".
      * @return {Number} The num value rounded to the specified number of decimal places.
      */
     fluid.roundToDecimal = function (num, scale, method) {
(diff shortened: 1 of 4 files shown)


src/framework/core/js/Fluid.js
```diff
@@ -826,6 +826,28 @@ var fluid = fluid || fluid_3_0_0;
...
     };
 
+    /**
+     * Derived from AGK's stack overflow answer ( http://stackoverflow.com/a/12830454 )
+     *
+     * Rounds the supplied number to at most the number of decimal places indicated by the scale, omiting any trailing 0s.
+     * If the scale is invalid (i.e falsey, not a number, negative value), it is treated as 0.
+     * If the scale is a floating point number, it is rounded to an integer.
+     *
+     * @param {number} num - the number to be rounded
```
amb26: We tend to use capital letters for these types - "Number", as per the ECMAScript standard - https://www.ecma-international.org/ecma-262/5.1/#sec-4.3.19 - even though, of course, typeof tests have to use the form with a lower case letter.

src/framework/core/js/Fluid.js
```diff
@@ -826,6 +826,28 @@ var fluid = fluid || fluid_3_0_0;
...
+     * @param {number} scale - the maximum number of decimal places to round to.
+     * @return {number} The num value rounded to the specified number of decimal places.
+     */
+    fluid.roundToDecimal = function (num, scale) {
+        // treat invalid scales as 0
+        scale = scale && scale >= 0 ? Math.round(scale) : 0;
+        var number = Math.round(num * Math.pow(10, scale)) / Math.pow(10, scale);
```
amb26: We should tidy up this code a bit by hoisting out the repeated values - I recommend definitions like var scalep = Math.pow(10, scale), and then replace Math.pow(10, scale + 1) with simply 10 * scalep

tests/framework-tests/core/js/FluidJSTests.js
```diff
@@ -230,6 +230,29 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
         });
     });
 
+    fluid.tests.roundToDecimalTests = [
+        {num: 1.555, scale: 5, expected: 1.555},
```
amb26: Please add negative versions of these values (e.g. -1.555) as well as testing a variety of other values as seen in the stackoverflow thread, e.g. 1.005, 1.5, 55.549, 54.9, -55.55, -55.551, -55.59 etc.

tests/framework-tests/core/js/ModelTransformationTests.js
```diff
@@ -1026,18 +1031,65 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
         transformWrap: true,
         transform: {
             type: "fluid.transforms.numberToString",
             inputPath: "floatyHighy"
         },
-        method: "assertValue",
+        expected: fluid.tests.transforms.source.floatInString,
+        expectedInputPaths: [ "floatyHighy" ]
+    }, {
+        message: "numberToString() converts float values - with scale = 2.",
```
amb26: I think we can remove the terminal full stops from all of these messages as being somewhat confusing. Messages are not sentences after all :)

tests/framework-tests/core/js/ModelTransformationTests.js
```diff
@@ -48,7 +48,12 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
         floaty2: -9876.789,
         hundredInString: "100",
         floatInString: "12.52",
-        floaty2InString: "-9876.789"
+        floatInStringScale1: "12.5",
+        floatInStringScale0: "13",
+        floaty2InString: "-9876.789",
```
amb26: I'm worried that you haven't tested your original use case - I would put in as one of the test numbers the expression 0.1 + 0.2, and try to display it to, say 2 decimal places and ensure that it appears as 0.30. For a number on the other side, try 0.1 + 0.7

tests/framework-tests/core/js/FluidJSTests.js
jobara: I'm trying to add more tests, but getting failures ;(

Part of the problem stems from the fact that Math.round will round -0.5 to 0 instead of -1. ( https://www.ecma-international.org/ecma-262/5.1/#sec-15.8.2.15 ). I suppose my tests should take that into account?

Another edge case is that 0.014999999999999999, 0.0150000000000000001 and 0.015 are all interpreted as the same number. Perhaps that's not a worth while case to test. Thoughts?

tests/framework-tests/core/js/ModelTransformationTests.js
jobara: so just to clarify, the algorithm strips out trailing zeros. So 0.1 + 0.2 should still evaluate to 0.3 even with a scale of 2.

jobara:
@amb26 re-implemented the roundToDecimal function. Ready for more review.

src/framework/core/js/Fluid.js
```diff
@@ -826,6 +826,28 @@ var fluid = fluid || fluid_3_0_0;
...
+     * @param {Number} num - the number to be rounded
+     * @param {Number} scale - the maximum number of decimal places to round to.
+     * @return {Number} The num value rounded to the specified number of decimal places.
+     */
+    fluid.roundToDecimal = function (num, scale) {
+        // treat invalid scales as 0
+        scale = scale && scale >= 0 ? Math.round(scale) : 0;
+
+        // The following is derived from https://github.com/sindresorhus/round-to/blob/v2.0.0/index.js#L17
+        var sign = num >= 0 ? 1 : -1; // manually calculating the sign because Math.sign is not supported in IE
```
amb26: Isn't it incredible that this is still not supported even in IE11! What an appalling browser, it will be great when we can ditch support for this.

tests/framework-tests/core/js/FluidJSTests.js
```diff
@@ -230,6 +230,53 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        {num: 55.59, scale: 1, expected: 55.6},
+        {num: -55.59, scale: 1, expected: -55.6},
+        // IEEE 754 binary floating point numbers treat 0.014999999999999999, 0.0150000000000000001 and 0.015 as equivalent.
+        {num: 0.014999999999999999, scale: 2, expected: 0.02},
+        {num: -0.014999999999999999, scale: 2, expected: -0.02},
+        {num: 0.0150000000000000001, scale: 2, expected: 0.02},
+        {num: -0.0150000000000000001, scale: 2, expected: -0.02},
+        {num: 0.015, scale: 2, expected: 0.02},
+        {num: -0.015, scale: 2, expected: -0.02}
+    ];
```
amb26: Cool - do add in the further case that you mentioned in the channel yesterday - 1.005 - 1

src/framework/core/js/Fluid.js
jobara: I was sort of wondering about that myself. I suppose as long as we want to support Win 7 and 8.

tests/framework-tests/core/js/FluidJSTests.js
jobara: I looked into adding a test for this, but it's not working and I'm not sure what to do about it. 

```javascript
{num: 1.005 - 1, scale: 2, expected: 0.01}
```

So, 1.005 - 1 should equal 0.005 however, it's returning 0.004999999999999893 instead. Probably in relation to the floating point precision. With that in mind, if I look at what it should really equate to mathematically I'd expect the algorithm to round up to 0.01. However, if I look at the number that is being calculated, I'd expect it to round down to 0.

amb26: Ok - I think it's worth adding in this test case with a comment explaining that this result is undesirable, but unavoidable given an imprecise FP representation.

src/framework/core/js/Fluid.js
```diff
@@ -826,6 +826,28 @@ var fluid = fluid || fluid_3_0_0;
...
+     * Rounds the supplied number to at most the number of decimal places indicated by the scale, omiting any trailing 0s.
+     * Numbers are rounded away from 0 (i.e 0.5 -> 1, -0.5 -> -1).
+     * If the scale is invalid (i.e falsey, not a number, negative value), it is treated as 0.
+     * If the scale is a floating point number, it is rounded to an integer.
+     *
+     * @param {Number} num - the number to be rounded
+     * @param {Number} scale - the maximum number of decimal places to round to.
+     * @return {Number} The num value rounded to the specified number of decimal places.
+     */
+    fluid.roundToDecimal = function (num, scale) {
```
amb26: I think it's worth preserving all of sindre's original implementation - it doesn't cost much for the extra flexibility - that is, to expose the possibility for round, ceil and floor rounding, and then to also expose these options in the transformSpec of fluid.transforms.numberToString . We can preserve the actual fixtures (num, scale) for our test cases and add in extra "expected" values to illustrate the results of different rounding.

jobara:
@amb26 ready for more review

PR #808, opened 2017-02-09
FLUID-6126: Fix and test case for failure to construct advised child of root component

amb26:


cindyli:
Merged at f830efb6c0d36e505e4372e24cd89939ab79959f

PR #809, opened 2017-02-10
FLUID-6127: Fix and test cases for faulty mapping of "DELETE" changes by wildcard modelListeners

amb26:


tests/framework-tests/core/js/DataBindingTests.js
```diff
@@ -911,6 +911,97 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        model: {
+            idToPath: {
+                first: 1
+            }
+        }
+    });
+
+    fluid.tests.fluid6127fixtures = [{
+        change: {
+            type: "ADD",
```
cindyli: Test cases here cover change type `ADD` and `DELETE`. Shall test cases when type is not specified (the `modify` case) be added?

amb26: There is currently no other separate change type in the system - the default when the type is not specified is "ADD"

cindyli: OK, I see.

cindyli:
Merged at 00f931479c38605f16106e474695e9eb624748e5

PR #812, opened 2017-02-16
FLUID-4713: fileSizeLimit specifies a size files must be smaller than…

aashay201297:
…, rather than the max size

changed the greater than equal to sign to greater only in HTML5UploaderSuppoprt.js

### DIFF
--- a/src/components/uploader/js/HTML5UploaderSupport.js
+++ b/src/components/uploader/js/HTML5UploaderSupport.js
@@ -251,7 +251,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             var file = files[i];
             if (fileLimit && remainingUploadLimit === 0) {
                 that.events.onQueueError.fire(file, fluid.uploader.queueErrorConstants.QUEUE_LIMIT_EXCEEDED);
-            } else if (file.size >= sizeLimit) {
+            } else if (file.size > sizeLimit) {
                 file.filestatus = fluid.uploader.fileStatusConstants.ERROR;
                 that.events.onQueueError.fire(file, fluid.uploader.queueErrorConstants.FILE_EXCEEDS_SIZE_LIMIT);
             } else if (!fileLimit || remainingUploadLimit > 0) {
--- a/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
+++ b/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
@@ -201,7 +201,7 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 
             jqUnit.module("Uploader Basic Tests");
 
-            var file1 = fluid.tests.uploader.html5.mockFile("file1", "emptyfile.zip", "application/zip", 0),
+            var file1 = fluid.tests.uploader.html5.mockFile("file1", "emptyfile.zip", "application/zip", 1000),
                 file2 = fluid.tests.uploader.html5.mockFile("file2", "tinyfile.jpg", "image/jpeg", 5),
                 file3 = fluid.tests.uploader.html5.mockFile("file3", "bigfile.rar", "application/x-rar-compressed", 200000);
 
--- a/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
+++ b/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
@@ -204,7 +204,8 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
             var file1 = fluid.tests.uploader.html5.mockFile("file1", "emptyfile.zip", "application/zip", 0),
                 file2 = fluid.tests.uploader.html5.mockFile("file2", "tinyfile.jpg", "image/jpeg", 5),
                 file3 = fluid.tests.uploader.html5.mockFile("file3", "bigfile.rar", "application/x-rar-compressed", 200000),
-                file4 = fluid.tests.uploader.html5.mockFile("file4", "justOver1KB.jpg", "image/jpeg", 1025);
+                file4 = fluid.tests.uploader.html5.mockFile("file4", "justOver1KB.jpg", "image/jpeg", 1025),
+                file5 = fluid.tests.uploader.html5.mockFile("file5", "exact1KB.jpg", "image/jpeg", 1024);
 
 
             /******************************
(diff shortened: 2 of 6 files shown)


src/components/uploader/js/HTML5UploaderSupport.js
```diff
@@ -251,7 +251,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             var file = files[i];
             if (fileLimit && remainingUploadLimit === 0) {
                 that.events.onQueueError.fire(file, fluid.uploader.queueErrorConstants.QUEUE_LIMIT_EXCEEDED);
-            } else if (file.size >= sizeLimit) {
+            } else if (file.size > sizeLimit) {
```
jobara: Could you add some unit tests for this?

jobara:
Also, the commit log should start like "FLUID-4713:"

aashay201297:
okay. I am on it.

aashay201297:
@jobara As per my understanding I have to add a test in addFiles() tests (https://github.com/fluid-project/infusion/blob/master/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js#L386)
I thought of inserting another element in addFilesTests list but couldn't make out the working of the variables there. I am little bit confused with the work flow of step1Events and step2Events. Can you please guide me a little bit?

jobara:
@aashay201297 so it looks like that fixture is iterated over ( https://github.com/fluid-project/infusion/blob/master/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js#L448 ) and with each case processed by [`fluid.tests.uploader.addFilesTest`](https://github.com/fluid-project/infusion/blob/master/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js#L426-L446). The step1 and step2 seem to refer to separate operations for adding [files](https://github.com/fluid-project/infusion/blob/master/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js#L204-L206) to the queue. Hope that helps.

aashay201297:
@jobara thanks for the info. 
I have another doubt in https://github.com/fluid-project/infusion/blob/master/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js#L204, zero is the size of mockfile but in what units? 
For example, in https://github.com/fluid-project/infusion/blob/master/tests/component-tests/uploader/js/HTML5UploaderSupportTests.js#L390  fileSizeLimit is in KB (this is evident when I run the tests in browser, it shows: "the fileSizeLimit is 1 KB"). 
I have this doubt because when I change the size of file1 to 5, it still doesn't give any error in the first test of addFilesTests which has fileSizeLimit: 1.

jobara:
@aashay201297 I'd assume it's in bytes, but let me know if that doesn't work, and I'll try to dig in more.

aashay201297:
@jobara Okay. I'll do some trial and error, and will let you know if this works.

aashay201297:
@jobara You were right about the size thing. 
So, what I did is change the existing test which had 0 fileSizeLimit to 1000 (i.e. 1KB) and this passes the first test (in the addFilesTests) that means that I am able to upload a file of size equal to the fileSizeLimit.

jobara:
@aashay201297 could you also add a test for something that is just over the limit. Also, not 100% that 1000 is correct, it may actually be 1024 that's the limit in bytes.

aashay201297:
@jobara Yes, the limit for bytes is 1024 indeed. 
So, I changed 1000 to 1024.
Also, I added a test case for a file of size 1025 which gives error upon uploading (as expected).
Any more suggestions you would like to give?

tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
```diff
@@ -388,38 +389,45 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
             var addFilesTests = [ {
                 fileUploadLimit: 3,
                 fileSizeLimit: 1,
-                step1Events: ["afterFileQueued", "afterFileQueued", "onQueueError"],
-                step1Files: 2, // Test #1: Two out of three files should have been added to the queue. The third is too large.
-                step2Events: ["afterFileQueued", "onQueueError", "onQueueError"],
+                step1Events: ["afterFileQueued", "afterFileQueued", "onQueueError", "onQueueError"],
+                step1Files: 2, // Test #1: Two out of four files should have been added to the queue. The third and fourth is too large.
```
jobara: minor grammar change needed. Should be "are too large".

tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
```diff
@@ -201,9 +201,10 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 
             jqUnit.module("Uploader Basic Tests");
 
-            var file1 = fluid.tests.uploader.html5.mockFile("file1", "emptyfile.zip", "application/zip", 0),
+            var file1 = fluid.tests.uploader.html5.mockFile("file1", "emptyfile.zip", "application/zip", 1024),
```
jobara: It seems that this first one was specifically an empty file, as the name suggest. May instead make another new file for the 1KB size.

tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
```diff
@@ -201,9 +201,10 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
                 file2 = fluid.tests.uploader.html5.mockFile("file2", "tinyfile.jpg", "image/jpeg", 5),
-                file3 = fluid.tests.uploader.html5.mockFile("file3", "bigfile.rar", "application/x-rar-compressed", 200000);
+                file3 = fluid.tests.uploader.html5.mockFile("file3", "bigfile.rar", "application/x-rar-compressed", 200000),
+                file4 = fluid.tests.uploader.html5.mockFile("file4", "justbig.jpg", "image/jpeg", 1025);
```
jobara: maybe call this "justOver1KB" or something like that.

tests/component-tests/uploader/js/HTML5UploaderSupportTests.js
```diff
@@ -388,38 +389,45 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-                step2Files: 3  // Test #2: All three files should have been added to the queue.
+                step1Events: ["afterFileQueued", "afterFileQueued", "afterFileQueued", "afterFileQueued"],
+                step1Files: 4, // Test #1: All four files should have been added to the queue.
+                step2Events: ["afterFileQueued", "afterFileQueued", "afterFileQueued", "afterFileQueued"],
+                step2Files: 4  // Test #2: All four files should have been added to the queue.
+            }, {
+                fileUploadLimit: 4,
+                fileSizeLimit: 1,
+                step1Events: ["afterFileQueued", "afterFileQueued", "onQueueError", "onQueueError"],
+                step1Files: 2, // Test #1: Two out of four files should have been added to the queue. The third and fourth is too large.
```
jobara: similar grammar error, should be "are too large"

aashay201297:
@jobara I think the file with size 200000 is not really testing some particular case. So, I guess I should substitute it with the file of 1KB (i.e. 1024). What do you suggest?

jobara:
@aashay201297 hmm.. not sure, perhaps that one was related to a past bug.

aashay201297:
@jobara okay cool. I will add a new file and update the test cases accordingly.

jobara:
Merged at 81f4a143870966fda0ce8384916efd6eedc75586

PR #821, opened 2017-03-02
FLUID-6141: Address mapping of "model." preference map paths.

jobara:
Only paths starting with "model." and assigned the keyword value of "default" will have a model relay created. The rest will be assigned directly into the options.

https://issues.fluidproject.org/browse/FLUID-6141

Related Docs PR: https://github.com/fluid-project/infusion-docs/pull/118


amb26:
The use of the name "default" for this path seems a bit misleading, given you are binding it into a relay relationship where it is constantly updated. Does it make sense for defaults to keep changing? We should talk some more about what the meaning of these paths actually is. From the writeup on the JIRA, the problem sounds so severe that one wonders how the prefs framework worked at all. Perhaps "value" is a more sensible name for this path? This is a convention we've used elsewhere.

jobara:
@amb26 the reason for "defaults", at least historically, is because it maps to the "default" property in the primary schema, when no stored value is present. 

The reason why things currently work is that any model path is bound to a model relay to the preference's model value. Currently all preferenceMaps express only a single model path.

jobara:
Had some further discussion in the fluid-work IRC channel with @amb26 this morning. https://botbot.me/freenode/fluid-work/2017-03-06/?msg=82000476&page=1 The suggestion is still to change "default" to "value".

jobara:
@amb26 this and the related docs pr are ready for more review.

PR #822, opened 2017-03-10
FLUID-6139: Add code coverage...

the-t-in-rtf:
See [FLUID-6139](https://issues.fluidproject.org/browse/FLUID-6139) for details.


the-t-in-rtf:
@amb26, this is in a state where I can consistently run coverage reports against Infusion, which may be of use to you in other work.

I can't assign this, but IMO this would be a good one for you to pick up once gpii-testem itself is reviewed and released.

the-t-in-rtf:
I can finally get these tests to run consistently, see [this comment on the gpii-testem PR](https://github.com/GPII/gpii-testem/pull/1#issuecomment-293245319).  I will update this pull to remove previous approaches, bring in the conflicting all-tests file, and advise when it is ready for review.

the-t-in-rtf:
I have cleaned up the cruft from previous approaches and updated to use the latest dev release of gpii-testem, which includes the defaults that address the "browser disconnect" errors seen earlier.  I can now consistently run tests in Vagrant, including code coverage reporting.  In summary, the latest work:

1. Uses gpii-testem instead of a testem.json file.
2. Adds code coverage reporting across browsers and node.
3. Preserves the qunit-composite rollups for browser tests.
4. Replaces many previous usages of Grunt with named (and documented) npm scripts.

@amb26, this is ready for further review.

amb26:
This PR can now be updated to the npm release of gpii-testem, as well as resolving conflicts. Hopefully there is some way of getting github's diff view to omit any changes resulting from drift in Infusion's master over the course of this pull.
What is our current feeling on trajectory of yarn, npm5, etc? Do we still feel it is appropriate to add yarn/npm lockfiles or both/neither?

the-t-in-rtf:
I have updated the PR to use the released version of gpii-testem and verified working.  I will work on the conflicts shortly.

My feeling based on discussions and experiences to date is that infusion and universal shouldn't add lock files until we have exercised things further in more peripheral packages.

the-t-in-rtf:
I have verified the tests working in Vagrant.

README.md
```diff
@@ -181,37 +199,67 @@ All of these libraries are already bundled within the Infusion image.
...
+* Does not require testem to be installed on the host machine.
+* Allows other applications on the host machine to have focus while the tests are run.
+* Isolates the test run from issues specific to one operating system or machine.
 
-When this VM is first created Chrome and Firefox will be upgraded to the latest versions available in the Fedora and Google package repositories. The ``vagrant provision`` command can be used at a later time to trigger the browser upgrade and general VM provisioning mechanism.
+Before you can run tests within a VM, your machine will need to meet the
+[QI development VM requirements](https://github.com/GPII/qi-development-environments/#requirements).  Once you have that,
+a [Fedora VM](https://github.com/idi-ops/packer-fedora) can be automatically created using tools provided by the
+[Prosperity4All Quality Infrastructure](https://github.com/GPII/qi-development-environments/). The command
+``npm run test:vagrant`` (or ``yarn run test:vagrant``) will create the VM (if neeeded) and run the Infusion tests in
```
cindyli: Wonder if commands `npm run test:vagrantBrowser` and `npm run test:vagrantNode` shall be provided to be in consistent with running tests on your machine.

tests/component-tests/progress/html/Progress-test.html
```diff
@@ -21,6 +21,10 @@
         <script type="text/javascript" src="../../../test-core/jqUnit/js/jqUnit.js"></script>
         <script type="text/javascript" src="../../../test-core/jqUnit/js/jqUnit-browser.js"></script>
 
+        <!-- Required to ensure we sent coverage data at the end of the test run on this page -->
+        <script src="/testem.js"></script>
```
cindyli: It would be nice to be consistent with including `type="text/javascript"` at using `<script>` tags.

tests/framework-tests/core/html/Caching-test.html
```diff
@@ -21,6 +21,10 @@
         <script type="text/javascript" src="../../../lib/mockjax/js/jquery.mockjax.js"></script>
         <script type="text/javascript" src="../../../test-core/jqUnit/js/jqUnit.js"></script>
 
+        <script src="/testem.js"></script>
```
cindyli: Would be helpful to indicate these .js files are for reporting coverage data, just like what you did with some other test files such as `UploaderCompatibility-test.html`, the one above. This comment applies across to other files that have the same issue.

tests/framework-tests/core/html/FluidPromises-test.html
```diff
@@ -17,10 +17,18 @@
         <script type="text/javascript" src="../../../../src/framework/core/js/ModelTransformation.js"></script>
         <script type="text/javascript" src="../../../../src/framework/core/js/ModelTransformationTransforms.js"></script>
 
+
+        <!-- This is required not for the tests themselves, but for the coverage client -->
+        <script type="text/javascript" src="../../../../src/framework/core/js/FluidDocument.js"></script>
```
cindyli: Is this js file still required for the coverage client? This comment also applies to other files that have this file added for the need of the coverage client.

tests/framework-tests/core/html/Keyboard-a11y-test.html
```diff
@@ -14,9 +14,17 @@
 
     <!--  These are the jqUnit test js files -->
     <script type="text/javascript" src="../../../lib/qunit/js/qunit.js"></script>
+    <script src="/testem.js"></script>
```
cindyli: Would be neater to move the inclusion of `testem.js` down below to be with other files that are added for the coverage client.

tests/lib/qunit/addons/composite/qunit-composite.js
```diff
@@ -39,6 +39,10 @@ function runSuite( suite ) {
 	}
 
 	QUnit.asyncTest( suite, function() {
+		// See: https://github.com/JamesMGreene/qunit-composite/issues/20
```
cindyli: Is this change required? Seems improper to me to make change directly in a lib file.

tests/framework-tests/core/html/Keyboard-a11y-test.html
```diff
@@ -14,9 +14,17 @@
...
     <script type="text/javascript" src="../../../../src/framework/core/js/Fluid.js"></script>
     <script type="text/javascript" src="../../../test-core/jqUnit/js/jqUnit.js"></script>
 
+
+    <!-- These are required not for the tests themselves, but for the coverage client -->
+    <script type="text/javascript" src="../../../../src/framework/core/js/DataBinding.js"></script>
+    <script type="text/javascript" src="../../../../src/framework/core/js/FluidIoC.js"></script>
```
cindyli: Could you double check if these 2 files are needed? The same comment applies to other spots.

package.json
```diff
@@ -23,25 +23,33 @@
     "main": "./src/module/fluid.js",
     "scripts": {
         "prepublish": "grunt buildDists",
-        "test": "node tests/node-tests/basic-node-tests.js"
+        "pretest": "node node_modules/rimraf/bin.js reports/* coverage/*",
+        "test": "npm run test:browser && npm run test:node",
+        "test:browser": "node node_modules/testem/testem.js ci --file tests/testem.js",
+        "test:node": "node node_modules/istanbul/lib/cli.js cover tests/node-tests/basic-node-tests.js --report none --print none",
```
cindyli: Thanks for taking the chance to automate the node tests as well. This is great and helpful!

tests/lib/qunit/addons/composite/qunit-composite.js
the-t-in-rtf: It was required when I did my original testing, will confirm whether it's still needed now.

tests/framework-tests/core/html/Keyboard-a11y-test.html
the-t-in-rtf: Very good catch, this was back when I thought that was an actual requirement of the code coverage client, when in fact it was only a requirement of the test fixtures in the gpii-testem package.  Removed throughout.

tests/framework-tests/core/html/Keyboard-a11y-test.html
the-t-in-rtf: Agreed, and done throughout.

tests/framework-tests/core/html/FluidPromises-test.html
the-t-in-rtf: See below.

tests/component-tests/progress/html/Progress-test.html
the-t-in-rtf: Done.

tests/framework-tests/core/html/Caching-test.html
the-t-in-rtf: Done.

README.md
the-t-in-rtf: Good idea, I added additional scripts and updated the docs.  I checked the previous universal PR to confirm that the wording roughly matches.

tests/lib/qunit/addons/composite/qunit-composite.js
the-t-in-rtf: It appears that this is no longer needed, at least not in single runs with Vagrant (Firefox and Chrome tested) or on my machine (Safari, Opera, Chrome, Firefox, and PhantomJS tested).  I have removed that workaround, if it comes up again we have earlier commits and this conversation to refer to.

package.json
the-t-in-rtf: I'm happy to do it, if we take the time to write tests we should also make it easy to run them often.

the-t-in-rtf:
OK, @cindyli, I believe I've addressed your feedback for this round.  Thanks again for your incredible eye for detail and good ideas.

the-t-in-rtf:
There was a merge conflict, I resolved it (additional dependencies in one test and minor formatting changes).

the-t-in-rtf:
I hadn't looked at the coverage report in a while, looks like something is wrong, it's only reporting on one file.  @cindyli, I will look into it and report back when it's ready to review again.

the-t-in-rtf:
Here's an example of the current output:

http://the-t-in-rtf.github.io/coverage-reports/infusion/20170717/

You can see that only `src/module/fluid.js` itself is counted in the tests.

the-t-in-rtf:
I figured it out late in my shift, there's a timing bug in `gpii-testem` where Testem can start before the routing rules that load the instrumented code are in place.  I will write up a ticket and work on that tomorrow.

the-t-in-rtf:
Here's the bug report, which I will be working to fix today: [GPII-2507](https://issues.gpii.net/browse/GPII-2507)

the-t-in-rtf:
@cindyli, with the new fix, this should be ready for review again.

jobara:
@the-t-in-rtf I just ran this and had every test fail. When watching it run using vagrant, all the browser tests said "fluid" is undefined. However, in the console log on the host machine the error is "Uncaught ReferenceError: jQuery is not defined".

I ran my tests by merging your branch with the current master. I ran `npm install` and destroyed my existing vm. I then launched the tests using `npm run test:vagrant`

Here's the summary of results.

1..339
# tests 339
# pass  0
# skip  0
# fail  339

cindyli:
I hit the same global failures of `Uncaught ReferenceError: jQuery is not defined` and `Uncaught ReferenceError: jQuery is not defined` that @jobara encountered at running `npm run test:vagrantNode` with the vagrant VM. 

All node tests pass with `npm run test:vagrantNode` command.

the-t-in-rtf:
I suspect that `npm run test:vagrantBrowser` fails in a fresh VM because `npm install` hasn't been run yet.  In the (later) linux pull, I explicitly added that to the npm script.  I will test this theory and confirm.

the-t-in-rtf:
OK, I figured it out, basically the instrumentation pass that's supposed to be performed by the gpii-testem component is somehow not happening, or not happening in time.  Since my testem had successfully instrumented the code at some time in the past, I did at least have a copy of infusion, but I was getting errors because I'm missing newer methods from master that are used in the tests. 
 Once I blew the instrumented directory away, I received "jQuery is not defined" errors on the next run.  Investigating now.

the-t-in-rtf:
I went through two passes, one to add proper error handling so we could see the instrumentation error in gpii-testem, another to work around the error.

There is some follow-up work to be done in [GPII-2518](https://issues.gpii.net/browse/GPII-2518), but with my temporary fixes to gpii-testem, the instrumentation succeeds and the browser tests pass again, both on a local machine, and in Vagrant.

the-t-in-rtf:
The longer-term fix for GPII-2518 is now in place, @cindyli  and @jobara, this should be ready for further review.

cindyli:
I tried the latest change on 2 macOS a couple of times and hit the same error everytime:

```
ok 1345 Chrome 60.0 - PrefsEditor Tests: Init with site defaults different from PrefsEditor control values
not ok 1346 Firefox 54.0 - error
    ---
        message: >
            Error: Browser disconnected
            
        Log: |
            { type: 'error', text: 'Error: Browser disconnected' }
    ...
ok 1347 Chrome 60.0 - PrefsEditor Tests: PrefsEditor Integration tests
```

As you can see, the error says "Browser disconnected" which seems not tests relevant. Also, the error complains about the disconnection on Firefox while running tests in Chrome, at the moment running all tests in firefox has finished.

@jobara, can you give it a try? Thanks.

the-t-in-rtf:
@cindyli, recent changes to gpii-testem should have prevented the disconnect errors, the timeouts should not be hit for five minutes.  Can you blow away node_modules, run npm install, and confirm whether you still see the problem?

cindyli:
@the-t-in-rtf, I re-populated /node_modules and re-provisioned VM, then ran `npm run test:vagrantBrowser` 4 times. Only the first time hit `Browser disconnected` and the following tests passed all. Now I suspect the error at the first time might be caused by npm cache and your update might have fixed this issue. Let's see how it goes with @jobara's test run.

the-t-in-rtf:
@cindyli and @jobara, any updates? I will have more bandwidth to look into this in the next few days, it would help to have feedback.

cindyli:
ummm.., running `npm run test:vagrantBrowser` runs into `Browser disconnected` error from time to time. Just ran that command 3 times and hit that error in 2 of them.

the-t-in-rtf:
Hi, @cindyli.  I just ran the browser tests four times in a row in a new VM with zero errors, so we need a way to narrow this down further.

My first theory is again regarding the npm dependencies.  Your comments earlier about the npm cache are interesting, as that would only be relevant if you were installing the npm dependencies from the host machine and not from within the VM (the VM should have its own npm cache).  If the cache was indeed relevant, and you were installing the dependencies from the host, then you may be hitting another weird variation of cross-platform dependencies being loaded within the VM.  Just to be absolutely sure, can you try the following from the client VM?

1. `cd /home/vagrant/sync`
2. `rm -rf node_modules`
3. `npm install`
4. `npm run test:browser && npm run test:browser && npm run test:browser`

My second theory is regarding performance, which is more difficult to track down.  It may simply be that my machine is fast and/or unloaded enough to not encounter the problem.  I am trying a few strategies to load my machine enough, but thus far haven't been able to reproduce.  If the above doesn't help, can you monitor your machine's stats during a successful/unsuccessful run?  That might help confirm or set aside that theory.

the-t-in-rtf:
UPDATE:  I managed to reproduce, but only once, and only when my machine was sufficiently loaded.   Just to confirm, I got a single Firefox error.  Cindy, are you seeing FF or Chrome errors in your runs?

the-t-in-rtf:
This is one FF failure in around 20 runs, and I had to run another VM and a CPU-intensive program in the background even to get that.

cindyli:
Seems the error might be caused by machine performance.

I did the steps you mentioned within the VM, except `rm -rf node_modules` which throws `device or resource busy` error that might be caused by [this path being mounted as a separate drive(?)](https://unix.stackexchange.com/questions/11238/how-to-get-over-device-or-resource-busy#comment-175806), so this particular step was done via my host machine terminal.

Then I ran `npm run test:browser` 3 times inside VM and hit `Browser disconnected` at the first 2 times. The CPU usage for each run was super high and each run took over 40 mins to complete.

screen shot 2017-09-08 at 9 34 14 am

This machine is a bit old for dealing with VMs. My office machine is a two year old high performance mac but I also run into this issue sometimes even though less.

Yes, all errors I hit were with firefox.

the-t-in-rtf:
So, I was scratching my head thinking how to proceed.  We previously had similar problems with Chrome, which I addressed by adding additional performance-related options to the defaults gpii-testem uses.  Unfortunately, looking at [the documentation for the command-line options for Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options), I don't see as many fine-grained controls.  The only thing even close is --safe-mode, which seems dangerous to me, as it might disable default extensions and make the browser less of a fair test of what people are running.

In looking at the Firefox command-line options, I noticed the "headless" option.   I read up a bit, "headless" mode for Firefox is already live on Windows, and hopefully not far on the Mac (you have to run the developer preview to see it).  I think "headless" mode might help with this problem.

A bit of background, a few weeks ago, I was experimenting with "headless" Chrome and gpii-webdriver.  In my admittedly informal testing, "headless" Chrome took about 40% of the time to run, which to me suggests that it might require less of a given machine, especially in running many tests.  I have created [GPII-2590](https://issues.gpii.net/browse/GPII-2590) to update gpii-testem to use headless mode by default for Chrome and Firefox.  With luck, this will give us another option in avoiding inconsistent problems like these.  I will report back when I have made any progress there.

the-t-in-rtf:
I updated a VM to use headless versions of Chrome and Firefox.  The speed was roughly the same, but CPU usage was around 5% lower throughout the run.  Unfortunately there are two unique test failures running in FF headless, so to take advantage of the lower CPU usage we'd need to figure out how to troubleshoot errors unique to headless without any kind of display output.  If I can reproduce locally, I may be more easily able to attach a remote debugger to a headless instance, but even then, it's likely to be a lot of work for little return.

For future reference, the errors in FF headless seem to be with checking focus:

```
not ok 115 Firefox 55.0 - GPII OAuth2 Component Tests: ../../../gpii/node_modules/gpii-oauth2/gpii-oauth2-authz-server/webTests/all-tests.html

        message: >
            Component Tests: ./components/privacySettings/html/PrivacySettings-test.html: Add authorization: Verify the entire process of adding an authorization: The focus returns back to the add service button - at sequence position 6 of 6

not ok 105 Firefox 55.0 - Add authorization: Verify the entire process of adding an authorization
        message: >
            The focus returns back to the add service button - at sequence position 6
```
In reviewing for this, I see that we disable extensions in Chrome to avoid disconnects, I will try the same with FF (headed) to see if that offers more performance gains without breaking any tests.

the-t-in-rtf:
Safe mode doesn't seem to offer much if any performance benefit, and requires manual intervention to click the "start in safe mode" prompt, which makes it totally unsuitable for CI.  Will keep thinking about this and comment further in a bit.

the-t-in-rtf:
@avtar and others, we have a situation here where tests fail on a particular older local machine, it would be really helpful if we could confirm that the tests consistently pass.  Is there some reason why we aren't seeing comments from the CI bot in this PR?  They'd be very helpful in this particular conversation.

amb26:
I tried this out in my raw Windows box (Windows 7 x64) and got the following:

Chrome tests ran without error, although gave dozens of error reports of the following style:
```
Error: request aborted
    at IncomingMessage.onAborted (E:\Source\gits\infusion-master\node_modules\gpii-express\node_modules\raw-body\index.js:269:10)
    at emitNone (events.js:86:13)
    at IncomingMessage.emit (events.js:185:7)
    at abortIncoming (_http_server.js:280:11)
    at Socket.serverSocketCloseListener (_http_server.js:293:5)
    at emitOne (events.js:101:20)
    at Socket.emit (events.js:188:7)
    at TCP._handle.close [as _onclose] (net.js:492:12)
```
I guess we can probably suppress these if they are irrelevant - but do we know what the cause of them is?

IE11 tests aborted on the first run, with the following:
```
not ok 705 IE - error
    ---
        message: >
            Error: Browser exited unexpectedly

        Log: |
            { type: 'error', text: 'Error: Browser exited unexpectedly' }
    ...

1..705
# tests 705
# pass  704
# skip  0
# fail  1
```
(The 704 passed tests were from the Chrome run)

avtar:
@the-t-in-rtf Fluid's CI service only runs jobs when changes are merged into master branches and not for PR jobs. Fluid's CI has lagged behind what's been put in place for GPII. That also meant we learned what didn't work well for the latter, i.e. the abandoned Jenkins PR plugin that requires [workarounds](https://github.com/GPII/ci-service/blob/master/jenkins_jobs/macros.yml#L48-L58) so that it uses the correct PR branches but the workaround also prevents it from running jobs for branch merges. To avoid putting all that in place again we'll be looking at GitLab CI for Fluid after the October P4A plenary meetings.

the-t-in-rtf:
I confirmed with @amb26 that IE11 tests are required, I will look at that shortly.

the-t-in-rtf:
Hi, all.  I finally had a chance to revisit this.  I updated to pull down recent changes to master and updated the dependency on gpii-testem.  Although I can still periodically see "browser disconnect" errors in Firefox when my machine is heavily loaded, under light load the Firefox tests run consistently.   I have zero problems with IE11 and the most recent VM.   @amb26, can you pull down the recent changes and confirm if IE11 is working for you?

amb26:
For the first test run in IE11 I got the following failure, which may reflect a genuine race condition in the test - 
```
not ok 733 IE 11.0 - Framework Tests: ./framework-tests/core/html/FluidJS-test.html
    ---
        expected: >
            5
        stack: >
               at runLoggingCallbacks (http://localhost:7357/2847/tests/lib/qunit/js/qunit.js:1609:4)
               at push (http://localhost:7357/2847/tests/lib/qunit/js/qunit.js:958:3)
               at assert.strictEqual (http://localhost:7357/2847/tests/lib/qunit/js/qunit.js:613:3)
               at jsUnitCompat.assertEquals (http://localhost:7357/2847/tests/test-core/jqUnit/js/jqUnit.js:137:13)
               at Anonymous function (http://localhost:7357/2847/tests/framework-tests/core/js/FluidJSTests.js:323:13)
        message: >
            Fluid JS Tests: fluid.debounce: The last value should be returned when accepting the trailing response
        Log: |
    ...
```
What do you think, @jobara  ?

amb26:
I got the same failure on the 3rd run. On the 2nd run the FluidView tests failed because somehow the IE11 window didn't get focused when it started up.

amb26:
On the 4th run I got this failure from Chrome: 
```
not ok 1 Chrome 62.0 - Global error: Uncaught ReferenceError: jQuery is not defined at http://localhost:7357/coverage/client/coverageSender.js, line 103
    ---
        Log: |
            { type: 'error',
              text: 'Uncaught ReferenceError: jQuery is not defined at http://localhost:7357/coverage/client/coverageSender.js, line 103\n' }
    ...
not ok 2 Chrome 62.0 - global failure
    ---
        stack: >
            http://localhost:7357/coverage/client/coverageSender.js:103
        message: >
            Uncaught ReferenceError: jQuery is not defined
        Log: |
    ...
```
Is there anything we could do about the millions of "BadRequestError: request aborted" notices in the transcript? it makes it hard to look for the genuine failures

amb26:
@Jobara's #866 seems to help with the IE11 problems, will merge after one more test.
There are still some issues with the coverage report - 
i) jquery's files are included from source/gits/infusion-master/src/lib/jquery which should not be
ii) The coverage aggregation still seems to be incomplete. For example, looking at reports/source/gits/infusion-master/src/framework/core/js/DataBinding.js.html it claims that the interior of fluid.model.accessWithStrategy at line 98 is never reached, which I know it is - it is tested by this file's own tests

amb26:
A fresh Christmas present of a previously uncharacterised variety:
```
not ok 1273 IE 11.0 - Component Tests: ./component-tests/reorderer/html/ImageReorderer-test.html
    ---
        stack: >
               at runLoggingCallbacks (http://localhost:7357/481/tests/lib/qunit/js/qunit.js:1609:4)
               at pushFailure (http://localhost:7357/481/tests/lib/qunit/js/qunit.js:996:3)
               at Test.prototype.run (http://localhost:7357/481/tests/lib/qunit/js/qunit.js:212:4)
               at Anonymous function (http://localhost:7357/481/tests/lib/qunit/js/qunit.js:365:5)
               at process (http://localhost:7357/481/tests/lib/qunit/js/qunit.js:1457:4)
               at next (http://localhost:7357/481/tests/lib/qunit/js/qunit.js:1450:3)
        message: >
            Image Reorderer Tests: MultiKeySetMovement: Died on test #5    at Anonymous function (http://localhost:7357/481/tests/component-tests/reorderer/js/I
mageReordererTests.js:548:9)
               at mightThrow (http://localhost:7357/481/src/lib/jquery/core/js/jquery.js:9:154347)
               at __cov_aTqLg4zzx9FJmEqTPxs7Zw.b[563][1] (http://localhost:7357/481/src/lib/jquery/core/js/jquery.js:9:156359): Unspecified error.
        Log: |
    ...
ok 1274 IE 11.0 - Component Tests: ./component-tests/reorderer/html/LayoutReorderer-test.html
```

the-t-in-rtf:
@amb26, I am reviewing my current fix to confirm that I can reliably avoid test failures in Firefox, and will commit that before investigating the next problem.  Just to confirm, how frequently does the new error come up, and does it only come up in IE?

the-t-in-rtf:
I finally hit another "jQuery is undefined" error yesterday.  In thinking that through, I think I have a good idea of a potential reason why pausing before running tests helps with test stability.

I suspect the instrumentation phase is not actually complete when the tests are started.  This might result in random failures to include content, which is world-breaking and obvious if that content happens to be jQuery, but which might be a lot less obvious and seemingly random if other content is left out.

Excluding our copy of jQuery from the report is fairly simple.  The missing coverage data is more complex.  That could be either the result of not instrumenting the code, or not transmitting the browser coverage data, or not including that data when creating the final report.

I have been thinking through hardening the instrumentation phase and adding tests for complex scenarios like we see with infusion and universal, I may need to pause and work on that (see [GPII-2736](https://issues.gpii.net/browse/GPII-2736)) before I can truly unstick this pull.

the-t-in-rtf:
Hi, all.  I have done a deep dive on this one this week, and it seems to be in a much better state after the fix to gpii-testem.  In recent tests on all platforms, the tests complete with the same number of test results and good coverage reports.   I do still see "BadRequest" messages, but the timeout didn't seem to help with that and the tests and reports complete consistently.

I left the previous "pause on startup" harness commented out for now, @amb26, if you can confirm the recent state completes without browser disconnects, I'll get rid of that.

the-t-in-rtf:
Hi, all.  Following the recent changes, I have not seen any further errors in multiple linux, OS X, or Windows runs.  The code to artificially introduce a pause in the Testem runs is not currently enabled, I would still like to confirm with @amb26 whether the pause seems necessary in his environment.

The CI builds are failing, but they are failing to provision the machine rather than run the tests.  Copying in @avtar to get his input.

avtar:
@the-t-in-rtf I did a search for ``eslint`` in the output and looks like grunt is complaining about that task. I haven't gone through your changes but are you able to finish this build locally?

the-t-in-rtf:
I had honestly been testing heavily in a windows VM, so I went back to confirm that a) the lint checks pass with the current branch, and b) `vagrant up && npm run test:vagrant` works for me.

@amb26 also reported build instability with on focus tests with the uploader, I'm hoping to get more details to reproduce.

the-t-in-rtf:
Hi, all.  I have been working on improvements to gpii-testem that will better support instrumenting and collecting coverage data in cross-environment projects like infusion.  I'm happy to report that I was able to create a much cleaner report as a result of these efforts:

http://the-t-in-rtf.github.io/coverage-reports/infusion/20180110/

As there were previously stability problems in Windows, I plan to update this a bit further:

1. To enable running Vagrant tests against a Windows VM.
2. To force the use of npm@3 in the linux scripts to avoid creating package lock files when running tests).

I will comment here once that's ready, but my latest commit should at least give us something to talk about WRT to the buildkite builds.

the-t-in-rtf:
The latest build failure was my fault, I am testing the fix locally.

the-t-in-rtf:
With the fix, the buildkite builds pass.  I am testing manually in Windows now.

the-t-in-rtf:
In refreshing my memory, I see that the builds use the old manual one-off config and provisioning, I'd rather wait to add Windows support in a separate pull that converts this to using the Vagrant CI Plugin.

the-t-in-rtf:
I have run the tests four times in Windows with no unexpected test failures.  IMO this is ready to review again. I'd especially like to confirm that it works for @amb26.

amb26:
I still get about one "Request aborted" message for every 2 tests, but am happy to merge this since it seems it works reliably

PR #824, opened 2017-03-28
FLUID-6148: Implement "potentia II" constituting aligned, transactional records of author's creational intent

amb26:
This branch is complete and ready to review. It resolves the following issues:

https://issues.fluidproject.org/browse/FLUID-6146 Mutual reference between createOnEvent components
https://issues.fluidproject.org/browse/FLUID-6147 Transactional units for component creation
https://issues.fluidproject.org/browse/FLUID-4925 "wave of explosions"
https://issues.fluidproject.org/browse/FLUID-5519 Problematic timing of model init transaction
https://issues.fluidproject.org/browse/FLUID-5614 "double deep trees"
https://issues.fluidproject.org/browse/FLUID-5790 cancellable promises
https://issues.fluidproject.org/browse/FLUID-6390 "lensed components"
https://issues.fluidproject.org/browse/FLUID-5912 {arguments} in members and models
https://issues.fluidproject.org/browse/FLUID-6418 Failure when referring to dynamic component location during afterDestroy
https://issues.fluidproject.org/browse/FLUID-6413 Model-driven resource localisation
https://issues.fluidproject.org/browse/FLUID-6414 Allow dynamic grades to be contributed via expander

and also a basic quality implementation of 
https://issues.fluidproject.org/browse/FLUID-4982 "asynchronous ginger world", sufficient to allow progress on new renderer for FLUID-4260

It also includes fixes for the following issues internal to this branch:
https://issues.fluidproject.org/browse/FLUID-6404 Failure when calling destroy during onDestroy
https://issues.fluidproject.org/browse/FLUID-6406 Workflow algorithm allows an early block to run ahead
https://issues.fluidproject.org/browse/FLUID-6408 Obnoxiously instantiated free component disrupts workflow
https://issues.fluidproject.org/browse/FLUID-6410 Zero-element sequence leaves incomplete workflow
https://issues.fluidproject.org/browse/FLUID-6411 Corruption when cleaning up listeners to resource-based models
https://issues.fluidproject.org/browse/FLUID-6415 Corruption in graph structure with lensed components
https://issues.fluidproject.org/browse/FLUID-6424 Contextualised relay doesn't notify init modelListener


the-t-in-rtf:
I worked my way up through gpii-launcher and gpii-testem, and cut dev releases of both.  Here's the latest gpii-testem, which includes a newer launcher: 

https://github.com/fluid-project/infusion/pull/824

That should be "new enough turtles all the way down".

jobara:
@simonbates would you be able to review this PR

sgithens:
@amb26 From your comment at the top, it sounds as if this is still an early prototype version of the work, but from conversations outside of github it appears that this is now essentially ready to be fully reviewed as if it was going to go into master correct?

amb26:
Cheers @sgithens - I've updated the comment to reflect the state of the branch.

src/framework/core/js/FluidIoC.js
```diff
@@ -1499,51 +1806,154 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         return instantiator.parseEL(shadow.path);
     };
 
-    /** Construct a component with the supplied options at the specified path in the component tree. The parent path of the location must already be a component.
+    /** Begin a fresh transaction against the global component tree. Any further calls to `fluid.registerPotentia`,
+     * `fluid.construct` or `fluid.destroy` may be contextualised by this transaction, and then committed as a single
+     * unit via `fluid.commitPotentiae` or cancelled via `fluid.cancelTreeTransaction`.
+     * @param {Object} [transactionOptions] - [optional] A set of options configuring this tree transaction. This may include fields
+     *     {String} breakAt - one of the values:
```
sgithens: Is there some some reason you can only break at one of these 2 places, and not potentially both?  
`breakAt: ["shells", "observation"]`

amb26: One of them happens before the other one. If you break at the first, you won't reach the other

package.json
```diff
@@ -41,18 +41,19 @@
...
         "grunt-contrib-copy": "1.0.0",
         "grunt-contrib-stylus": "1.2.0",
-        "grunt-contrib-uglify": "3.1.0",
+        "grunt-contrib-uglify": "3.3.0",
         "grunt-contrib-watch": "1.0.0",
         "grunt-jsonlint": "1.1.0",
         "grunt-modulefiles": "0.3.1",
         "grunt-shell": "2.1.0",
-        "lodash": "4.17.4",
+        "istanbul": "git://github.com/the-t-in-rtf/istanbul#828eb930add7bd663ca20649a917f3f3829e48c5",
```
jobara: Is this being used by Infusion?

the-t-in-rtf: I believe this is no longer necessary.  The improvements I submitted in that pull (and resubmitted when they relocated the project) have been merged and released.  Everything we use (nyc, testem) should be able to use a released version rather than our work.

src/framework/core/js/Fluid.js
```diff
@@ -13,7 +13,6 @@ Copyright 2007-2009 University of Toronto
 Copyright 2007-2009 University of California, Berkeley
 Copyright 2010-2011 Lucendo Development Ltd.
 Copyright 2010-2015 OCAD University
-Copyright 2011 Charly Molter
```
simonbates: Just to check: is this removal intentional?

src/framework/core/js/Fluid.js
```diff
@@ -288,6 +312,7 @@ var fluid = fluid || fluid_3_0_0;
     /** Undo the effect of the most recent "setLogging", returning the logging system to its previous state
      * @return {LogLevel} The logLevel that was just popped
      */
+    // TODO: This function is documented, but untested
```
simonbates: We have an existing test in `FluidJSTests.js`:

https://github.com/amb26/infusion/blob/FLUID-6148/tests/framework-tests/core/js/FluidJSTests.js#L696-L709

Were you thinking we wanted something more specific for `fluid.popLogging`? Is that something we should do as part of this PR, or is just adding the `TODO` comment for the future good?

src/framework/core/js/Fluid.js
```diff
@@ -224,6 +223,31 @@ var fluid = fluid || fluid_3_0_0;
         }
     };
 
+    fluid.notrycatch = false;
+
+    // A wrapper for the try/catch/finally language feature, to aid debugging in the QUnit UI by means of exception breakpoints for
+    // uncaught exceptions, since so many libraries, e.g. jQuery throw junk caught exceptions on startup
+    fluid.tryCatch = function (tryfun, catchfun, finallyfun) {
```
simonbates: Probably worthwhile to add some tests for `fluid.tryCatch`

src/framework/core/js/Fluid.js
```diff
@@ -855,6 +880,15 @@ var fluid = fluid || fluid_3_0_0;
         return isFinite(string) && ((string % 1) === 0) ? Number(string) : NaN;
     };
 
+    /**
+     * "Ponyfill" for Number.isInteger - can be removed once we drop support for IE11
+     * @param {Any} value - A value to be tested for being an integer
+     * @return {Boolean} `true` if the value is a Number and a finite integer
+     */
+    fluid.isInteger = function (value) {
```
simonbates: I think it's worthwhile to add some simple tests for this new function.

src/framework/core/js/Fluid.js
amb26: The code they contributed was removed by previous work from @the-t-in-rtf

src/framework/core/js/Fluid.js
amb26: Good catch, thanks!

src/framework/core/js/Fluid.js
amb26: added

src/framework/core/js/Fluid.js
amb26: added

amb26:
Thanks all for comments, ready for another look

src/framework/core/js/Fluid.js
jobara: @amb26 should his Copyright be removed from FluidJSTests.js too?

src/framework/core/js/ResourceLoader.js
```diff
@@ -13,59 +13,672 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+     * during component startup (e.g. as initial model values or renderer templates), and so the `onResourcesLoaded`
+     * event may fire immediately
+     * @param {fluid.resourceLoader} that - The loader for which the I/O fetch process is to be started
+     */
+    fluid.resourceLoader.loadResources = function (that) {
+        var completionPromise = that.resourceFetcher.fetchAll();
+        completionPromise.then(function () {
+            that.events.onResourcesLoaded.fire(that.resources);
+        }, function (error) {
+            // Note that if the failure was for a resource demanded during startup, this component will already have
```
the-t-in-rtf: Shouldn't you be calling onResourceError here?

amb26: It will already have been fired by the individual resource failure on line 640

tests/framework-tests/core/js/ResourceLoaderTests.js
```diff
@@ -163,4 +269,241 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+    fluid.tests.fluid4982OverrideFixtures = [{
+        grade: "fluid.tests.fluid4982overrideBase",
+        expected: "Après moi, le déluge!"
+    }, {
+        grade: "fluid.tests.fluid4982overrideDerived",
+        expected: "upper, middle, lower"
+    }, , {
+        grade: "fluid.tests.fluid4982overridePromise",
+        expected: "upper, middle, lower"
```
the-t-in-rtf: Can you use messages1_en.json and its expected payload for the promise example?   It seems a little cleaner than getting to exactly the same outcome via two mechanisms in test definitions that are right next to one another.

amb26:
CI aborted for no visible reason - ok to test

src/framework/core/js/FluidView.js
```diff
@@ -21,13 +21,13 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
-        initFunction: "fluid.initView",
         argumentMap: {
             container: 0,
             options: 1
         },
-        members: { // Used to allow early access to DOM binder via IoC, but to also avoid triggering evaluation of selectors
-            dom: "@expand:fluid.initDomBinder({that}, {that}.options.selectors)"
+        members: {
+            container: "@expand:fluid.containerForViewComponent({that}, {that}.options.container)",
+            dom: "@expand:fluid.initDomBinder({that}, {that}.options.selectors, {that}.container)"
```
jobara: Do members support priorities for when they resolve? In UIO+ there is a"[content](https://github.com/jobara/gpii-chrome-extension/blob/master/extension/src/content_scripts/contentView.js#L49-L54)" member that uses an expander to trigger a function that uses [that.locate](https://github.com/jobara/gpii-chrome-extension/blob/master/extension/src/content_scripts/contentView.js#L87). At the moment the error "that.locate is not a function" is thrown; likely because the domBinder has yet to be instantiated.

amb26: They don't support priorities, and it would make a total mess of everyone's definitions if they needed to write such things everywhere. There are two things going on here - firstly, there is a bug in very old definition kept over from the old framework that assigns that.locate in a faulty, side-effectful way that means that the framework can't determine where it is coming from. Secondly, there is the issue we were talking about in the channel the other week, that definitions written in code need to be assisted by proxies so that they can force the framework to evaluate material like that.locate which hasn't yet been visited.
I'm totally amazed that the definition you have in UIO+ ever worked in the old framework and I can't see a clear reason for it, although it might have been something to do with the very odd way that viewComponents were instantiated through custom "init functions".
Firstly I need to fix the bug left in this branch which assigns to that.locate side-effectfully via fluid.initDomBinder, and then after I've done this you will be able to pass {that}.locate explicitly as an argument into your expander in the way we currently do to force evaluation (until we have proxies)

jobara: @amb26 thanks for the response. +1 for things just working and not having to write priorities. 

I'm assuming things are breaking now because it tries to resolve my member before the "dom" member. Maybe I was just lucky before with the timing. 

I'll hold off on continuing my integration efforts until after the bug for the assignment of that.locate has been addressed.

amb26: Hi there @jobara - you were "lucky" before because the initialisation of the DOM binder used to be special-cased in the old framework by means of the "init function" initViewComponent - init functions have now been abolished so all material attached to the component now needs to take its chances with the ordinary data-driven workflow - which was broken for that.locate since we still retained the old side-effectful function fluid.initDomBinder from the old framework. I've issued a fix that regularises the initialisation of the DOM binder (which is now defined in a new base grade fluid.baseViewComponent) and rolled a new dev release for you to test at version 3.0.0-dev.20190913T161245Z.3910173d0.FLUID-6148 . fluid.initDomBinder is now abolished and only the more reasonable fluid.createDomBinder remains.

Do give it a try when you can and see if you get get a bit further with UIO+. You will unfortunately need to supply "{that}.locate" as a surplus final argument to your locateInContext functions but at least it will now be possible for you to force the evaluation somehow.
These kinds of issues seem to make it more urgent for use to move to the proxy-wrapped framework described in https://issues.fluidproject.org/browse/FLUID-6372 so we can get rid of these surplus arguments.

src/framework/core/js/FluidIoC.js
```diff
@@ -1812,6 +1887,20 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         return togo;
     };
 
+    fluid.findWorkflowShadows = function (shadows, blockStart, blockEnd, workflowRecord) {
+        var workflowShadows = [];
+        for (var i = blockStart; i < blockEnd; ++i) {
```
jobara: Can this use `fluid.forEachInRange`

jobara:
@amb26 it looks like lensed components use `{source}`, this is similar to how array driven dynamic components work. Can the event driven dynamic components also use `{source}`? I think that would address issues where we want to access the values in places that would already use `{arguments}` and it would keep things consistent across all dynamic component types.

amb26:
I don't think swapping {arguments} for {source} would give much consistency, since the semantics are quite different. Notice that the original source material in the dynamicComponents record is called "sources" and then the particular member of sources corresponding to a particular dynamicComponent gets bound to {source} - whereas for dynamicComponents which are event driven there is just one component generated from a particular event firing which gets all of its arguments.
I take the point about {source} being less likely to hide the {arguments} context names which is in use in listeners and invokers - but this would represent a major breaking change in behaviour that I'd want to avoid at this point - it's something to consider when we are in the "smash everything" mode where, e.g. we rename fundamental things like fluid.defaults and gradeNames - but by then hopefully noone will be using event-driven dynamicComponents anyway!

tests/framework-tests/core/js/FluidIoCTests.js
```diff
@@ -5329,6 +5339,61 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+
+    fluid.defaults("fluid.tests.FLUID6148notate", {
+        gradeNames: "fluid.component",
+        listeners: {
+            "onCreate.note": {
+                funcName: "fluid.tests.FLUID6148note",
+                args: ["{fluid.tests.FLUID6148create}", "{that}"],
+                // Note that if this priority is not supplied, the order will not be as expected since component
+                // creation via "onCreate" itself will not be fully enqueued
+                priority: "after:fluid-componentConstruction"
```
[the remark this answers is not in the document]
amb26: The createOnEvent system isn't magic and has to be honoured by an actual, concrete listener added to "onCreate" - which has to take its natural place and priority with respect to the other ones. Simply adding any old "onCreate" listener to check for the creation point of the component runs the risk of racing against the framework's one.
This is a case where there has also been a change in behaviour between the old framework and the new one. The old framework did indeed use "any old listener", with respect to which the timing of another listener would be largely random - but on the other hand, guaranteed that the subcomponent would be FULLY constructed if you made your observation after the framework's one. Given that a major goal of this rewrite was to enable mutual reference between createOnEvent components created off the same event, the new framework now applies the creation in two stages. Firstly there is the same ordinary listener which starts the subcomponent creation process - but now, crucially, the framework also opens a transaction at this point, waiting for any other subcomponent creations kicked off by other ordinary listeners. Whilst this transaction is open, the subcomponents sit in a partially constructed state.
The framework then supplies a special, synthetic listener with priority "last:fluid-componentConstruction" which closes the transaction and kicks off the complete construction of all the createOnEvent subcomponents as a unit, which then allows them to refer to one another in their options.

This system is indeed rather over-complicated and arcane, but it is the only way out of a tight design corner bridging between the rather procedural semantics of events and the declarative semantics of component reference - without it, people had to make annoying redesigns of their component tree so that all the createOnEvent components had a single ancestor. This is another reason that we plan for createOnEvent components to be disused, now that we have finally implemented lensed components.

src/framework/core/js/FluidIoC.js
amb26: Yes, I guess it could - in practice I only implemented fluid.forEachInRange to get around a tight corner with scoping issues where I was binding to what should be variables scoped to the loop in the various promises, and didn't want to use "let" since it is not properly supported in IE11. But now we've paid the expense of writing the 4-line function, we may as well use it : P

tests/framework-tests/core/js/FluidIoCTests.js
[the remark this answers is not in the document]
jobara: Thanks for the explanation.

amb26:
ok to test

amb26:
Authentication failure - ok to test

amb26:
Apparently CI is now fixed - ok to test

amb26:
@gtirloni - I'm just noticing that the Github Actions CI test configuration doesn't seem correct for the browser tests - it seems only set up to run core framework tests rather than the complete test suite (hence it completes so quickly) - it should be running tests/all-tests.html rather than tests/framework-tests/all-tests.html

jobara:
@amb26 I looked at the [test run](https://github.com/fluid-project/infusion/pull/824/checks?check_run_id=706700378) for this PR and it does look like it's running both the framework and component tests. There were 246 tests with both Firefox and Chrome tested. If you expand the browser job you can see the logs. Also, it looks like @gtirloni has it setup to run the node and browser tests in separate jobs.

amb26:
Sorry, @gtirloni, @jobara, I see you're quire right, all the tests do run, I just missed the scrollbar because of the pane styling. Do we think it is ok to turn off our old Jenkins runs now then?

jobara:
I don't think this impacts if we stop running Jenkins, but in the future it would be good to explore if we can output the testem coverage reports.

PR #825, opened 2017-04-13
FLUID-5532: Responsive UIO

jobara:
Responsive styling for UIO.

Notes:
* desktop styling for 640px and wider screens
* the mobile version of the sliding panel tab should work to at least a minimum of 320px width
* adjuster panels for mobile version may require horizontal scrolling to view all of their contents. This is a design decision that @dayotte and myself discussed
* there is a bug with the initial opening of the sliding panel on mobile when lazy loading is enabled. ( See: [FLUID-6151](https://issues.fluidproject.org/browse/FLUID-6151) ) ( This issue is caused by the OpenSans font, see [comment](https://issues.fluidproject.org/browse/FLUID-6151?focusedCommentId=36720&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-36720))
* iOS does not support Comic Sans, and will just fall back to a sans serif font.

https://issues.fluidproject.org/browse/FLUID-5532


demos/prefsFramework/index.html
```diff
@@ -93,14 +95,22 @@
         <div class="flc-overviewPanel fl-overviewPanel-container"></div>
         <!-- BEGIN markup for Preference Editor -->
         <div class="flc-prefsEditor-separatedPanel fl-prefsEditor-separatedPanel">
+            <!-- This div is for the sliding panel that shows and hides the Preference Editor controls in the mobile view-->
+            <div class="fl-panelBar fl-panelBar-above">
```
cindyli: What's the reason to have two identical `div`s for showing prefs editor in the mobile view and in the desktop view?

src/framework/preferences/css/stylus/SeparatedPanelPrefsEditor.styl
```diff
@@ -23,45 +23,55 @@
...
+        &.fl-panelBar-below {
+            display: none;
         }
-    }
 
-    .fl-prefsEditor-showHide {
-        min-width: 15em;
+        .fl-prefsEditor-buttons {
+            width: 100%;
+            height: 66px;
```
cindyli: Is there a reason that the button height is defined in "px" instead of "rem"? The same comment applies to line 68 & 74.

src/framework/preferences/css/stylus/SeparatedPanelPrefsEditor.styl
```diff
@@ -73,23 +83,74 @@
...
+            font-size: 1.1rem;
+            padding: 0;
+        }
+
+        .fl-prefsEditor-showHide {
+            margin: 0 0.5rem;
+        }
+
+        .fl-prefsEditor-reset {
+            // border-right: 1px solid #ccc;
```
cindyli: Remove the commented line.

demos/prefsFramework/index.html
jobara: @cindyli one is above the sliding panel and one is below. The one above is used for mobile and the one below for desktop. It is possible to have a single one and change the placement with styling; however, doing so will break the tab order for one of the implementations. I had spoken to Lisa about that, and she indicated that it would be a WCAG failure. In the future we plane to only have the panel bar above, in all cases. However, there needs to be some design consideration worked out for how this will affect the interactions on the desktop view first.  Meaning that the change won't happen in the near future.

src/framework/preferences/css/stylus/SeparatedPanelPrefsEditor.styl
jobara: @cindyli yes, so that the button size remains fixed on mobile. When using rem values, it will cause the button to grow and push the sliding panel down as a user is interacting with it. For the other lines, it's to save space when adjusting text size in the panels. I'll add some comments to the styl file.

cindyli:
Please add the border for "show preference" button when UIO is set to be used: 1. on a mobile device; 2. when a high contrast theme is turned on.

screen shot 2017-04-17 at 11 29 40 am

jobara:
@cindyli fixed the missing border. The border was actually always there, it was just being hidden when the line-spacing and other adjustments were applied.

cindyli:
According to [the mobile designs](https://wiki.fluidproject.org/pages/viewpage.action?pageId=137658573) specified in [FLUID-5532 JIRA](https://issues.fluidproject.org/browse/FLUID-5532), moving in between UIO panels is by touching "<" to go to the previous panel and ">" to go to the next panel. The implementation in this pull request doesn't have this functionality. Instead, panels are separated with a vertical line and the scrolling of panels is same as what's for the larger screen devices (see attached screenshot). I wonder what's the thought of this implementation. Is it intermediate?

screen shot 2017-04-19 at 2 58 17 pm

dayotte:
Yes it's intermediate.

cindyli:
Merged at f8d92a56023a3ae6c1c2a20773f7ac44e1514d31

PR #827, opened 2017-04-24
FLUID-6153: Removing scrollbars from descriptions

jobara:
https://issues.fluidproject.org/browse/FLUID-6153

### DIFF
--- a/src/framework/preferences/css/stylus/SeparatedPanelPrefsEditorFrame.styl
+++ b/src/framework/preferences/css/stylus/SeparatedPanelPrefsEditorFrame.styl
@@ -151,10 +151,11 @@ html {
                 .fl-prefsEditor-panel-descr {
                     clip: auto;
                     height: auto;
-                    margin: 1rem 0;
-                    overflow: auto;
+                    margin: 0;
+                    margin-bottom: 1rem;
+                    overflow: visible;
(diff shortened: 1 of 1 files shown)


src/framework/preferences/css/stylus/SeparatedPanelPrefsEditorFrame.styl
```diff
@@ -151,10 +151,11 @@ html {
                 .fl-prefsEditor-panel-descr {
                     clip: auto;
```
cindyli: According to [CSS clip property usage at w3schools](https://www.w3schools.com/cssref/pr_pos_clip.asp): `Note: The clip property does not work if "overflow:visible".` The use of `clip: auto` is unnecessary. 

Also wondering if the use of `overflow: visible` is needed. If not specified, the inherited `hidden` value seems good too.

jobara: I think we actually want the content to be visible though. So i figure it's better to ensure that it is or at least be explicit.

jobara:
@cindyli ready for more review

cindyli:
Merged at 56cc1caafba6907f1ecf93cc3baf17b0e357ee97

PR #828, opened 2017-04-26
FLUID-5534: Adding arrows to indicate additional content

jobara:
This is an interim solution, as the arrows are not interactive and can not
be used to scroll the panels. Rather in this implementation they are just
used to indicate that there is additional content that can be manually scrolled to.

https://issues.fluidproject.org/browse/FLUID-5534

NOTE: There is a bug in IE 11 where the headers between panels do not line up. ( see: [FLUID-6154](https://issues.fluidproject.org/browse/FLUID-6154) )


cindyli:
Merged at c3b3cdf783cdc48f89e8b57ceb46552d02ec967a

PR #833, opened 2017-07-10
FLUID-6169: removed fluid.prefs.store as grade for temp and cookie store

jobara:
All stores now derive fluid.prefs.dataSource. fluid.prefs.store defaults
to use the cookie store but can use other stores based on context.

https://issues.fluidproject.org/browse/FLUID-6169


cindyli:
Merged at 7ffeb42e5d812f4d65d9a67bd833623fdb9f731f

PR #836, opened 2017-07-19
FLUID-6171: adding IoC resoltion to fluid.renderer.selection.inputs

jobara:
The renderer expander will now also resolve IoC references for the
selectID. rowID, labelID, and inputID all point at values in the
selectors block and do not resolve IoC references.

https://issues.fluidproject.org/browse/FLUID-6171

### DIFF
--- a/src/framework/renderer/js/RendererUtilities.js
+++ b/src/framework/renderer/js/RendererUtilities.js
@@ -307,17 +307,17 @@ fluid_3_0_0 = fluid_3_0_0 || {};
     fluid.renderer.selection.inputs = function (options, container, key, config) {
         fluid.expect("Selection to inputs expander", options, ["selectID", "inputID", "labelID", "rowID"]);
         var selection = config.expander(options.tree);
-        var selectID = config.expandLight(options.selectID);
+        var expandedOpts = config.expandLight(options);
         var rows = fluid.transform(selection.optionlist.value, function (option, index) {
             var togo = {};
-            var element =  {parentRelativeID: "..::" + selectID, choiceindex: index};
-            togo[options.inputID] = element;
-            togo[options.labelID] = fluid.copy(element);
+            var element =  {parentRelativeID: "..::" + expandedOpts.selectID, choiceindex: index};
+            togo[expandedOpts.inputID] = element;
+            togo[expandedOpts.labelID] = fluid.copy(element);
             return togo;
         });
         var togo = {}; // TODO: JICO needs to support "quoted literal key initialisers" :P
-        togo[selectID] = selection;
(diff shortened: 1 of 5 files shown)


amb26:
Even though rowID, labelID and inputID indirect into the selectors block, I think we may as well add resolution for those as well - since it is possible that users may want to configure these also.

jobara:
@amb26 I was talking that over with @cindyli in the morning. Because those look into the selectors block, it may actually be easier/more direct for someone to setup options distributions to remap their selector to the one that is being searched for. What do you think?

amb26:
Hi there - what you determined with @cindyli is indeed correct, the user could always configure the selectors rather than the id references. However, I think consistency is an important framework value - it wouldn't be particularly ergonomic to say "some of these values are IoC expanded and some are not" - whilst we are in this area making improvements I think it would be helpful to users of this expander to make sure that anything which could reasonably be expanded is expanded.

As a use case sketch - imagine that for some collateral reason the user would not like to reconfigure the selector in the selectors block, since its name is already being used by the component for some other purpose, e.g. some event binding.

jobara:
@amb26 I've addressed your comment. Ready for more review.

src/framework/renderer/js/RendererUtilities.js
```diff
@@ -307,16 +307,17 @@ fluid_3_0_0 = fluid_3_0_0 || {};
     fluid.renderer.selection.inputs = function (options, container, key, config) {
         fluid.expect("Selection to inputs expander", options, ["selectID", "inputID", "labelID", "rowID"]);
         var selection = config.expander(options.tree);
+        var expandedOpts = config.expandLight(options);
```
amb26: This is dangerous/inefficient since the options will contain some material that should not be fed through expandLight - primarily the tree, which will contain strings whose syntax could be confused with IoC references. I suggest making a copy using http://docs.fluidproject.org/infusion/development/CoreAPI.html#fluid-censorkeys-tocensor-keys- to filter out tree and any other similar dangerous material. This would be a great opportunity to add a docs comment to this function to explain what is the structure and meaning of the accepted arguments.

It appears that the closest thing we have resembling docs is on this page - https://wiki.fluidproject.org/display/docs/Renderer+Component+Tree+Expanders#RendererComponentTreeExpanders-SelectionToInputsExpander

jobara: There are some docs on the docs site for this at http://docs.fluidproject.org/infusion/development/RendererComponentTreeExpanders.html#selection-to-inputs-expander

jobara:
@amb26 this is ready for more review.

PR #837, opened 2017-07-25
FLUID-6172: Adding rounding to decimal to round transformation.

jobara:
https://issues.fluidproject.org/browse/FLUID-6172

Infusion-Docs PR: https://github.com/fluid-project/infusion-docs/pull/124

### DIFF
--- a/src/framework/core/js/ModelTransformationTransforms.js
+++ b/src/framework/core/js/ModelTransformationTransforms.js
@@ -102,10 +102,13 @@ var fluid = fluid || fluid_2_0_0;
         invertConfiguration: "fluid.transforms.invertToIdentity"
     });
 
-    fluid.transforms.round = function (value) {
-        return Math.round(value);
+    fluid.transforms.round = function (value, transformSpec) {
+        if (typeof transformSpec.scale === "number" && !isNaN(transformSpec.scale)) {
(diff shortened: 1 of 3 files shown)


src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -102,11 +102,14 @@ var fluid = fluid || fluid_2_0_0;
         invertConfiguration: "fluid.transforms.invertToIdentity"
     });
 
-    fluid.transforms.round = function (value) {
-        return Math.round(value);
+    fluid.transforms.round = function (value, transformSpec) {
+        if (typeof transformSpec.scale === "number" && !isNaN(transformSpec.scale)) {
```
amb26: No point copy-pasting this check from numberToString. In fact, the docs for roundToDecimal state that invalid scales will be treated as 0, and the default method is "round", so we can eliminate this check and the other branch entirely. For courtesy, we can leave a comment that validation of "scale" is done by roundToDecimal.

jobara: @amb26 and I discussed this in the [fluid-work channel](https://botbot.me/freenode/fluid-work/2017-07-27/?msg=89126963&page=1). We decided to use fluid.roundToDecimal for all the rounding.

jobara:
@amb26 ready for more review

PR #838, opened 2017-07-29
FLUID-6174: Adding an input option to fluid.transforms.valueMapper

jobara:
https://issues.fluidproject.org/browse/FLUID-6174

Infusion-Docs PR: https://github.com/fluid-project/infusion-docs/pull/125

### DIFF
--- a/src/framework/core/js/ModelTransformationTransforms.js
+++ b/src/framework/core/js/ModelTransformationTransforms.js
@@ -244,7 +244,7 @@ var fluid = fluid || fluid_2_0_0;
         if (!transformSpec.match) {
             fluid.fail("valueMapper requires an array or hash of matches at path named \"match\", supplied ", transformSpec);
         }
-        var value = fluid.model.transform.getValue(transformSpec.defaultInputPath, transformSpec.input, transformer);
+        var value = fluid.model.transform.getValue(transformSpec.defaultInputPath, transformSpec.defaultInput, transformer);
 
         var matchedEntry = (fluid.isArrayable(transformSpec.match)) ? // long form with array of records?
             fluid.transforms.valueMapper.longFormMatch(value, transformSpec, transformer) :
--- a/src/framework/core/js/ModelTransformationTransforms.js
+++ b/src/framework/core/js/ModelTransformationTransforms.js
@@ -244,7 +244,7 @@ var fluid = fluid || fluid_2_0_0;
         if (!transformSpec.match) {
             fluid.fail("valueMapper requires an array or hash of matches at path named \"match\", supplied ", transformSpec);
         }
-        var value = fluid.model.transform.getValue(transformSpec.defaultInputPath, undefined, transformer);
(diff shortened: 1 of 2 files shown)


jobara:
@amb26 could you please review this PR along with the related Infusion-Docs PR

amb26:
@kaspermarkus - please see commentary on the JIRA - is it right that this is simply a blunder that never got completely implemented when we did https://issues.fluidproject.org/browse/FLUID-5891 ?

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -244,7 +244,7 @@ var fluid = fluid || fluid_2_0_0;
         if (!transformSpec.match) {
             fluid.fail("valueMapper requires an array or hash of matches at path named \"match\", supplied ", transformSpec);
         }
-        var value = fluid.model.transform.getValue(transformSpec.defaultInputPath, undefined, transformer);
+        var value = fluid.model.transform.getValue(transformSpec.defaultInputPath, transformSpec.input, transformer);
```
amb26: The more I look at this now I am clearheaded, the more I realise we have a problem. There is already a facility implemented in the valueMapper for sourcing values directly, rather than via path - which is via the "inputValue" member.
From the docs, 
```
inputValue
The value to check the input against.
This can also be implicitly provided as a key in the top-level of options block.
Will always be interpreted literally (ie. no transforms allowed here)
```
If we *did* want to support a separate top-level property, it would seem to be better to be consistent with the general pattern with transforms of matching an xxxPath property with an xxx value - in which case this should be named "defaultInput" rather than "input".
But it does seem that this use of the term "default" here is confusing, as it frequently has been in the valueMapper. We have been plagued with various properties prefixed with "default" which use the term in slightly different senses. Note the confusion that we try to dispel in the docs with "The meaning of defaultOutputValue is NOT "the output in case no case matches" but "the outputValue that should be used in a case where it has not been explicitly written"."

It seems that there might be a case to be made for removing "default" from all of these top-level properties. This might start to create confusion from seeing properties with the same name at different levels of containment - but on the other hand might create more clarity based on the fact that the properties with the same names will directly override each other, a pattern we are familiar with elsewhere in the framework - as well as allowing us to restore the standard behaviour with respect to other transforms that the "principal input" of the whole transform is simply named "input".

What I suggest you do here is to for the time being rename this property as "defaultInput" and we separately JIRA and discuss with @kaspermarkus when he returns the overall issue of whether and when we should deprecate all top-level defaultXxxx properties in favour of implementing them as xxxx. I imagine our current policies would allow us to remove the defaultXxxx properties in something like Infusion 4.0.

jobara: I filed a new JIRA ( https://issues.fluidproject.org/browse/FLUID-6179 ), but you will probably want to flesh it out. Regarding the removal time, we could remove it at Infusion 3.0, but it would be more user friendly if we had at least a point release with it deprecated.

jobara:
@amb26 ready for more review.

PR #841, opened 2017-08-09
FLUID 5879: Exclude web icon fonts from font changes

grrrero:
https://issues.fluidproject.org/browse/FLUID-5879

Moved font family styling to its own stylus so that it's scalable when adding more font families in the future. Also excluding any of class "icon"


src/framework/preferences/css/stylus/Enactors.styl
```diff
@@ -44,3 +26,6 @@
 
 // Theming
 build-themes-Enactors(themes);
+
+// Fonts
+build-fonts-Enactors(fonts);
```
jobara: It's good practice to end the files in an empty newline

src/framework/preferences/css/stylus/utils/Fonts.styl
```diff
@@ -0,0 +1,81 @@
...
+                :not([class*='icon']) {
+                    font-family: unquote(font) !important;
+                }
+            }
+        }
+    }
+}
+
+
+build-fonts-Enactors(textFont) {
```
jobara: `build-fonts-prefsEditor`, `build-fonts-separatedPanelPrefsEditorFrame`, `build-fonts-separatedPanelPrefsEditor` and `build-fonts-Enactors` all appear to be the same. You should be able to reuse a single function in multiple places. Also similar to our JS coding style, you should probably leave a space between the function name and the arguments block. Like:

```
build-fonts-Enactors (textFont) {}
```

src/framework/preferences/css/stylus/Enactors.styl
grrrero: Moved them all to build-fonts-prefs

src/framework/preferences/css/stylus/utils/Fonts.styl
colinbdclark: Just for the record, @jobara, I think we should be thoughtful when applying code style conventions imported from other programming languages. Stylus clearly has a very different syntax, and you'll notice that the definition of anonymous functions in Stylus is unambiguous (due to the special <code>@</code> symbol), unlike in JavaScript. If I remember correctly, this was the rationale for the original Crockford-ist use of the space in JS function definitions that are on the right side of a variable assignment.

Plus, well, I don't think spaces in function definitions are syntactically valid in Stylus, as @grrrero noted. :)

src/framework/preferences/css/stylus/Enactors.styl
```diff
@@ -44,3 +26,6 @@
 
 // Theming
 build-themes-Enactors(themes);
+
+// Fonts
+build-fonts-prefs(fonts);
```
jobara: Could you please add an empty newline at the end of the file.

jobara:
@grrrero I think it would make sense to also mention this change to the [API Changes from 2.0 to 3.0](http://docs.fluidproject.org/infusion/development/APIChangesFrom2_0To3_0.html) documentation. You can find the corresponding markdown file at https://github.com/fluid-project/infusion-docs/blob/master/src/documents/APIChangesFrom2_0To3_0.md

I think you can just add an enactors section under Preferences Framework and mention that the text style change will not affect elements with class names containing "icon". And maybe mention that this is to prevent icon fonts from being overridden, or something like that.

src/framework/preferences/css/stylus/Enactors.styl
grrrero: Done!

src/framework/preferences/css/stylus/SeparatedPanelPrefsEditorFrame.styl
```diff
@@ -211,3 +212,6 @@ html {
 
 // Theming
 build-themes-separatedPanelPrefsEditorFrame(themes);
+
+// Fonts
+build-fonts-prefs(fonts);
```
jobara: @grrrero do we need this and the themes styles in this file, it seems that the generated CSS file is always paired with PrefsEditor.css which should already include it. Similar comment for SeparatedPanelPrefsEditor.styl. Could you check if we can remove those functions from these two files?

grrrero: Yep I can confirm. Just tested them now. I'll remove them.

grrrero: @jobara If we can remove those functions from those styl files, then we can probably remove them from Themes.styl too, right?

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -1,8 +1,8 @@
 /* General Preferences Editor styles, used for all layouts */
+// Deprecated function call to contrastThemes and fontFamily styling
```
jobara: The depreciation comments should actually be in Themes.styl, also don't need to mention fontFamily. Similar for other depreciation comments throughout.

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -257,13 +256,7 @@
...
         content: "\e001";
     }
-    .fl-icon-left:before {
-        content: "\e002";
-    }
-    .fl-icon-contrast:before {
-        content: "\e005";
-    }
-    .fl-icon-line-space:before {
+    .ffl-icon-line-space:before {
```
jobara: I don't think you meant to delete these styles, and there is an extra "f" added as well.

grrrero: Eep sorry thanks for catching that

src/framework/preferences/css/stylus/utils/Fonts.styl
```diff
@@ -0,0 +1,31 @@
+// Contains utility functions for changing font family
```
jobara: Maybe this should be "Contains a utility function for changing font family"

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -1,5 +1,6 @@
 // Contains utility functions for building contrast themes
 // !important is used for all styles to ensure that Preferences Framework's enactors are able to override the pages default styles.
+// Deprecated functions to separatedPanelPrefsEditor, separatedPanelPrefsEditorFrame, prefsEditor
```
jobara: It's probably better to add a comment above each of the deprecated files, as it's more likely that it will be seen in that case.

jobara:
@grrrero I'm seeing lots of different behaviour with the contrast themes. I wonder if when you were trying git out before you forgot to run `grunt buildStylus` after making removing the theme functions.

Screenshot showing some of the issues with contrast. Notice the background of the content and the styling of the contrast options:
screen shot 2017-08-10 at 3 40 05 pm

jobara:
@grrrero the text options in the drop down should use their respective font. This may not work in all browsers. It should work at least in Firefox though.

The screen shot below was taken from the build site.
screen shot 2017-08-15 at 9 24 33 am

src/framework/preferences/css/stylus/utils/Fonts.styl
```diff
@@ -0,0 +1,31 @@
...
+}
+
+build-fonts-Enactors(textFont) {
+
+    for fontName in textFont {
+        font = textFont[fontName].fontFamily;
+        
+        {fontName},
+        {fontName} * {
+            :not([class*='icon']) {
```
jobara: I think this is missing an "&".

```
&:not([class*='icon']) {
```

That way the output would look like.

```css
.fl-font-arial:not([class*='icon']),
.fl-font-arial *:not([class*='icon']) {
    font-family: Arial !important;
}
```

jobara:
@grrrero seems that newer versions of Firefox don't support the styling in the dropdown. So don't worry about that one anymore.

jobara:
@grrrero i was just looking through the commit logs and I noticed that after the first one, the standard commit log form seems to have been missed. I think the easiest thing would be to just squash all of your commits down to a single one, and to use the proper form of commit message for that.

jobara:
Merged at 2e5bd5bf9be43ecf0f5303cf3a206b866e35ba6f

PR #843, opened 2017-08-28
FLUID-6188: Make sure that fluid.loadTestingSupport is idempotent.

amb26:
Also fix for FLUID-5920 to ensure that default global uncaught exception handler causes system exit with failure. General tidying and reorganisation of basic node tests and their entanglement with TAP reporting


the-t-in-rtf:
@amb26, is there a dev release of this?  It makes sense to test in gpii-webdriver, which is where I use the TAP support.

cindyli:
Merged at 392b2f8266725c41df1c8290b69c0e72100d0b8c

amb26:
Thanks for the merge! Dev release at 3.0.0-dev.20170830T182157Z.392b2f8

PR #844, opened 2017-09-06
FLUID-5534: UIO responsiveness and unit testing

grrrero:
Expanded unit tests, updated fluid.prefs.separatedPanel.scrollToPanel to constrain panelIndex within the range of panels, added arrows for navigation for responsive version.
https://issues.fluidproject.org/browse/FLUID-5534

Infusion-Docs PR: https://github.com/fluid-project/infusion-docs/pull/127


src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,77 @@
...
+        },
+        listeners: {
+            "onReady.windowResize": {
+                "this": window,
+                method: "addEventListener",
+                args: ["resize", "{that}.events.onSignificantDOMChange.fire"]
+            }
+        },
+        invokers: {
+            scrollToPanel: {
```
amb26: This should be a modellised component rather than using DOM coordinates as an implicit model - also the range limiting should be done within the model space using a relay. See https://github.com/fluid-project/infusion/blob/master/src/components/textfieldControl/js/TextfieldSlider.js#L64 for examples of the latter.

Rather than a "scrollToPanel" invoker, the component should expose "panelIndex" and "panelCount" as model elements, with the final line "scrollLeft" implemented as a modelListener to the former.

tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorResponsiveTests.js
```diff
@@ -0,0 +1,218 @@
...
+    };
+
+    fluid.tests.direction = {
+        LEFT: false,
+        RIGHT: true
+    };
+
+    fluid.tests.clickArrow = function (elm, direction) {
+        var keyEvent = $.Event("click");
+        keyEvent.offsetX = direction ? elm.width() : 0;
```
cindyli: It's a bit indirect and confusing to use `{LEFT: false, RIGHT: true}` to calculate `offsetX`. What do you think about something like `keyEvent.offsetX = direction === "RIGHT" ? elm.width() : 0;`?

tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorResponsiveTests.js
```diff
@@ -0,0 +1,218 @@
...
+
+        elm.trigger(keyEvent);
+    };
+
+    fluid.defaults("fluid.tests.separatedPanelResponsiveTester", {
```
cindyli: Can you test the re-rendering at window resize too, when the window changes from a small screen to a wide screen and vice versa.

src/framework/preferences/js/ArrowScrolling.js
grrrero: @amb26 Thanks for the comment! I'm not sure I totally understand though. Do you mind going through it step-by-step?

tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorResponsiveTests.js
grrrero: @cindyli Hi Cindy, that makes sense, and less confusing.

tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorResponsiveTests.js
grrrero: @cindyli Tested it, no bugs so far, and is there anything I should be looking out for?

cindyli: @grrrero, I mean to add a unit test to make sure the window size change triggers the re-rendering of panels, not only via the manual test.

src/framework/preferences/js/ArrowScrolling.js
cindyli: @grrrero, a similar example that has implemented @amb26's suggestion is [the scrolling feature in the language panel of the first discovery tool](https://github.com/GPII/first-discovery/blob/master/src/js/panels.js#L665-L667). 

The `panelIndex` model value in @amb26's comment serves the same purpose as the `selectedLang` n the example. It keeps track of the current panel index. A click on arrow keys triggers a model change on the `panelIndex` value (to the `scrollIndex` value you calculated [here](https://github.com/grrrero/infusion/blob/efe9df3a808ecba82d3858ed2004c18ebef6bfc3/src/framework/preferences/js/ArrowScrolling.js#L63)). Then the model listener on `panelIndex` will perform the scrolling.

@amb26, correct me if this explanation doesn't seem right.

jobara: @grrrero and @cindyli I think this may be similar to https://github.com/GPII/first-discovery/blob/master/src/js/nav.js

jobara: @grrrero I've made some changes to finish up the work we started the other day to modelize panelIndex. I've submit a pull request against your branch ( https://github.com/grrrero/infusion/pull/1 ). You should review those changes and merge them into your branch if you are happy with them. You will need to add more tests to SeparatedPanelPrefsEditorTests for verifying the panelIndex after scrolling and to also trigger scrolling by using the change applier to change the panelIndex value.

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,119 @@
...
+            // scrollToIndex: null, // the raw index set by translateToScroll, will be transformed to the panelIndex
+            panelIndex: 0
+        },
+        events: {
+            // beforeReset: null // should be provided by the fluid.prefs.prefsEditor grade
+        },
+        modelRelay: {
+            target: "panelIndex",
+            forward: {excludeSource: "init"},
+            singleTransform: {
```
amb26: Great improvement at modelising this component! Do put a namespace on this relay rule and the model listener which follows it in case anyone wants to override it someday

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,119 @@
...
+            target: "{that > fluid.prefs.panel}.options.listeners"
+        }]
+
+    });
+
+    fluid.prefs.arrowScrolling.calculatePanelMaxIndex = function (panels) {
+        return Math.max(0, panels.length - 1);
+    };
+
+    fluid.prefs.arrowScrolling.translateToScroll = function (that, event) {
```
amb26: Slightly better naming for this function/invoker - something like "eventToScrollIndex"?

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,119 @@
...
+                    }
+                }
+            },
+            "beforeReset.resetPanelIndex": {
+                listener: "{that}.applier.fireChangeRequest",
+                args: {path: "panelIndex", value: 0, type: "ADD", source: "reset"}
+            }
+        },
+        invokers: {
+            scrollToPanel: {
```
amb26: This invoker can be axed - the documentation that "scrollToIndex" is modelised is sufficient. Also it confusingly shares a name with the panelIndex listener

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,119 @@
...
+
+    fluid.defaults("fluid.prefs.arrowScrolling", {
+        gradeNames: ["fluid.modelComponent"],
+        selectors: {
+            // panels: "", // should be supplied by the fluid.prefs.prefsEditor grade.
+            scrollContainer: ".flc-prefsEditor-scrollContainer"
+        },
+        model: {
+            // panelMaxIndex: null, // determined by the number of panels calculated after the onPrefsEditorMarkupReady event fired
+            // scrollToIndex: null, // the raw index set by translateToScroll, will be transformed to the panelIndex
```
amb26: I think we can get by with just one model field between scrollToIndex and panelIndex, and have translateToScroll attempt to write to panelIndex directly

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,119 @@
...
+            panelIndex: 0
+        },
+        events: {
+            // beforeReset: null // should be provided by the fluid.prefs.prefsEditor grade
```
amb26: Best to uncomment this so there is a greater chance that the component can be self-sufficient (even if there are currently other reasons why it can't)

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,119 @@
...
+        },
+        modelListeners: {
+            "panelIndex": {
+                listener: "fluid.prefs.arrowScrolling.scrollToPanel",
+                args: ["{that}", "{change}.value"]
+            }
+        },
+        listeners: {
+            "onReady.windowResize": {
+                "this": window,
```
amb26: These global listeners will pile up if the user creates several of these components, especially dynamically, and the framework doesn't have enough insight into DOM listeners to get rid of them for you. You should add an onDestroy listener for this component that unregisters it using removeEventListener.

src/framework/preferences/js/ArrowScrolling.js
jobara: @amb26 will the model transformation still work in the that case?

jobara:
@amb26 the SeparatedPanelPrefsEditorResponsive-test.html test is not added to any of the all-tests.html files because the asynchrony needed to wait for the iframe to load is preventing the composite test from seeing any jqUnit tests. Do you know if there is a way to get around this?

amb26:
Yes - we have plenty of tests with asynchrony which work fine in the mix. The issue in this case will be most likely the test not beginning to execute sufficiently soon after the page load - this is a problem with QUnit's architecture which is well known. You have a 13ms window after the time QUnit itself is loaded to start queuing your tests, otherwise it will shut down assuming that there are none.

The problem is with the organisation of the test fixtures here - https://github.com/fluid-project/infusion/pull/844/files#diff-b55a73d1ab99b3bc42f1d4bc39a00744R251 - rather than waiting for the iframe to load and then starting to queue the tests afterwards, you should integrate the iframe loading into the test fixtures themselves, and start the tests immediately. This is similar to an issue that @michelled had with needing to wait for electron's app singleton to load before starting to run tests.

tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorResponsiveTests.js
```diff
@@ -139,17 +192,29 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 
     fluid.tests.triggerDOMEvent = function (elm, type) {
-        var event = new Event(type);
+        var event;
+        // In IE 11 Event is an object not a function
+        if (typeof Event === "function") {
+            event = new Event(type);
+        } else {
+            event = document.createEvent("Event");
+            event.initEvent("resize", true, true);
```
amb26: Presumably this should use the supplied type here

src/framework/preferences/js/ArrowScrolling.js
amb26: That's how it works in the original sample (the pager)

jobara:
@amb26 I believe this is ready for more review. I've also added an Infusion-Docs PR ( https://github.com/fluid-project/infusion-docs/pull/127 ).

tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorResponsiveTests.js
```diff
@@ -0,0 +1,337 @@
...
+            iframeLoaded: null
+        },
+        iframe: ".flc-iframeSetup",
+        sequence: [{
+            func: "fluid.tests.iframeSetup.load",
+            args: ["{that}"]
+        }]
+    });
+
+    fluid.tests.iframeSetup.load = function (that) {
```
amb26: Please put these definitions in a more descriptive namespace - since they are hardwired to start a pageEnhancer test. The name should at least indicate what broad area of the framework is being tests, e.g. fluid.tests.prefs or so

tests/framework-tests/preferences/html/SeparatedPanelPrefsEditorResponsive-test.html
```diff
@@ -0,0 +1,76 @@
...
+    <body>
+    <h1 id="qunit-header">SeparatedPanel Preferences Editor Responsive Test Suite</h1>
+    <h2 id="qunit-banner"></h2>
+    <div id="qunit-testrunner-toolbar"></div>
+    <h2 id="qunit-userAgent"></h2>
+    <ol id="qunit-tests"></ol>
+
+    <!-- Test HTML -->
+    <!-- <div id="qunit-fixture"> -->
+        <iframe class="flc-iframeSetup-responsiveTests"></iframe>
```
amb26: Add a comment here and/or in the corresponding testDefs to the considerable research you've done in establishing why this is necessary on all current versions of IE (including linking to the jQuery ticket that holds a hopefully related selection bug)

tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorTests.js
```diff
@@ -131,8 +136,9 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
     fluid.tests.afterHideFunc1 = function () {
```
amb26: Retrospectively we should take the opportunity to fix namespacing of lots of definitions like this. Even browser tests should be written with the model that one day all these definitions might coexist as part of a single global namespace

jobara:
@amb26 this is ready for more review

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -0,0 +1,121 @@
...
+        },
+        modelRelay: {
+            target: "panelIndex",
+            forward: {excludeSource: "init"},
+            namespace: "limitPanelIndex",
+            singleTransform: {
+                type: "fluid.transforms.limitRange",
+                input: "{that}.model.panelIndex",
+                min: 0,
+                max: "{that}.model.panelMaxIndex"
```
amb26: Now it is lovely!

jobara:
@amb26 the build process has been updated so that buildStylus is run on prepublish

PR #845, opened 2017-09-13
FLUID-6183: Added height variable to recalculate panel height.

grrrero:
MS Edge cuts off the bottom of the adjusters so the height needed to be recalculated in order to show the entire UIO panel.

### DIFF
--- a/src/framework/preferences/js/SeparatedPanelPrefsEditor.js
+++ b/src/framework/preferences/js/SeparatedPanelPrefsEditor.js
@@ -292,6 +292,9 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         }, "collapseFrame");
         separatedPanel.slidingPanel.events.afterPanelShow.addListener(function () {
             separatedPanel.iframeRenderer.iframe.show();
+
+            // FLUID-6183: Required for bug in MS EDGE that clips off the bottom of adjusters
+            var height = separatedPanel.iframeRenderer.iframe.height();
             separatedPanel.locate("reset").show();
         }, "openPanel");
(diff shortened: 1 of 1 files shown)


src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -292,6 +292,9 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         }, "collapseFrame");
         separatedPanel.slidingPanel.events.afterPanelShow.addListener(function () {
             separatedPanel.iframeRenderer.iframe.show();
+
+            // FLUID-6183: Required for bug in MS EDGE that clips off the bottom of adjusters
```
jobara: Could you please describe in the comment why it is that the code below fixes the issue in MS Edge.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -292,6 +292,9 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            var height = separatedPanel.iframeRenderer.iframe.height();
```
jobara: Assigning to a variable that is never uses causes a linting error. Does it have to be assigned to a variable to get the MS Edge fix? If so, you'll need to use an inline comment to disable the rule (see: https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments ).

jobara:
Merged at 89c0a0be14973e9fe0e0a25d4f6f0a9a48071f2a

PR #846, opened 2017-09-19
FLUID-6191: Improved diagnostic for failed indirect reference from implicit model relay

amb26:


tests/framework-tests/core/js/DataBindingTests.js
```diff
@@ -1273,6 +1273,26 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
 
+    /** FLUID-6191: Proper diagnostic on indirect model reference which fails to resolve **/
+
+    fluid.defaults("fluid.tests.fluid6191root", {
+        gradeNames: "fluid.modelComponent",
+        components: {
+            badRefHolder: {
+                type: "fluid.modelComponent",
+                options: {
+                    model: "{fluid6191root}.nonexistent.model.path"
```
cindyli: Another test case to be covered is, when the other component that the indirect model relay references to does exist but having a non-modelComponent type such as "fluid.component".

amb26: Good catch, @cindyli - test added

cindyli:
Merged at 1f32a0ca284686adb5e9c067de5d9be931fc153a

PR #848, opened 2017-09-19
FLUID-6192: Support modelRelay rule with source of ""

amb26:


src/framework/core/js/DataBinding.js
```diff
@@ -772,7 +772,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
     };
 
     fluid.parseModelRelay = function (that, mrrec, key) {
-        var parsedSource = mrrec.source ? fluid.parseValidModelReference(that, "modelRelay record member \"source\"", mrrec.source) :
+        var parsedSource = mrrec.source !== undefined ? fluid.parseValidModelReference(that, "modelRelay record member \"source\"", mrrec.source) :
```
cindyli: Shall `mrrec.source` be trimmed so it handles strings with leading or trailing spaces?

cindyli:
Shall it be documented that, when the `source` value is an empty string or not specified, the source is an equivalent of `{that}.model`?

cindyli:
Please fix the copyright or these js files.

src/framework/core/js/DataBinding.js
amb26: It is not specified to do that. In general we don't support extraneous spaces where the system expects paths or IoC references.

amb26:
For all strings p which are not IoC references, their referent is the same as "{that}.model."+p so I don't think we need to make any special case. For a source which is not specified at all, it is not the equivalent of {that}.model - in this case the relay is expected to be "auto-sourced" and any data dependencies will just be taken from the relay specification itself.

amb26:
Copyright years updated

cindyli:
Merged at 3be1824401c55d455fe21b2a1e3c4b45d08d345f

PR #849, opened 2017-09-19
FLUID-6193: Defeat corruption when distribution head is shorter-lived sibling

amb26:


cindyli:
Merged at 1fe914097336a552abeaf298e6175082e2d094dd

PR #851, opened 2017-09-21
FLUID-6194: Avoid relaying DELETEs backwards for transformations which do not have inverses

amb26:
@kaspermarkus , would you like to take a look?


cindyli:
Merged at 5dd8f82d9237174336c23e9c88c09fdebeab3003

PR #853, opened 2017-09-25
FLUID-6180: Adding persistence for adjuster selection in responsive display

jobara:
Whether scrolling with the keyboard (arrow keys, tabs), mouse, or clicking on the arrow buttons the last viewed adjuster should persist after closing and reopening the panel ( even on page reload if a cookie or other similar long term storage is used ).

https://issues.fluidproject.org/browse/FLUID-6180

This includes changes from https://github.com/fluid-project/infusion/pull/844

### DIFF
--- a/src/framework/preferences/js/ArrowScrolling.js
+++ b/src/framework/preferences/js/ArrowScrolling.js
@@ -33,7 +33,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         },
         events: {
             beforeReset: null, // should be fired by the fluid.prefs.prefsEditor grade
-            afterScroll: null
+            onScroll: null
         },
         modelRelay: {
             target: "panelIndex",
@@ -56,8 +56,16 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         },
         listeners: {
             "onReady.scrollEvent": {
-                "listener": "fluid.prefs.arrowScrolling.scrollDebounce",
-                args: ["{that}.dom.scrollContainer", "{that}.events.afterScroll.fire"]
+                "this": "{that}.dom.scrollContainer",
+                method: "scroll",
+                args: [{
+                    expander: {
+                        // Relaying the scroll event to onScroll but debounced to reduce the rate of fire.  A high rate
+                        // of fire may negatively effect performance for complex handlers.
+                        func: "fluid.debounce",
+                        args: ["{that}.events.onScroll.fire", 100]
+                    }
+                }]
             },
             "onReady.windowResize": {
                 "this": window,
@@ -84,7 +92,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                 listener: "{that}.applier.fireChangeRequest",
                 args: {path: "panelIndex", value: 0, type: "ADD", source: "reset"}
             },
-            "afterScroll.setPanelIndex": {
+            "onScroll.setPanelIndex": {
                 changePath: "panelIndex",
                 value: {
                     expander: {
@@ -146,18 +154,6 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             return a.offset - b.offset;
         });
         return panelArray[0].index;
-    };
-
-    // Based on scrollStop.js ( https://github.com/cferdinandi/scrollStop ),
-    // which is licensed under: MIT License.
-    fluid.prefs.arrowScrolling.scrollDebounce = function (elm, callback, delay) {
-        var timeoutID;
-        delay = delay || 66;
-
-        $(elm).scroll(function () {
-            window.clearTimeout(timeoutID);
-            timeoutID = setTimeout(callback, delay);
-        });
     };
 
 })(jQuery, fluid_3_0_0);
--- a/src/framework/preferences/js/ArrowScrolling.js
+++ b/src/framework/preferences/js/ArrowScrolling.js
@@ -143,17 +143,16 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
     };
 
     fluid.prefs.arrowScrolling.getClosestPanelIndex = function (panels) {
-        var panelArray = [];
-        panels.each(function (idx, panel) {
-            panelArray.push({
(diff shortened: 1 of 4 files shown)


src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -118,4 +134,30 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         }
     };
 
+    fluid.prefs.arrowScrolling.getClosesPanelIndex = function (panels) {
```
amb26: typo: Closest

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -118,4 +134,30 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+        });
+        panelArray.sort(function (a, b) {
+            return a.offset - b.offset;
+        });
+        return panelArray[0].index;
+    };
+
+    // Based on scrollStop.js ( https://github.com/cferdinandi/scrollStop ),
+    // which is licensed under: MIT License.
+    fluid.prefs.arrowScrolling.scrollDebounce = function (elm, callback, delay) {
```
amb26: It's unclear what this function achieves - it could do with some better commenting about its strategy and signature. Also, it seems like this would be something better done as part of the DataSource implementation for FLUID-6209. The specific piece of work that this pull request achieves doesn't seem to be written up in the JIRA, which just describes the requirement for persistence for the adjustor selection?

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -77,6 +83,16 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            },
+            "afterScroll.setPanelIndex": {
+                changePath: "panelIndex",
+                value: {
+                    expander: {
+                        funcName: "fluid.prefs.arrowScrolling.getClosesPanelIndex",
+                        args: "{that}.dom.panels"
+                    }
+                },
+                source: "manualScroll"
```
amb26: This source name doesn't seem too accurately descriptive - "manualScroll" might equally well describe swiping or sliding the entire panel or its scrollbar. Perhaps something like panelIndexScroll or arrowScroll?

src/framework/preferences/js/ArrowScrolling.js
amb26: Also, I think we have enough copied of this algorithm in circulation that it is probably time for it to go into the core framework - see jqUnit-node's version at https://github.com/fluid-project/node-jqunit/blob/master/lib/jqUnit-node.js#L158

src/framework/preferences/js/ArrowScrolling.js
jobara: @amb26 actually this is to handle the case where the panel was scrolled by swiping, sliding, or using the keyboard to change the scroll position. But the scroll event is also triggered when scrolling with the arrow buttons too.

src/framework/preferences/js/ArrowScrolling.js
jobara: @amb26 most of the persistence came for free with work on FLUID-5534; however, that only handled the case where scrolling happened via the arrow buttons. This PR finishes the task by addressing any type of scroll action that may have happened (e.g. swiping, sliding, using the keyboard).

jobara: I've added a [comment](https://issues.fluidproject.org/browse/FLUID-6180?focusedCommentId=37430&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-37430) to the JIRA to clarify what work this PR is covering.

jobara: I've cherry-picked the fluid.debounce out of my fluid-6102 work and added it here. Making use of that for debouncing now.

jobara:
@amb26 this PR is ready for more review. It seems that the work on FLUID-6102 will take  bit longer, so in the interest of saving time, I've cherry-picked fluid.debounce from there and used it here. I've addressed your other comments as well and added an explanation to a reply above and to a comment on the JIRA about the scope of the work in this PR in relation to the JIRA.

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -49,10 +50,23 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
         listeners: {
+            "onReady.scrollEvent": {
+                "this": "{that}.dom.scrollContainer",
+                method: "scroll",
+                args: [{
+                    expander: {
+                        // Relaying the scroll event to onScroll but debounced to reduce the rate of fire.  A high rate
+                        // of fire may negatively effect performance for complex handlers.
+                        func: "fluid.debounce",
+                        args: ["{that}.events.onScroll.fire", 100]
```
amb26: Please hoist this constant into a configuration option

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -118,4 +142,18 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         }
     };
 
+    fluid.prefs.arrowScrolling.getClosestPanelIndex = function (panels) {
+        var panelArray = [];
+        panels.each(function (idx, panel) {
```
amb26: Should be possible to make this a fluid.transform

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -118,4 +142,18 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            panelArray.push({
+                index: idx,
+                offset: Math.abs($(panel).offset().left)
+            });
+        });
+        panelArray.sort(function (a, b) {
+            return a.offset - b.offset;
+        });
+        return panelArray[0].index;
```
amb26: This is the first point we might get a failure when there are 0 panels - probably some kind of guarding is appropriate

jobara:
@amb26 I've made the changes you requested. Ready for more review.

PR #854, opened 2017-10-09
FLUID-6202: Both failing and non-failing tests plus de-Zalgoisation of fluid.fetchResources to avoid races in construct-time event binding code

amb26:


cindyli:
The test `tests/framework-tests/preferences/html/SeparatedPanelPrefsEditorResponsive-test.html` fails with this fix.

cindyli:
Please ignore my comment above. That test failed because I didn't run `grunt buildStylus`. Once stylus files are compiled, all tests pass.

cindyli:
Merged at 4612ab559ec9610b5fe22eb43c9af041ced82ab0

PR #859, opened 2017-11-15
FLUID-6225: Move registration of uncaught exception handler until after results of deduping have been determined

amb26:


the-t-in-rtf:
Thanks for figuring this out and creating a pull so quickly, @amb26.  Is there a dev release that includes this fix?  I would like to test it in my own work.

PR #861, opened 2017-11-21
FLUID-6228: Positioning radio button behind label.

jobara:
This is to keep it hidden, but also allow it scroll into view when tab
focused.

https://issues.fluidproject.org/browse/FLUID-6228

*NOTE: For Firefox and IE 11 when tab focusing to the element, it will only show the smallest amount necessary. The other browsers do a better job of scrolling the entire widget into view.*

### DIFF
--- a/src/framework/preferences/css/stylus/PrefsEditor.styl
+++ b/src/framework/preferences/css/stylus/PrefsEditor.styl
@@ -126,8 +126,8 @@
 
                 // position behind label
                 z-index: -2;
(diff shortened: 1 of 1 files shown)


jobara:
I've fixed the IE 11 issue by using rem values instead of % values for the top and left positioning.

cindyli:
Merged at 9b4fde781cdac30b4eb89fce9f7f197fafd2c97a

PR #867, opened 2017-12-21
FLUID-6209: Fix race conditions for Prefs framework

jobara:
https://issues.fluidproject.org/browse/FLUID-6209

This PR Adds:
* fluid.dataSource
* fluid.remoteModelComponent

It also updates the prefs framework's stores to use the fluid.dataSource component and the prefsEditor to be a fluid.remoteModelComponent


jobara:
@amb26 could you please review this PR to address the race conditions in the prefs framework.

src/framework/core/js/RemoteModel.js
```diff
@@ -0,0 +1,201 @@
...
+        var remoteChanges = fluid.modelPairToChanges(fetchedModel, that.model.remote, "local");
+        var localChanges = fluid.modelPairToChanges(that.model.local, that.model.remote, "local");
+        var changes = remoteChanges.concat(localChanges);
+
+        // perform model updates in a single transaction
+        var transaction = that.applier.initiate();
+        transaction.fireChangeRequest({path: "local", type: "DELETE"}); // clear old local model
+        transaction.change("local", that.model.remote); // reset local model to the base for applying changes.
+        transaction.fireChangeRequest({path: "remote", type: "DELETE"}); // clear old remote model
+        transaction.change("remote", fetchedModel); // update remote model to fetched changes.
```
amb26: Yes, yes, yes

src/framework/core/js/RemoteModel.js
```diff
@@ -0,0 +1,201 @@
...
+        fluid.fireChanges(transaction, changes); // apply changes from remote and local onto base model.
```
amb26: I don't know about you, but to me something seems interestingly and fundamentally right about this (even though there may well still be bugs in detail)

src/framework/core/js/RemoteModel.js
```diff
@@ -0,0 +1,201 @@
...
+        transaction.commit(); // submit transaction
+    };
+
+    fluid.remoteModelComponent.makeSequneceStrategy = function (payload) {
```
amb26: Typo here

src/framework/core/js/RemoteModel.js
```diff
@@ -0,0 +1,201 @@
...
+        return {
+            invokeNext: function (that) {
```
amb26: Add detailed comment here contrasting this implementation with the standard one (referencing its location in the code) and explaining why this improvement was necessary. Ideally we should just say that the standard impl has a bug/limitation and write it up as a JIRA and fix it here.

src/framework/core/js/RemoteModel.js
```diff
@@ -0,0 +1,201 @@
...
+        return sequencer;
+    };
+
+    fluid.remoteModelComponent.fireEventSequence = function (event, payload, options) {
+        var listeners = fluid.makeArray(event.sortedListeners);
+        var sequence = fluid.remoteModelComponent.makeSequence(listeners, payload, options);
+        return sequence.promise;
+    };
+
+    fluid.remoteModelComponent.fetch = function (that) {
```
amb26: Substantial comments needed on both fetch and write - ideally we should write up the indented pseudocode we worked on and polish it a bit. Perhaps some kind of tree diagram/sequence diagram?

src/framework/preferences/js/ArrowScrolling.js
```diff
@@ -30,7 +30,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         onScrollDelay: 100, // in ms, used to set the delay for debouncing the scroll event relay
         model: {
             // panelMaxIndex: null, // determined by the number of panels calculated after the onPrefsEditorMarkupReady event fired
-            panelIndex: 0
+            // panelIndex: 0 // the index of the panel to open on
```
amb26: We seem to have picked up some stray material from another pull - can we rebase this against current master?

jobara: Actually I needed to comment that line out in order for the panelIndex to be properly pulled in from the store. Otherwise this was overriding things back to 0, because of the model relay I believe.

src/framework/core/js/RemoteModel.js
jobara: I think you are referring to the use of apply. I have actually fixed this in FluidPromise as well, in this PR. https://github.com/jobara/infusion/blob/ca255c90147329a2a277c3c47efc267ac6f913f3/src/framework/core/js/FluidPromises.js#L209

The new makeSequenceStrategy was actually needed because this is a different type of sequence. Rather than passing along the result from one listener to the next, it actually passes along the original payload to each. This is because the synthetic events are treated more like typical events, but we need to know when they are all finished before resolving/rejecting the promise.

I've added a comment to try to explain the difference.

src/framework/core/js/RemoteModel.js
jobara: @amb26 I've added comments to the fetch and write functions. I've created the following diagrams that can be added to the documentation that should be written for the `fluid.remoteModelComponent`.

fetch diagram
write diagram

jobara:
@amb26 I've started writing up the docs for the changes in this PR. So far I have written documentation for the two new files: [DataSource](https://github.com/jobara/infusion-docs/blob/FLUID-6209/src/documents/DataSourceAPI.md) and [remoteModelComponent](https://github.com/jobara/infusion-docs/blob/FLUID-6209/src/documents/RemoteModelAPI.md).

src/framework/core/js/RemoteModel.js
```diff
@@ -0,0 +1,237 @@
...
+You may obtain a copy of the ECL 2.0 License and BSD License at
+https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
+*/
+
+var fluid_3_0_0 = fluid_3_0_0 || {};
+
+(function ($, fluid) {
+    "use strict";
+
+    fluid.defaults("fluid.remoteModelComponent", {
```
amb26: Could do with some substantial comments here explaining what is a "remote model" given clearly this component itself is local : P Perhaps with reference to some diagrams/workflow

src/framework/core/js/RemoteModel.js
```diff
@@ -0,0 +1,237 @@
...
+        gradeNames: ["fluid.modelComponent"],
+        events: {
+            afterFetch: null,
```
amb26: Having separate events for errors is ok, but I think separate onWrite/afterWrite and onFetch/afterFetch errors is dubious, because of the standard problem of associating together the particular firings of related events - that is, there is no clear way for a user to identify the *particular* firing of an afterFetch event that corresponds to a particular onFetch. For this reason I think we should combine these events and just have a more elaborate namespace structure within the fewer events that allows users to schedule things at a time of interest.

src/framework/preferences/js/ArrowScrolling.js
amb26: But there's no connection between the work in this pull and the ArrowScrolling work, right?

src/framework/preferences/js/FullPreviewPrefsEditor.js
```diff
@@ -41,7 +41,9 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                         }
                     },
                     listeners: {
-                        "onReady.boilOnPrefsEditorReady": "{fullPreview}.events.onPrefsEditorReady"
+                        "onReady.boil": {
```
amb26: Similarly this looks like stray material from another pull

src/framework/preferences/js/ArrowScrolling.js
amb26: Oh sorry, I see that there is. We should try to put the default value back in, and prevent it from propagating out with an excludeSource: "init" directive on the relay

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
```diff
@@ -143,22 +143,32 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                 options: {
                     gradeNames: ["fluid.prefs.uiEnhancerRelay", "fluid.prefs.arrowScrolling"],
                     // ensure that model and applier are available to users at top level
-                    model: "{separatedPanel}.model",
+                    model: {
+                        preferences: "{separatedPanel}.model.preferences",
+                        panelIndex: "{separatedPanel}.model.panelIndex",
+                        panelMaxIndex: "{separatedPanel}.model.panelMaxIndex",
+                        local: {
+                            panelIndex: "{that}.model.panelIndex"
```
amb26: What is the distinction between the local panelIndex and the top-level one?

src/framework/preferences/js/ArrowScrolling.js
jobara: @amb26 I think I'm remembering more of the problem. If there is a default value, it will be applied instead of the value from the cookie (or other datastore). If exclude init, then you can't pass in a  panelIndex through the component settings to have it default to a different panel.

Although now that I'm testing this out by uncommenting the default value for panelIndex, I'm getting a strange error at https://github.com/jobara/infusion/blob/ca255c90147329a2a277c3c47efc267ac6f913f3/src/framework/preferences/js/PrefsEditor.js#L352 It seems that the changeMap has the value "ADD". This results in the following error "DataBinding.js:496 Uncaught TypeError: Cannot read property '0' of undefined"

amb26: That's actually an ok value for a ChangeMap and the prefsEditor code has a bug. A ChangeMap isn't a flat map, it's an object isomorphic to the model with values "ADD" and "DELETE" at the locations where changes have occurred

src/framework/core/js/RemoteModel.js
jobara: I've started writing documentation for the [remoteModelComponent](https://github.com/jobara/infusion-docs/blob/FLUID-6209/src/documents/RemoteModelAPI.md). I'll add a link, but it won't be live yet as the documentation needs to be merged in still.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
jobara: @amb26 this is the setup for what is tracked for saving between the local and remote models, as part of the fluid.remoteModelComponent.

src/framework/preferences/js/FullPreviewPrefsEditor.js
jobara: This is so that it will properly override the "onReady.boil" listener from the prefsEditorLoader base grade. https://github.com/jobara/infusion/blob/6c4381917a85641beb6195a50006976f7ebad733/src/framework/preferences/js/PrefsEditor.js#L47

src/framework/core/js/RemoteModel.js
jobara: @amb26 could you expand on what you mean by "have a more elaborate namespace structure within the fewer events that allows users to schedule things at a time of interest"

src/framework/preferences/js/ArrowScrolling.js
jobara: @amb26 I've tried to exclude init for the panelIndex relay, but it seems to cause the issue I mentioned above, where it ignores values for the panelIndex in the components defaults.

src/framework/preferences/js/FullPreviewPrefsEditor.js
amb26: OK, I see

src/framework/preferences/js/ArrowScrolling.js
amb26: OK, this is a pretty interesting issue - do you think you could write up the details of how this situation works as a JIRA, with a title something like "model relay system cannot distinguish between different sources of initial values"

jobara: @amb26 I've filed a JIRA https://issues.fluidproject.org/browse/FLUID-6249 Let me know if you'd like any more information provided.

src/framework/core/js/RemoteModel.js
jobara: I spoke with @amb26 in the IRC channel about this ( https://botbot.me/freenode/fluid-work/2018-02-26/?msg=97241487&page=1 ). The suggestion is to try to move the items from the afterWrite and afterFetch to be part of the onWrite and onFetch sequence. However, the current implementation is working, so will timebox the exploration of adjusting the implementation.

jobara: I did some exploration into moving the afterFetch and afterWrite workflows into onFetch and onWrite respectively. The main issue preventing this is that the fireEventSequences used in the remote model, passes along the same payload to each listener ( https://github.com/jobara/infusion/blob/6c4381917a85641beb6195a50006976f7ebad733/src/framework/core/js/RemoteModel.js#L108-L126 ). This was done so that the listeners could be treated more like real listeners and not have to return back the payload at each step to the next listener in the chain. However, the onWrite and onFetch events don't have a payload. The payload was passed along in the afterWrite and afterFetch events.

jobara:
@amb26 I've reverted the change to the panelIndex model value, it is commented out again and documented under [FLUID-6249](https://issues.fluidproject.org/browse/FLUID-6249). I've also updated with the master and fixed merge conflicts. This is ready for more review.

src/framework/preferences/js/SeparatedPanelPrefsEditor.js
amb26: OK - it would be useful to add a comment on line 150 reminding the user that this structuring arises through the use of fluid.remoteModelComponent - and that what it is for is arranging for this to be persisted and synchronised there

jobara: @amb26 that makes sense. I've added a comment about it.

jobara:
@amb26 ready for more review.

PR #869, opened 2018-01-11
FLUID-6204: Letter-Spacing preference

jobara:
Added an adjuster and enactor for modifying the letter-spacing of a website. Also factored out a textfield stepper grade to make it easier to create new adjusters of this type.

https://issues.fluidproject.org/browse/FLUID-6204


jobara:
@cindyli would you mind to review this PR for me. You may wish to run the adjustments by one of the designers (maybe caren, dana, or lisa) to see if they think it looks okay.

jobara:
@cindyli just remembered I also need to update the icon used in the adjuster. That will need to be changed before this can be merged.

src/framework/preferences/js/LetterSpaceSchemas.js
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright 2015 OCAD University
```
cindyli: Fix the copyright. This comment also applies to other files.

src/framework/preferences/js/LetterSpaceSchemas.js
```diff
@@ -0,0 +1,71 @@
...
+https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
+*/
+
+var fluid_3_0_0 = fluid_3_0_0 || {};
+
+(function (fluid) {
+    "use strict";
+
+    /*******************************************************************************
+    * Starter auxiliary schema grade
```
cindyli: Please fix the incorrect comment. The same applies to comments down below for components.

src/framework/preferences/js/LetterSpaceSchemas.js
```diff
@@ -0,0 +1,71 @@
...
+    *
+    * Contains the settings for 7 preferences: text size, line space, text font,
+    * contrast, table of contents, inputs larger and emphasize links
+    *******************************************************************************/
+
+    // Fine-tune the starter aux schema and add speak panel
+    fluid.defaults("fluid.prefs.auxSchema.letterSpace", {
+        gradeNames: ["fluid.prefs.auxSchema"],
+        auxiliarySchema: {
+            "namespace": "fluid.prefs.constructed",
```
cindyli: Line 28-34 are common options at creating a new aux schema for a preference. Would be easier to reuse by extracting them into a base grade.

tests/framework-tests/preferences/html/LetterSpaceEnactor-test.html
```diff
@@ -0,0 +1,59 @@
...
+    </head>
+
+    <body style="font-size: 1.5rem;">
+        <h1 id="qunit-header">Letter Space Enactor Test Suite</h1>
+        <h2 id="qunit-banner"></h2>
+        <div id="qunit-testrunner-toolbar"></div>
+        <h2 id="qunit-userAgent"></h2>
+        <ol id="qunit-tests"></ol>
+        <p class="flc-letterSpace" style="display: none; font-size: 16px">
+            Reading text from DOM
```
cindyli: This line can be removed?

src/framework/preferences/js/LetterSpaceEnactor.js
```diff
@@ -0,0 +1,78 @@
...
+                        "input": "{that}.model.value"
+                    }
+                }
+            }
+
+        }
+    });
+
+    fluid.prefs.enactor.letterSpace.set = function (that, units) {
+        var targetSize = units  ? units + "em" : "normal";
```
cindyli: What if `letter-spacing` css has already been set on the html? Shall this function fetch it and performs an multiplication on the original value instead of a straight overwrite?

tests/framework-tests/preferences/js/LetterSpaceEnactorTests.js
```diff
@@ -0,0 +1,100 @@
...
+            }]
+        }]
+    });
+
+    fluid.tests.letterSpaceTester.assertLetterSpace = function (that, expectedModel) {
+        var pxVal = expectedModel.unit * 16; // convert from em to px
+        var expectedLetterSpace = pxVal ? pxVal + "px" : "0";
+        jqUnit.assertDeepEq("The model should be set correctly", expectedModel, that.model);
+        jqUnit.assertEquals("The letter-spacing css style should be set to " + expectedLetterSpace, expectedLetterSpace, that.root.css("letter-spacing"));
+        jqUnit.assertEquals("The letter-spacing of the content is set to " + expectedLetterSpace, expectedLetterSpace, that.container.css("letter-spacing"));
```
cindyli: This assertion assumes the container doesn't have `letter-spacing` css being set, which is not always true.

tests/framework-tests/preferences/js/LetterSpaceEnactorTests.js
```diff
@@ -0,0 +1,100 @@
...
+                container: ".flc-letterSpace",
+                createOnEvent: "{letterSpaceTester}.events.onTestCaseStart"
+            },
+            letterSpaceTester: {
+                type: "fluid.tests.letterSpaceTester"
+            }
+        }
+    });
+
+    fluid.defaults("fluid.tests.letterSpaceTester", {
```
cindyli: This test should also test variants of the combination of `letter-spacing` being set or not set on the root html container and the letter space container.

src/framework/preferences/js/LetterSpaceEnactor.js
```diff
@@ -0,0 +1,78 @@
...
+
+(function ($, fluid) {
+    "use strict";
+
+    /*******************************************************************************
+     * letterSpace
+     *
+     * Sets the letter space on the container to the number of units to increase
+     * the letter space by. If a negative number is provided, the space between
+     * characters will decrease. Setting the size to 0 will use the default letter
```
cindyli: It seems using the default letter space is when the size (`value`) is set to 1, instead of 0.

src/framework/preferences/js/LetterSpaceEnactor.js
jobara: @cindyli and spoke about this today after I had performed some experimentation. So the situation is a bit trickier for letter spacing. There are a couple of factors 1) we don't really know the base spacing (kerning, tracking) used within a font. The letter-spacing css property actually just adds space to whatever this original value is. Meaning we don't ever do true multiplication like in font size adjustments. 2) If an original CSS value was set to say 0.2em, and user increases the letter-spacing preference in UIO by 0.5. If we do a multiplication of these values, the size would actually become 0.1em, making the space smaller. Due to all of these factors we decided to add the original value to the UIO adjustment. That is is 0.2 + 0.5 resulting in 0.7em value being set.

jobara:
@cindyli this is ready for more review. I've also included a fix for FLUID-6257.

src/framework/preferences/js/LetterSpaceEnactor.js
```diff
@@ -0,0 +1,99 @@
...
+
+(function ($, fluid) {
+    "use strict";
+
+    /*******************************************************************************
+     * letterSpace
+     *
+     * Sets the letter space on the container to the number of units to increase
+     * the letter space by. If a negative number is provided, the space between
+     * characters will decrease. Setting the value to 1, unit to 0,  will use the
```
cindyli: "Setting the value to 1, unit to 0" -> "Setting the value to 1 or unit to 0". Is it missing "or", which makes it sound like both model paths need to be set.

jobara:
@cindyli I've updated the comment. Ready for more review.

cindyli:
Below is how the header and the breadcrumb navigation of `demos/prefsFramework/` looks like when letter spacing is set to 1.6. During the increase, the indentations and words in these 2 section move around and overlap. Some words also become unreadable. I wonder if it's possible to improve the css of this demo to make it look nicer with larger letter spacing?

screen shot 2018-03-12 at 11 14 39 am

cindyli:
The main body of the overview panel is not scrollable, increasing letter spacing squeezes its bottom out of the screen so that the link and the close button at the bottom become unreachable (inaccessible via mouse).

screen shot 2018-03-12 at 11 28 18 am

jobara:
@cindyli regarding the prefs framework demo, that whole demo needs to be replaced/updated to work with the responsive view. https://issues.fluidproject.org/browse/FLUID-6117 

The CSS and markup in that demo are not in a particularly easy to work with state unfortunately. I'd prefer to fix this when the demo is updated. What do you think?

jobara:
@cindyli for the overview panel, I've added `overflow: auto` to the footer. This will make the footer scrollable for cases where it goes off screen. This is a temporary fix. I talked to @jhung about this and have filed a new JIRA to redesign the overview panel in general. https://issues.fluidproject.org/browse/FLUID-6261

cindyli:
Merged at e4d2974d22938cc20a2157f5fd687486fec15024

PR #870, opened 2018-01-17
FLUID-6237: Adding code coverage artifacts to npmignore

jobara:
https://issues.fluidproject.org/browse/FLUID-6237

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,12 +6,17 @@
 .vagrant
 report.tap
 build/
+coverage/
 demos/
 examples/
+instrumented/
 products/
+reports/
 tests/component-tests/
+tests/configs/
 tests/framework-tests/
 tests/manual-tests/
 tests/node-tests/
 tests/test-core/testTests/
 tests/all-tests.html
+tests/testem.js


jobara:
@amb26 and @the-t-in-rtf would you like to take a look at this PR?

.npmignore
```diff
@@ -6,12 +6,17 @@
...
 products/
+reports/
 tests/component-tests/
+tests/configs/
 tests/framework-tests/
 tests/manual-tests/
 tests/node-tests/
 tests/test-core/testTests/
 tests/all-tests.html
+tests/testem.js
```
the-t-in-rtf: This is not test output, but a crucial file required to run the browser tests.

.npmignore
```diff
@@ -6,12 +6,17 @@
...
 report.tap
 build/
+coverage/
 demos/
 examples/
+instrumented/
 products/
+reports/
 tests/component-tests/
+tests/configs/
```
the-t-in-rtf: I added this in the last pull and it absolutely should not be excluded.  It includes the configuration for the test instrumentation.

the-t-in-rtf:
Happy to help, @jobara.  I commented on two excludes that should be left out, the rest looks fine to me.

.npmignore
the-t-in-rtf: It replaced the previous testem.json file.

jobara: @the-t-in-rtf thanks for looking over this. The ignores here and the previous one you mentioned for the configs are only for the npm package. I figure that testing should have been completed before publishing and the only things we typically publish from the test directory are the infrastructure pieces that are needed for others to write tests. Also, these files haven't been added to the gitignore so they will still show up in the git repo.

jobara:
@the-t-in-rtf thanks for looking over this. The ignores for `configs` and `testem.js` are only for the npm package. I figure that testing should have been completed before publishing and the only things we typically publish from the test directory are the infrastructure pieces that are needed for others to write tests. 

Also, these files/directories haven't been added to the gitignore so they will still show up in the git repo.

the-t-in-rtf:
Sorry, just saw the bit between the dot and ignore.  I have seen other packages that distribute enough to run their tests, which seems a good idea to me, as it helps in tracking down environment-specific problems.  So, why do we not include our tests in the package?

the-t-in-rtf:
I guess it's a space concern, and that we expect people who get to that point to check out the repo?

jobara:
@the-t-in-rtf good question. I don't know the exact reason, but I'd guess it is partially because our builds of infusion have excluded the tests, and the npm package was made in a similar fashion. To extend my guesses, I believe that this has to do with package size, availability of tests in the repo, and generally just removing extra bits that may not be used.

Regarding the npm package, here are my thoughts on including the tests.
- The size issues matters less in npm distribution than our zip packages for web apps
- Due to the dev-dependencies, I don't think the code coverage can be run after an npm install. The integrator would have to also pull down Infusion's dev-dependencies in another step.
- If we don't ignore the code coverage related parts, we need to stop ignoring all parts of the tests directory.

the-t-in-rtf:
@jobara, good point about the coverage and dev dependencies.  Happy for it to work the way you've set it up, I was mainly curious.

PR #871, opened 2018-01-22
FLUID-6223: Add support for EL paths to `fluid.stringTemplate`.

the-t-in-rtf:
See [FLUID-6223](https://issues.fluidproject.org/browse/FLUID-6223) for details.

### DIFF
--- a/src/framework/core/js/Fluid.js
+++ b/src/framework/core/js/Fluid.js
@@ -2862,21 +2862,82 @@ var fluid = fluid || fluid_3_0_0;
     };
 
     /**
-     * Simple string template system.
-     * Takes a template string containing tokens in the form of "%value".
-     * Returns a new string with the tokens replaced by the specified values.
-     * Keys and values can be of any data type that can be coerced into a string.
      *
-     * @param {String}    template    a string (can be HTML) that contains tokens embedded into it
-     * @param {object}    values      a collection of token keys and values
+     * Take an original object and represent it using top-level sub-elements whose keys are EL Paths.  For example,
+     * `originalObject` might look like:
+     *
+     * ```
+     * {
+     *   deep: {
+     *     path: {
+     *       value: "foo",
+     *       emptyObject: {},
+     *       array: [ "peas", "porridge", "hot"]
+     *     }
+     *   }
+     * }
+     * ```
+     *
+     * Calling `fluid.flattenObjectKeys` on this would result in a new object that looks like:
+     *
+     * ```
+     * {
+     *   "deep": "[object Object]",
+     *   "deep.path": "[object Object]",
+     *   "deep.path.value": "foo",
+     *   "deep.path.array": "peas,porridge,hot",
+     *   "deep.path.array.0": "peas",
+     *   "deep.path.array.1": "porridge",
+     *   "deep.path.array.2": "hot"
+     * }
+     * ```
+     *
+     * This function preserves the previous functionality of displaying an entire object using its `toString` function,
+     * which is why many of the paths above resolve to "[object Object]".
+     *
+     * This function is an unsupported non-API function that is used in by `fluid.stringTemplate` (see below).
+     *
+     * @param `originalObject` `{Object}` - An object.
+     * @returns `{Object}` - A representation of the original object that only contains top-level sub-elements whose keys are EL Paths.
+     *
+     */
+    // unsupported, non-API function
+    fluid.flattenObjectPaths = function (originalObject) {
+        var flattenedObject = {};
+        fluid.each(originalObject, function (value, key) {
+            if (typeof value === "object") {
+                var flattenedSubObject = fluid.flattenObjectPaths(value);
+                fluid.each(flattenedSubObject, function (subValue, subKey) {
+                    flattenedObject[key + "." + subKey] = subValue;
+                });
+                if (typeof fluid.get(value, "toString") === "function") {
+                    flattenedObject[key] = value.toString();
+                }
+            }
+            else {
+                flattenedObject[key] = value;
+            }
+        });
+        return flattenedObject;
+    };
+
+    /**
+     * Simple string template system.  Takes a template string containing tokens in the form of "%value" or
+     * "%deep.path.to.value".  Returns a new string with the tokens replaced by the specified values.  Keys and values
+     * can be of any data type that can be coerced into a string.
+     *
+     * @param `{String}` `template` - A string (can be HTML) that contains tokens embedded into it.
+     * @param `{Object}`  `values` - A collection of token keys and values.
+     * @returns `{String}` - A string whose tokens have been replaced with values.
      */
     fluid.stringTemplate = function (template, values) {
-        var keys = fluid.keys(values);
+        var flattenedValues = fluid.flattenObjectPaths(values);
+        var keys = fluid.keys(flattenedValues);
         keys = keys.sort(fluid.compareStringLength());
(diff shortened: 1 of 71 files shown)


src/framework/core/js/Fluid.js
```diff
@@ -2851,32 +2851,88 @@ var fluid = fluid || fluid_3_0_0;
...
+     *
+     */
+    // unsupported, non-API function
+    fluid.flattenObjectPaths = function (originalObject) {
+        var flattenedObject = {};
+        fluid.each(originalObject, function (value, key) {
+            if (typeof value === "object") {
+                var flattenedSubObject = fluid.flattenObjectPaths(value);
+                fluid.each(flattenedSubObject, function (subValue, subKey) {
+                    flattenedObject[key + "." + subKey] = subValue;
```
amb26: This function should expose configuration which allows the user the option of using the `fluid.model.composePath" algorithm https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/DataBinding.js#L183 (or some other, e.g. the JSON pointer algorithm) as well as this default which simply composes path segments as strings. We may consider this out of scope for now in which case this can just be written up as a JIRA.

src/framework/core/js/Fluid.js
```diff
@@ -2851,32 +2851,88 @@ var fluid = fluid || fluid_3_0_0;
...
+     * }
+     * ```
+     *
+     * This function preserves the previous functionality of displaying an entire object using its `toString` function,
+     * which is why many of the paths above resolve to "[object Object]".
+     *
+     * This function is an unsupported non-API function that is used in by `fluid.stringTemplate` (see below).
+     *
+     * @param `originalObject` `{Object}` - An object.
+     * @returns `{Object}` - A representation of the original object that only contains top-level sub-elements whose keys are EL Paths.
```
amb26: We should try to be consistent with these doc comments in case anyone does write a parser for them. i) I think the general convention is for @return rather than @returns, and ii) it doesn't look like we escape these with backticks

src/framework/core/js/Fluid.js
```diff
@@ -2851,32 +2851,88 @@ var fluid = fluid || fluid_3_0_0;
...
         keys = keys.sort(fluid.compareStringLength());
         for (var i = 0; i < keys.length; ++i) {
             var key = keys[i];
-            var re = fluid.stringToRegExp("%" + key, "g");
-            template = template.replace(re, values[key]);
+            var templatePlaceholder = "%" + key;
+            var replacementValue = flattenedValues[key];
+
+            var indexOfPlaceHolder = -1;
+            while  ((indexOfPlaceHolder = template.indexOf(templatePlaceholder)) !== -1) {
```
amb26: Stray sapce here

src/framework/core/js/MessageResolver.js
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     *
+     * Look up the first matching message template, starting with the current grade and working up through its parents.
+     * Returns both the template for the message and the function used to resolve the localised value.  By default
+     * the resolve function is `fluid.stringTemplate`, and the template returned uses its syntax.
+     *
+     * @param `that` `{Object}` - The component itself.
+     * @param `messagecodes` `{Array}` - One or more message codes to look up templates for.
+     * @returns `{Object}` - An object that contains`template` and `resolveFunc` members (see above).
```
amb26: See doc comment comments above

src/framework/core/js/Fluid.js
the-t-in-rtf: I guess I got the idea that it was OK by omission, i.e. comments [like this one](https://github.com/GPII/gpii-diff/pull/1#pullrequestreview-75075782) that overlooked the backticks.  It's a shame, as I finally trained myself to use them consistently, but I agree, prior art in Infusion doesn't use it (or `returns`).  Will fix shortly.

the-t-in-rtf: In reflecting on it further, I think I picked up the habit when adapting JSDocs into markdown.

the-t-in-rtf: Agree on the goal and the specifics.  I notice a few other inconsistencies within the package, for example, some JSDocs use `@param variableName {Type}`, others use `@param {Type} variableName`.  I propose standardising on the former.

amb26: Yes, that seems good - feel free to normalise any examples of the other that you find

the-t-in-rtf: Looks like we can at least enforce the consistent use of `@returns`:  https://eslint.org/docs/rules/valid-jsdoc#options

If you're game, I can submit a PR against our comment ESLint rules to add that.

src/framework/core/js/Fluid.js
the-t-in-rtf: Wow, good eye!

src/framework/core/js/Fluid.js
the-t-in-rtf: Anyway, fixed throughout.

src/framework/core/js/MessageResolver.js
the-t-in-rtf: Done.

src/framework/core/js/Fluid.js
the-t-in-rtf: I wrote up a ticket for future work in this area:  https://issues.fluidproject.org/browse/FLUID-6243

the-t-in-rtf:
@amb26, I have addressed this round of feedback, this should be ready for review again.

the-t-in-rtf:
@amb26  and I discussed a few more oddities in the JSDocs via chat that have now been addressed.  This is ready for review again.

the-t-in-rtf:
@amb26, I need you to cut a dev release of this so I can exercise it in my work with the i18n spike against gpii-handlebars.

tests/framework-tests/core/js/FluidJSTests.js
```diff
@@ -391,6 +391,61 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        jqUnit.assertDeepEq("A top-level array should be flattened correctly.", { 0: "monkey", 1: "fighting", 2: "snakes"}, output);
+    });
+
+    jqUnit.test("flattenObjectPaths: deep array", function () {
+        var originalObject = { "plane": ["monday", "to", "friday"]};
+        var output = fluid.flattenObjectPaths(originalObject);
+        jqUnit.assertDeepEq("A deep array should be flattened correctly.", { "plane": "monday,to,friday", "plane.0": "monday", "plane.1": "to", "plane.2": "friday"}, output);
+    });
+
+    jqUnit.test("flattenObjectPaths: deep empty objects", function () {
```
amb26: We should add a test to verify that the code doesn't explode with a `null` value as a member. I think, by eye, that it won't, but because it features a typeof === "object" check there is the suggestion of a risk

tests/framework-tests/core/js/FluidJSTests.js
```diff
@@ -391,6 +391,61 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+                    value: "foo"
+                }
+            },
+            otherPath: {
+                to: {
+                    otherValue: "bar"
+                }
+            }
+        };
+        var output         = fluid.flattenObjectPaths(originalObject);
```
amb26: Stray spaces of that peculiar nature

PR #876, opened 2018-02-20
FLUID-5633: Fix and test case for failure to deregister IoC Testing framework listener that was registered via IoCSS

amb26:


tests/test-core/testTests/js/IoCTestingTests.js
```diff
@@ -307,14 +317,20 @@ fluid.defaults("fluid.tests.fluid5559Tree", {
...
                         }, {
-                            func: "fluid.tests.onTestCaseStart.assertOnce",
+                            func: "fluid.tests.onTestCaseStart.assertValue",
                             args: "{targetTree}"
+                        }, { // Try firing the event again in order to test FLUID-5633
+                            func: "{fluid5559Tree}.targetTree.events.hearIt.fire",
+                            args: 5
+                        }, {
+                            event: "{fluid5559Tree targetTree}.events.hearIt",
+                            listener: "fluid.tests.onTestCaseStart.assertValue"
```
cindyli: Shall the value be verified to ensure that it receives the argument from the second firing of the `hearIt` event instead of the residue of the first registered listener on this event?

cindyli:
Merged at 29c3e677e6405b0e288075cdf94d8936905d85af

PR #879, opened 2018-03-07
FLUID-6256: Harden browser test instrumentation.

the-t-in-rtf:
See [FLUID-6256](https://issues.fluidproject.org/browse/FLUID-6256) for details.

cc: @amb26


the-t-in-rtf:
I see a caching test failure unique to CI, I saw nothing like that locally or in Vagrant.  Anyone seen this before?

cc: @amb26

the-t-in-rtf:
Just confirmed that I don't see the caching error in Vagrant locally.

the-t-in-rtf:
Testing this further this morning.  I do see cache failures in local Vagrant instances, especially if I adjusted the setTimeout calls upwards by 150ms or so.  I converted the previously hard-coded values of 100 (ms) to a single named variable and adjusted it downwards to 75 (ms).  Testing now in CI.

the-t-in-rtf:
@amb26 and I have been chatting about this, his suggestion was to adjust a bunch of the timings upward, including the new shared value for the timeouts.  Trying that now.

the-t-in-rtf:
Per longer discussions with @amb26 I committed his best guess at reasonable timings, the cache test failures are gone.

tests/testem.js
```diff
@@ -6,41 +6,28 @@ fluid.setLogging(true);
...
+    },
+    sourceDirs: {
+        src: "%infusion/src"
+    },
+    contentDirs: {
+        tests:   "%infusion/tests"
+    },
     testemOptions: {
-        skip: "PhantomJS,Opera,Safari",
+        skip: "PhantomJS,Opera,Safari,IE",
```
amb26: We can't skip IE here since the UI tested here really will be delivered to end users

the-t-in-rtf: Done. Once the Vagrant CI plugin is a bit more mature, I think it's fair to suggest that we migrate to testing infusion against a Windows VM.

amb26:
This runs ok for me but produces a batch of errors as follows:
```
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\jquery.standalone.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\Fluid.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\FluidDebugging.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\FluidPromises.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\FluidIoC.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\FluidRequests.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\DataBinding.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\JavaProperties.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\MessageResolver.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\enhancement\js\ContextAwareness.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\ModelTransformation.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\core\js\ModelTransformationTransforms.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\lib\fastXmlPull\js\fastXmlPull.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\renderer\js\fluidParser.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\renderer\js\fluidRenderer.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\framework\renderer\js\RendererUtilities.js ; return original code
content.match is not a function
Transformation error for E:\source\gits\infusion-master\src\module\module.js ; return original code
content.match is not a function
```
Also, the coverage data is quite incomplete. For example, it claims that /reports/components/pager/js/Pager.js.html line 117 is uncovered, but this function has dedicated tests.

the-t-in-rtf:
@amb26, I think we might be seeing different results, the coverage report I get only reports that particular `if` branches in that neighbourhood are not exercised.  See a sample created using Vagrant here:

https://the-t-in-rtf.github.io/coverage-reports/infusion/20180308/components/pager/js/Pager.js.html

Just to confirm, you saw the warnings and the incomplete report when running `npm test` in Windows 7?  If so, I can try firing up a Windows 10 VM for a bit and look around, the different results would make sense if (for example) the warnings you see result in failing over to copying rather than instrumenting particular code.

amb26:
Yes, that's interesting. For me, most of the functions in "Pager" showed as completely uncovered rather than showing only a few holes as in your version. Yes, the incomplete report was from my bare metal machine, I'll try reproducing in a vagrant box now.

the-t-in-rtf:
@amb26, I just generated a report within a Windows 10 Vagrant box and confirmed that the coverage (at least where I've reviewed) matches what I generated in a linux VM.

the-t-in-rtf:
I didn't see any warnings like those you saw.

amb26:
Running this in my "standard Windows 10 vagrant VM" gave this:

````
not ok 141 Chrome - error
    ---
        message: >
            Error: Browser exited unexpectedly

        Log: |
            { type: 'error', text: 'Error: Browser exited unexpectedly' }
    ...
03:18:10.456:  Stopping express...
03:18:10.487:  Express stopped...
not ok 142 Chrome 64.0 - Global error: Uncaught TypeError: Cannot read property 'moduleStart' of undefined at http://localhost:7357/4680/tests/lib/qunit/addons/composite/qunit-composite.js, line 59
03:18:26.659:  Promise chain terminated by promise rejection. Firing testem callback.
````

amb26:
I've merged this since it seems to  be working well enough - however the "content.match is not a function" errors are consistent for me, even within the Windows 10 VM. They occur just before the run of the node-based tests in basic-node-tests.js . 
I notice that the file src/module/module.js is excluded from the coverage report. I'm not clear why this is or what I could do to remedy it.

PR #882, opened 2018-03-09
Fluid 5057: Replace separated panel background image with CSS

shadxx7:
I have removed the background texture and changed the bg color to #bcb5ad which looks very similar and hence is a worthy replacement. - https://issues.fluidproject.org/browse/FLUID-5057


jobara:
Merged at ea2d67ec6e9ed17608fcd50fe7705ed5db275f4d

PR #884, opened 2018-03-14
NOJIRA: Adds more timeouts to CI stages

avtar:
These timeouts take into account how long Infusion CI jobs currently take. They should avoid scenarios like the one experienced today where [stalled tests prevented other jobs from running](https://buildkite.com/fluid-project/fluid-infusion/builds/121).


jobara:
Merged at 8c6a109b155f615f43543640d99d020040255102

PR #888, opened 2018-03-26
FLUID-6265 - Add coverage and reports directories as build artifacts

gtirloni:


gtirloni:
The way Buildkite works, we would have to start using Amazon S3 ourselves to publish these files. I'm checking what's available in the new Buildkite-agent v3.0.0 in the form of plugins to support publishing to our own servers instead.

gtirloni:
No solution has been arrived at and this will likely change due to FLUID-6282. Closing.

PR #889, opened 2018-03-28
FLUID-6266 - Remove extra CI steps, add linting and rename stages

gtirloni:
The VM provisioning phase already [installs dependencies and builds infusion](https://github.com/fluid-project/infusion/blob/master/provisioning/vars.yml#L15-L17), so this PR removes those steps from the CI configuration.

Additionally, I'm proposing to rename the jobs:
* Create VM -> Build
* Run tests -> Test
* Destroy VM -> Cleanup

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,25 +2,28 @@
 # is configured using the Buildkite web UI and has a "Run this CI job" label. It is not included 
 # in this YAML configuration to prevent it from being accidentally removed as part of a PR change.
 steps:
-  - name: "Create VM"
-    command: "DISPLAY=:0 vagrant up --provider virtualbox && npm install"
+  - name: "Build"
+    command: "DISPLAY=:0 vagrant up --provider virtualbox"
     timeout_in_minutes: 20
 
   # Wait and make sure the VM was successfully created before proceeding. Otherwise the remaining steps will not run.
   - wait
 
-  - name: "Install dependencies"
-    command: "vagrant ssh -c 'cd /home/vagrant/sync; npm install'"
-    timeout_in_minutes: 20
+  - name: "Code Linter"
+    command: "vagrant ssh -c 'cd /home/vagrant/sync/; $(npm bin)/grunt lint'"
+    timeout_in_minutes: 2
 
-  - name: "Build Infusion"
-    command: "vagrant ssh -c 'cd /home/vagrant/sync; grunt clean stylus modulefiles:all pathMap:all copy:all copy:necessities uglify:all concat:all compress:all'"
-    timeout_in_minutes: 20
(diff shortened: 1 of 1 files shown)


gtirloni:
I've have made some more changes:
* Browser and node tests are separate steps
* Added code linter step
* Made last step (Cleanup) run even if previous steps failed (so the VM gets destroyed)

colinbdclark:
@jobara, any chance you have some time to take a look at this one?

.buildkite/pipeline.yml
```diff
@@ -2,25 +2,28 @@
...
+  - name: "Code Linter"
+    command: "vagrant ssh -c 'cd /home/vagrant/sync/; $(npm bin)/grunt lint'"
+    timeout_in_minutes: 2
 
-  - name: "Build Infusion"
-    command: "vagrant ssh -c 'cd /home/vagrant/sync; grunt clean stylus modulefiles:all pathMap:all copy:all copy:necessities uglify:all concat:all compress:all'"
-    timeout_in_minutes: 20
+  - name: "Browser Tests"
+    command: "npm run test:vagrantBrowser"
+    timeout_in_minutes: 10
```
jobara: I wonder if we should bump this up to timeout at 15 min. On the last CI build it took almost 8 minutes to run the tests, and we'll likely be adding more tests to the mix.

.buildkite/pipeline.yml
```diff
@@ -2,25 +2,28 @@
...
 
-  - name: "Run tests"
-    command: "npm run test:vagrant"
-    timeout_in_minutes: 20
+  - name: "Node Tests"
+    command: "npm run test:vagrantNode"
+    timeout_in_minutes: 10
```
jobara: We could probably leave this at 10 min, but looking at CI, it only took about 33 seconds to run. If you want to keep the same overall time we could drop this to 5 min when we bump up the the browser test timeout. Or we could just leave this at 10 min.

gtirloni: I think that makes sense!

jobara:
@gtirloni I just noticed that the first commit to your branch is actually for FLUID-6265. Are you able to remove that commit?

gtirloni:
@jobara I removed those changes in this commit https://github.com/fluid-project/infusion/pull/889/commits/e08f0c6c81b7890f43e6653afbd007622d7fed9c

Sorry about the mess. I don't know if you use the GitHub interface for merges but it should make it easier to squash them all. I can try doing it here myself but I'm afraid to break the branch further. I could also open a new PR if that's okay.

gtirloni:
I can't get this PR's build to update GitHub (maybe it's due to all the stashing and forced pushes).

Closed in favor of #896

PR #890, opened 2018-04-02
NOJIRA - Fix Ansible inventory warning

gtirloni:
Ansible used to include an inventory file that had "localhost". The default inventory now is empty and this warning is shown:

```
    default:  [WARNING]: provided hosts list is empty, only localhost is available. Note
    default: that the implicit localhost does not match 'all'
```

This shows in red in log output and can confuse users.

I'm proposing an inventory is defined on the fly through a command line parameter, just to satisfy Ansible and clean up the logs. This should be harmless.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,19 +4,23 @@
 steps:
   - name: "Create VM"
     command: "DISPLAY=:0 vagrant up --provider virtualbox && npm install"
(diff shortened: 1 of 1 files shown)


gtirloni:
This was auto-closed because I deleted my branch and forcefully pushed it to my repo. I'll open a new PR (this one got changes that shouldn't be here, sorry for the noise).

PR #892, opened 2018-04-03
FLUID-6176: Highlighting text that is being self voiced

jobara:
Will highlight the visible text that is being self voiced. This should follow along word-by-word as it is being synthesized. The highlighting should work in the various contrast modes. To test the self voicing, please use the prefs framework demo, and make sure to run `grunt buildStylus` to ensure the necessary styles have been generated.

NOTE: that text that wouldn't be included in the accessibility tree (e.g. display: none, visibility: hidden, aria-hidden: true) will not be added to the self voicing and highlighting.

https://issues.fluidproject.org/browse/FLUID-6176


jobara:
@amb26 would you like to review this PR to add the highlighting of text as it is being self voiced. If I remember correctly we were talking about implementation ideas for this several months ago when I first started to look into it.

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            var parent = elm.parent();
+            elm.contents().unwrap();
+            parent[0].normalize();
+        }
+    };
+
+    fluid.prefs.enactor.selfVoicing.handleUtteranceEndEvent = function (that) {
+        that.parseQueue.shift();
+        that.parseIndex = 0;
+        fluid.prefs.enactor.selfVoicing.unWrap(that.locate("mark"));
```
amb26: Very awkward mix of "model-based" action on the queues and dispatch to a global function. The last action should be implemented by a suitably-named invoker, and this handler should be split into two separate listeners

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+        return fluid.isValue(str) && /\S/.test(str);
+    };
 
-            if (node.nodeType === fluid.prefs.enactor.selfVoicing.nodeType.ELEMENT_NODE && window.getComputedStyle(node).display !== "none") {
-                if (node.nodeName === "IMG") {
-                    var altText = node.getAttribute("alt");
-                    if (altText) {
-                        that.tts.queueSpeech(altText);
+    /**
+     * Determines if there is rendered text to in an element.
```
amb26: Typo here

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            $.each(childNodes, function (childIndex, childNode) {
+                if (childNode.nodeType === fluid.prefs.enactor.selfVoicing.nodeType.TEXT_NODE) {
+                    var words = childNode.textContent.split(/(\s+)/); // split on whitespace, and capture whitespace
+                    // charIndex is the start index of the word in the nested block of text
+                    var charIndex = 0;
+
+                    fluid.each(words, function (word) {
+                        if (fluid.prefs.enactor.selfVoicing.isWord(word)) {
+                            parsed.push({
+                                blockIndex: blockIndex,
```
amb26: Factor these two pushes into a common function

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+     *                              It defaults to 0 and is primarily used internally for recursive calls.
+     *
+     * @return {Array} - An array of data points, objects of the with the following structure.
+     *                   {
+                             blockIndex: {Number}, // the index into the entire block of text being parsed
+                             startOffset: {Number}, // the start offset of the current `word` relative to the closest enclosing DOM element
+                             endOffset: {Number}, // the start offset of the current `word` relative to the closest enclosing DOM element
+                             node: {node}, // the current child node being parsed
+                             childIndex: {Number}, // the index of the child node being parsed relative to its parent
+                             parentNode: {node}, // the parent DOM node
```
amb26: Inconsistent use of "node" in some places and "element" in others. It looks like in our core docs we use "DomElement" https://docs.fluidproject.org/infusion/development/ViewAPI.html#fluidgetidelement

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+     * NOTE: consecutive whitespace is collapsed to the first whitespace character.
+     * NOTE: hidden text is skipped.
+     *
+     * @param {jQuery|element} elm - the DOM node to parse
+     * @param {Number} blockIndex - The `blockIndex` represents the index into the entire block of text being parsed.
+     *                              It defaults to 0 and is primarily used internally for recursive calls.
+     *
+     * @return {Array} - An array of data points, objects of the with the following structure.
+     *                   {
+                             blockIndex: {Number}, // the index into the entire block of text being parsed
```
amb26: We tend to be more specific when we can with "Integer" when we know that the value can't be an arbitrary number

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+     * - elm or its parent has `aria-hidden="true"` set.
+     *
+     * @param {jQuery|element} elm - either a DOM node or a jQuery element
+     *
+     * @return {Boolean} - returns true if there is rendered text within the element and false otherwise. (See rules above)
+     */
+    fluid.prefs.enactor.selfVoicing.hasRenderedText = function (elm) {
+        elm = fluid.unwrap(elm);
+
+        return elm && !!elm.offsetHeight && fluid.prefs.enactor.selfVoicing.isWord(elm.innerText) && !$(elm).closest("[aria-hidden=\"true\"]").length;
```
amb26: This is a very long line - I think in other contexts we have been talking about some kind of line limit like 120 or 132? I can't remember the exact figure, perhaps @simonbates can remember. The current git review interface shows me 132 characters.

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            that.parseQueue.push(parsedFromElm);
+            that.events.onReadFromDOM.fire(parsedFromElm);
+            that.tts.queueSpeech(fluid.prefs.enactor.selfVoicing.parsedToString(parsedFromElm));
+        }
+    };
+
+    /**
+     * Returns the index of the closest data point from the parseQueue based on the boundary provided.
+     *
+     * @param {Array} parseQueue - An array data points generated from parsing a DOM structure
```
amb26: This should read something like {ParseQueueElement[]} and the structure defined in a block comment somewhere

src/framework/preferences/js/SelfVoicingEnactor.js
```diff
@@ -92,45 +108,280 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            return fluid.prefs.enactor.selfVoicing.getClosestIndex(parseQueue, prevIndex, boundary);
+        }
+
+        var isWithinNextBound = parseQueue[nextIndex] ? boundary < parseQueue[nextIndex].blockIndex : boundary <= maxBoundary;
+
+        if (currentBlockIndex === boundary || (currentIndex <= maxIndex && isWithinNextBound)) {
+            return currentIndex;
+        }
+
+        return fluid.prefs.enactor.selfVoicing.getClosestIndex(parseQueue, nextIndex, boundary);
```
amb26: JavaScript still does not have "tail recursion elimination" so I suggest a clearer implementation would just be iterative - especially since there is a lot of setup that you do at the beginning of every function call which is simply duplicated each time

src/framework/preferences/js/SelfVoicingEnactor.js
simonbates: I think for GPII, we talked about using a 120 char limit.

src/framework/preferences/js/SelfVoicingEnactor.js
jobara: @amb26 part of the variances may be because in some cases it may refer to a child node of a DOM element which may just be a textnode.

src/framework/preferences/js/SelfVoicingEnactor.js
jobara: @amb26 i'll explore that. However, i do have to down into sibling and descendant DOM Elements to pull out all of the data.

jobara: @amb26 I posted a question in the fluid-work IRC channel about this. ( https://botbot.me/freenode/fluid-work/2018-04-12/?msg=98918443&page=1 ). 

In sure I haven't found a simple way to make this purely iterative, because we don't know ahead of time the nesting of DOM elements. I tried using querySelectorAll or jQuery's find method to get all of the DOM elements, but this would still require dealing with nested elements to preserve the order of the text.

amb26: The DOM may be nested, but as far as I can see this method can be purely linear since it works only on the parseQueue structure which is an array, and the only varying coordinate is the 2nd "index" argument which is just an integer. As far as I can see this body could be replaced by one or two "while" loops, with a great increase in clarity

jobara: @amb26 the method navigates through a nested structure to output a flattened one. I'm still struggling to think of how best to achieve this with the while loops given that the childnodes of an element may be text nodes or elements themselves. And these sub elements may have one or more sub elements and so on.

jobara:
Closing in favour of https://github.com/fluid-project/infusion/pull/898

PR #897, opened 2018-04-17
Add CI status badge

gtirloni:
It seems like good marketing to have a status badge for CI builds (limited to the master branch only, so broken PRs don't have an impact).

It's below the first header because [markdownlint](https://github.com/markdownlint/markdownlint) complains if I put it up top ([MD041](https://github.com/markdownlint/markdownlint/blob/1571bedb7e84be6420c8b2d47461c0bfcc1bea27/lib/mdl/rules.rb#L612) rule).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
-
-## What Is Infusion? ##
+# Infusion #
 
 [![Build status](https://badge.buildkite.com/5c8634255695aaaeda0e48799272f9f0e758e6512829737c94.svg?branch=master)](https://buildkite.com/fluid-project/fluid-infusion)
+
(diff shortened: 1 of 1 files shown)


jobara:
@gtirloni how do you feel about adding an h1 titled Infusion and putting the badge below that?

gtirloni:
@jobara I think that's an excellent idea. I tried it here and it looks okay to me.

avtar:
So fancy! 🎩 Just a heads-up that people without Buildkite Fluid Project accounts will see a login page until they're roll out [read-only access](https://github.com/buildkite/feedback/issues/137).

PR #898, opened 2018-05-01
FLUID-6175: Improved self voicing

jobara:
Adds an Orator component which provides a UI for playing/pausing the self voicing. The orator can be used on its own or as part of the selfVoicing enactor in the prefs framework. 

This PR also includes the text highlighting from FLUID-6176

https://issues.fluidproject.org/browse/FLUID-6175
https://issues.fluidproject.org/browse/FLUID-6176


jobara:
@amb26 could you please review this PR. It includes changes from https://github.com/fluid-project/infusion/pull/892; which has now been closed in favour of this one.

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -121,102 +123,63 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
-            timeout = null;
-            result = func.apply(context, args);
-            if (!timeout) {
-                context = args = null;
-            }
+    // Cancel all synthesis when a page is unloaded.
+    // This is necessary so that the speech synthesis stops when navigating to a new page and so that paused
+    // speaking doesn't prevent new self voicing after a page reload.
+    fluid.textToSpeech.cleanupOnUnload = function () {
+        window.onbeforeunload = function () {
```
amb26: Rather than binding to this directly, create a fluid.window.onBeforeUnload global event that this is added to as a namespaced listener - this will give other integrators some control over this policy if they need to. Preferably also bind to this declaratively rather than programmatically.

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+            play: "play",
+            pause: "pause"
+        },
+        container: "",
+        scope: "",
+        model: {
+            playing: false
+        },
+        members: {
+            container: "@expand:fluid.orator.controller.container({that}.options.container, {that}.options.markup.defaultContainer, {that}.options.scope)"
```
amb26: This is somewhat abuse of the expected workflow of "newViewContainer" which is effectively the standard one in the upcoming framework. The linkage between options.container and members.container should be standard in the base grade, and all actions which are necessary to establish it pushed into the evaluation of options.container, for example like this: https://github.com/fluid-project/fluid-authoring/blob/FLUID-4884/src/js/ComponentGraph.js#L581

src/framework/core/js/Fluid.js
```diff
@@ -890,7 +890,7 @@ var fluid = fluid || fluid_3_0_0;
     /**
      * Copied from Underscore.js 1.4.3 - see licence at head of this file
      *
-     * Will execute the passed in function after the specified about of time since it was last executed.
+     * Will execute the passed in function after the specified ammout of time since it was last executed.
```
amb26: Typo amount

src/framework/core/js/NewViewSupport.js
```diff
@@ -0,0 +1,38 @@
...
+     * which can render its own container into the DOM on startup, whilst the container remains immutable.
+     */
+    fluid.defaults("fluid.newViewComponent", {
+        gradeNames: ["fluid.modelComponent"],
+        members: {
+            // 3rd argument is throwaway to force evaluation of container
+            dom: "@expand:fluid.initDomBinder({that}, {that}.options.selectors, {that}.container)"
+            // The container must be supplied by a concrete implementation, however the below example using
+            // fluid.container can be used in most cases.
+            // container: "@expand:fluid.container({that}.options.container)"
```
amb26: This should be uncommented and be the standard definition, not expected to be overridden by derived grades. See comment above

tests/component-tests/orator/js/OratorTests-Utils.js
```diff
@@ -0,0 +1,124 @@
...
+    fluid.defaults("fluid.tests.orator.domReaderStubs", {
+        gradeNames: ["fluid.tests.orator.stubs"],
+        methods: ["play", "pause"],
+        listeners: {
+            "onCreate.addSpies": {
+                funcName: "fluid.tests.orator.addStubs",
+                priority: "first",
+                args: ["{that}.stubs", "{that}", "{that}.options.methods"]
+            },
+            "onDestroy.restore": {
```
amb26: This namespace should be somehow complementary to the one added to onCreate. But if we are being destroyed, why do we restore something?

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+
+    fluid.orator.domReader.pause = function (that) {
+        if (that.model.speaking && !that.model.paused) {
+            that.tts.pause();
+        }
+    };
+
+    // Accepts a speechFn (either a function or function name), which will be used to perform the
+    // underlying queuing of the speech. This allows the SpeechSynthesis to be replaced (e.g. for testing)
+    fluid.orator.domReader.queueSpeech = function (that, speechFn, text, interrupt, options) {
```
amb26: No public docs for this function, including the interesting "interrupt" argument. Also its argument list has become very bulky. Better to structure this entirely as a (that, options) structure with named arguments.
Also, it would be better to structure this activity as an event pipeline with namespaced listeners, which will make it easier to chop and change implementations for mocking. In that way we should be able to get rid of the peculiar "speechFn" argument

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+        var parsed = [];
+        elm = fluid.unwrap(elm);
+        blockIndex = blockIndex || 0;
+
+        if (fluid.orator.domReader.hasTextToRead(elm)) {
+            var childNodes = elm.childNodes;
+
+            $.each(childNodes, function (childIndex, childNode) {
+                if (childNode.nodeType === fluid.orator.domReader.nodeType.TEXT_NODE) {
+                    var words = childNode.textContent.split(/(\s+)/); // split on whitespace, and capture whitespace
```
amb26: Stick this "split" action into an invoker so that it can be a configurable policy

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+                    // charIndex is the start index of the word in the nested block of text
+                    var charIndex = 0;
+
+                    fluid.each(words, function (word) {
+                        if (fluid.orator.domReader.isWord(word)) {
+                            fluid.orator.domReader.addParsedData(parsed, word, childNode, blockIndex, charIndex, childIndex);
+                            blockIndex += word.length;
+                        // if the current `word` is not an empty string and the last parsed `word` is not whitespace
+                        } else if (word && fluid.orator.domReader.isWord(fluid.get(parsed, [(parsed.length - 1), "word"]))) {
```
amb26: Invoke this as an invoker too

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+            });
+        }
+
+        return parsed;
+    };
+
+    /**
+     * Combines the parsed text into a String.
+     *
+     * @param {Array} parsed - An array of parsed data points
```
amb26: Array of what? Make a JSDocs @typedef if it is some widely used structure

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+
+        return elm &&
+               !!elm.offsetHeight &&
+               fluid.orator.domReader.isWord(elm.innerText) &&
+               !$(elm).closest("[aria-hidden=\"true\"]").length;
+    };
+
+    /**
+     * Adds a data point to an array of parsed DOM elements.
+     * Structure of each data point is as follows:
```
amb26: @typedef this

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+     * @param {Component} that - the component
+     * @param {String|jQuery|DomElement} elm - The DOM node to read
+     */
+    fluid.orator.domReader.readFromDOM = function (that, elm) {
+        elm = $(elm);
+
+        // only execute if there are nodes to read from
+        if (elm.length) {
+            var parsedFromElm = fluid.orator.domReader.parse(elm[0]);
+            that.parseQueue = parsedFromElm;
```
amb26: I worry about this wholesale obliteration of a mutable structure. I think we need a bit more formality and modelisation of this structure even if we do not store the parseQueue itself in a model for reasons of efficiency.

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+     * the appropriate text in the markup specified by `that.options.markup.highlight`.
+     *
+     * @param {Component} that - the component
+     * @param {Integer} boundary - the boundary point used to find the text to highlight. Typically this is the
+     *                             utterance boundary returned from the utteranceOnBoundary event.
+     */
+    fluid.orator.domReader.highlight = function (that, boundary) {
+        that.removeHighlight();
+
+        if (that.parseQueue.length) {
```
amb26: This should be a modelListener to something rather than an arbitrarily scheduled invoker

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+        markup: {
+            highlight: "<mark class=\"flc-orator-highlight fl-orator-highlight\"></mark>"
+        },
+        events: {
+            onReadFromDOM: null
+        },
+        model: "{tts}.model",
+        members: {
+            parseQueue: [],
+            parseIndex: 0,
```
amb26: I think a reasonable compromise would be to modelise i) parseIndex, and ii) parseQueue.length and key most of this component's activities off that - even if the current ChangeApplier doesn't make it a great idea to stash the entire queue in the model.

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+        selectors: {
+            highlight: ".flc-orator-highlight"
+        },
+        markup: {
+            highlight: "<mark class=\"flc-orator-highlight fl-orator-highlight\"></mark>"
+        },
+        events: {
+            onReadFromDOM: null
+        },
+        model: "{tts}.model",
```
amb26: This will then end up being nested one level deeper.

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+            that.parseQueue = parsedFromElm;
+            that.events.onReadFromDOM.fire(parsedFromElm);
+            that.tts.queueSpeech(fluid.orator.domReader.parsedToString(parsedFromElm));
+        }
+    };
+
+    /**
+     * Returns the index of the closest data point from the parseQueue based on the boundary provided.
+     *
+     * @param {Array} parseQueue - An array data points generated from parsing a DOM structure
```
amb26: An array "of" - also, write out something like Array < ParseQueueElement >

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+     * @param {Integer} currentIndex - The index into the paraseQueue to start searching from. The currentIndex will be
+     *                                constrained to the bounds of the parseQueue.
+     * @param {Integer} boundary - The boundary value used to compare against the blockIndex of the parsed data points.
+     *
+     * @return {Integer|undefined} - Will return the index of the closest data point in the parseQueue. If the boundary
+     *                               cannot be located within the parseQueue, `undefined` is returned.
+     */
+    fluid.orator.domReader.getClosestIndex = function (parseQueue, currentIndex, boundary) {
+        var maxIdx  = Math.max(parseQueue.length - 1, 0);
```
amb26: I don't see a need for the inconsistency wrt. "Idx" and "Index"

src/components/orator/js/Orator.js
```diff
@@ -0,0 +1,571 @@
...
+        var currentBlockIndex = parseQueue[currentIndex].blockIndex;
+        var maxBoundary = parseQueue[maxIdx].blockIndex + parseQueue[maxIdx].word.length;
+
+
+        if (!fluid.isValue(boundary) || boundary < 0 || boundary > maxBoundary ) {
+            return undefined;
+        }
+
+        if (currentBlockIndex > boundary) {
+            return fluid.orator.domReader.getClosestIndex(parseQueue, prevIdx, boundary);
```
amb26: I still don't see any reason for this function to be recursive, least of all doubly recursive, and would be much clearer if it was rewritten so that all the initial setup happens just once

src/components/orator/js/Orator.js
jobara: In speaking with @amb26 today it was suggested to make a domParser subcomponent on the domReader. We'd put in all the specific parsing functions into that subcomponent.

src/components/orator/js/Orator.js
jobara: @amb26 and I discussed this in more detail today. The definition of the "container" should be in the base grade and not overridden. Rather we can have the expander refer to an invoker where any necessary changes can be applied.

jobara: Align the structure more with what's in "fluid.author.containerRenderingView", see link above. And add an additional invoker for rendering into the DOM.. e.g to allow for prepending or appending into the parentContainer.

src/components/orator/js/Orator.js
jobara: Explore re-writing this as a while loop with break conditions.

src/components/orator/js/Orator.js
jobara: should probably modelise the boundary index as well.

jobara: May be able to setup a modelRelay to a closestIndex model value which has a relay rule that contains the functionality currently in getClosestIndex. Then an updateHighlight method could be used to highlight the appropriate portion using this index.

tests/component-tests/orator/js/OratorTests-Utils.js
jobara: The restore on destroy can be removed because we are just adding spies to our own methods.

src/components/orator/js/Orator.js
jobara: Talked about the event pipeline in the channel.
https://botbot.me/freenode/fluid-work/2018-07-05/?msg=101795899&page=1

jobara:
The comments on this PR have been addressed in the work for FLUID-6177 and have been included in its PR. This PR is being closed in favour of https://github.com/fluid-project/infusion/pull/903

PR #899, opened 2018-05-03
FLUID-6279: improving handling of "normal" initial line-height.

jobara:
When a browser returns "normal" as the initial line-height, the enactor will set the line-height back to "normal" when no line-height multiplier needs to be set. That is, when there are no preferences for line-height to apply.

https://issues.fluidproject.org/browse/FLUID-6279


jobara:
@amb26 could you also take a look at this PR to address the jank issues when adding UIO to a page.

amb26:
Thanks so much for coming up with a fix for this issue so quickly, together with great test cases. I checked that the line-height based jank is now gone from the internal demo and so hopefully it will be from our deployed site as well.

PR #900, opened 2018-05-04
FLUID-6277 - Cleanup code coverage reports

gtirloni:
Just a cleanup step to cleanup the code coverage reports directories so they don't start to buildup  over time (causing memory issues).

We could simply delete the build directory and do a clean `git clone` every time, but this would be too slow (and increase the chances of issues in GitHub causing builds to break).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -13,7 +13,7 @@ steps:
   - name: "Build"
     command:
       - "DISPLAY=:0 vagrant up --provider virtualbox"
-      - "vagrant ssh -c 'cd /home/vagrant/sync/; $(npm bin)/npm run pretest'"
+      - "vagrant ssh -c 'cd /home/vagrant/sync/; npm run pretest'"
     timeout_in_minutes: 20
     <<: *agent_queue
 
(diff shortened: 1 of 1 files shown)


the-t-in-rtf:
The test runs should already clean up the coverage and reports directories before each run, it should never "build up" (that would in fact be counter to the point of creating a report):

https://github.com/fluid-project/infusion/blob/master/package.json#L23

If you're talking about each build directory containing that data afterward, then as long as it's available somewhere, cleaning up afterwards sounds fine.

However, it really sounds like you're saying that there is significant reuse of material between runs.  Could you explain what is preserved between runs a bit better?

gtirloni:
Thanks @the-t-in-rtf! The default behavior is to preserve everything that is in .gitignore (`git clean -fqd`). We could switch to `git clean -fqdx` to really remove everything but it causes some issues with Vagrant and Buildkite (but could be worked around).

Answering your other question, yes, CI runs the steps individually so it can be reported as distinct phases in the UI.

I've switched to using `pretest` in this PR to avoid having this code duplicated in the repo and in the CI configuration.

PR #902, opened 2018-05-17
FLUID-6281: Adding test case to verify issue.

jobara:
Adding a test case to verify the Issue.

https://issues.fluidproject.org/browse/FLUID-6281


gtirloni:
ok to test

PR #904, opened 2018-05-24
NOJIRA - Remove edit-hosts.sh hack

gtirloni:
This doesn't seem necessary anymore. All tests pass without it.

@avtar do you remember what failed because of this?


avtar:
I think this was a workaround for a [Vagrant issue](https://github.com/hashicorp/vagrant/pull/6203) back in the day which was causing Ansible to freak out. It looks like Hashicorp addressed it with a later release.

gtirloni:
@avtar thanks, it seems Ansible is fine with that now. Just a very minor cleanup then.

gtirloni:
Ready for review.

gtirloni:
Not a priority and too small for review. I'll remember to add it to a more substantial PR in the future to make more efficient usage of the review process.

PR #909, opened 2018-06-18
FLUID-6282 - Use containers and headless browsers for CI

gtirloni:
This switches CI from using Vagrant and a full desktop environment to using Docker containers and headless browsers.

Test results are pushed as separate Docker images `fluidprojectorg/infusion-build-results` tagged by commit hash.

This is a work in progress. Sending PR to collect feedback.


gtirloni:
Build results: https://buildkite.com/fluid-project/infusion/builds/270

avtar:
@gtirloni I used macOS 10.12.6 and Docker 18.03.1-ce to test these changes. I didn't test the publish/push related commands though but based on the Buildkite logs it seems they're good as-is.
```
export TMP_ID="infusion-build-results-$(openssl rand -hex 4)"

docker build -t ${TMP_ID} .
docker run --rm -ti -v ${TMP_ID}:/app ${TMP_ID} true
docker run --rm -ti -v ${TMP_ID}:/app -w /app inclusivedesign/node-browsers grunt lint
docker run --rm -ti -v ${TMP_ID}:/app -w /app inclusivedesign/node-browsers npm run test:node
docker run --rm -ti --cap-add=SYS_ADMIN -e HEADLESS=1 -v ${TMP_ID}:/app -w /app inclusivedesign/node-browsers npm run test:browser

docker run --name ${TMP_ID} -v ${TMP_ID}:/app busybox sh -c 'cp -R /app /build'
docker container rm -f ${TMP_ID}
docker volume rm ${TMP_ID}
docker image rm -f ${TMP_ID}
```
Do you think it's worth using ``npm publish`` in addition to, or maybe just instead of, publishing [these temporary images](https://github.com/fluid-project/infusion/pull/909/files#diff-8b496edab86012a437e80897eb7129c0R6)?

Any thoughts regarding [node-browsers](https://hub.docker.com/r/inclusivedesign/node-browsers/) vs. 'headless-browsers' as the name for this Docker image? 

BTW seeing these tests run without a VM is just awesome!

gtirloni:
> Do you think it's worth using npm publish in addition to, or maybe just instead of, publishing these temporary images?

I would like to have a separate pipeline for publishing because publishing all and every build might make things too cluttered in the npm registry w.r.t. to PR's. Maybe only for the master branch? I still have to read more about [fluid-publish](https://github.com/fluid-project/fluid-publish) and how people would like to approve releases in general.

I'm still concerned about these temporary images (besides how to garbage collect them) because I haven't gotten any feedback about what is actually being used from the test results. My guess is that the 3.5MB `reports` directory is all that matters (and then publishing a image so big would be a huge waste). I think I can figure out a BK plugin setup where these things are just uploaded to a static place or committed to some object store. BK supports S3 but we got all sorts of issues when we uploaded the test results there (the HTML seems too complex for the restrictive S3 settings that BK is using).

> Any thoughts regarding node-browsers vs. 'headless-browsers' as the name for this Docker image?

I think my only objection would be that the browsers baked into these images could also work in a Xorg, Xfvb or VNC environment in full non-headless mode.

Thanks for reviewing this! :+1:

avtar:
> Maybe only for the master branch?

+1 for that. ``npm view connect versions`` shows a bunch of dev releases but I don't think there's a set schedule for them. Perhaps in the future those could be published using a [dashboard type of approach](https://github.com/buildkite/block-step-example).

> My guess is that the 3.5MB reports directory is all that matters (and then publishing a image so big would be a huge waste). 

Yeah that sounds like quite the overhead.

> I think I can figure out a BK plugin setup where these things are just uploaded to a static place or committed to some object store.

Uploading just the generated reports seems ideal. Perhaps to the CDN or build server in directories named after the git commit hash?

> I think my only objection would be that the browsers baked into these images could also work in a Xorg, Xfvb or VNC environment in full non-headless mode.

Ah right, that makes sense.

gtirloni:
Closing PR to cleanup CI statuses. I'll open a new one with a simplified proposal.

Thanks for the review.

PR #914, opened 2018-07-23
FLUID-6296: YouTube captions preference

jobara:
Adding a preference and enactor to turn on/off captions in an embedded YouTube video.

https://issues.fluidproject.org/browse/FLUID-6296


jobara:
@the-t-in-rtf would you be able to look at the test reports. I'm noticing that they don't include the tests for the news files (e.g. CaptionsPanel.js and CaptionsEnactor.js).

the-t-in-rtf:
Hi, @jobara.  They aren't showing up in the report because you didn't add them to [the "all tests" rollup](https://github.com/jobara/infusion/blob/FLUID-6296/tests/all-tests.html#L49).  I added them there to confirm and reran, and sure enough, the results showed up.

For the sake of completeness, you should also add each of the new test fixtures to [the preferences rollup](https://github.com/jobara/infusion/blob/FLUID-6296/tests/framework-tests/preferences/all-tests.html#L17).   It's not used in the automated tests, but devs in the larger group do use them to troubleshoot problems.

jobara:
&#129318; thanks @the-t-in-rtf

jobara:
@amb26 could you please review this PR

src/framework/preferences/js/CaptionsEnactor.js
```diff
@@ -0,0 +1,245 @@
...
+        },
+        listeners: {
+            "onCreate.checkYouTubeAPI": "fluid.prefs.enactor.captions.waitForYouTubeAPI"
+        }
+    });
+
+    // After FLUID-6148 (https://issues.fluidproject.org/browse/FLUID-6148) is complete. It should be possible to
+    // just source the dynamic components with the videos selector. The framework should at that time be able to
+    // handle the asynchrony of waiting for YT in the expander that creates the player.
+    fluid.prefs.enactor.captions.waitForYouTubeAPI = function (that) {
```
amb26: Standard docs would be helpful - say what is returned and what aspects of the component are relied on

src/framework/preferences/js/CaptionsEnactor.js
```diff
@@ -0,0 +1,245 @@
...
+
+    fluid.defaults("fluid.prefs.enactor.captions", {
+        gradeNames: ["fluid.prefs.enactor", "fluid.viewComponent"],
+        preferenceMap: {
+            "fluid.prefs.captions": {
+                "model.enabled": "default"
+            }
+        },
+        events: {
+            onVideoElmLocated: null
```
amb26: I don't think the abbreviation of "Element" to "Elm" hlps readability

src/framework/preferences/js/CaptionsEnactor.js
```diff
@@ -0,0 +1,245 @@
...
+            });
+        }, function (error) {
+            fluid.log(fluid.logLevel.WARN, error);
+        });
+
+        if (fluid.get(window, ["YT", "Player"])) {
+            promise.resolve();
+        } else {
+            // the YouTube iframe api will call onYouTubeIframeAPIReady after the api has loaded
+            window.onYouTubeIframeAPIReady = promise.resolve;
```
amb26: Rather than blindly bung this onto a global, please make a singleton component using resolveRootSingle which mediates this interaction

src/framework/preferences/js/CaptionsEnactor.js
```diff
@@ -0,0 +1,245 @@
...
+                var paramVals = strParam.split("=");
+                params[paramVals[0]] = paramVals[1];
+            });
+        }
+
+        return params;
+    };
+
+    /**
+     * Adds the "enabledjsapi=1" query paramater to the query string at the end of the src attribute.
```
amb26: Typo parameter

src/framework/preferences/js/CaptionsEnactor.js
```diff
@@ -0,0 +1,245 @@
...
+     * If "enabledjsapi" already exists it will modify its value to 1. This is required for API access
```
amb26: Typo enabled

src/framework/preferences/js/CaptionsEnactor.js
```diff
@@ -0,0 +1,245 @@
...
+    });
+
+    /**
+     * Convert queryString into a parameter object. Note that repeated parameters will be replaced with the last one.
+     *
+     * @param {String} queryStr - the query string to parse. It assumes the first character is not "?";
+     *
+     * @return {Object} - An Object representation of the key/value pairs from the query string.
+     */
+    fluid.prefs.enactor.captions.youTubePlayer.parseQueryString = function (queryStr) {
```
amb26: I believe this function is not required in the modern world - https://developer.mozilla.org/en-US/docs/Web/API/URL

jobara: It doesn't look like the URL API is supported in IE 11 ( https://caniuse.com/#search=URL ). I'll leave a comment that this is deprecated and should be switched to use the URL API after IE 11 is no longer supported.

jobara:

jobara:
@amb26 suggested that the singleton should be responsible for creating the promise to determine if the YT API is available. See: https://botbot.me/freenode/fluid-work/2018-07-25/?msg=102446834&page=1

jobara:
@amb26 I have added the polyfill and switched to using the URL api from the browser to handle the video source url. The PR is ready for more review.

the-t-in-rtf:
@jobara, I've just cut a new dev release of gpii-grunt-lint-all that [has global excludes](https://issues.gpii.net/browse/GPII-3245) by default.  I have tested it in my own work on separate
projects, but would appreciate it if you could quickly update to  1.0.1-dev.20180731T094032Z.112cde7 and confirm that you're able to remove the excludes you added in the other pull.

jobara:
@the-t-in-rtf I tested out your dev release of gpii-grunt-lint-all ( 1.0.1-dev.20180731T094032Z.112cde7 ). I'm still finding that doing things like "**/*.js" will go into the node_modules for the lintspaces test. Also "**/*.md" goes into node modules for the markdownlint test. "**/*.json" goes into node_modules as well for the jsonlint task.

I didn't yet test that package-lock.json is excluded. 

I committed the updated config for fluid-publish into a [branch](https://github.com/jobara/fluid-publish/tree/test-lint-all).

the-t-in-rtf:
Thanks, @jobara.  I took a few minutes to figure out what was going on in your branch.

For starters, I refined the gpii-grunt-lint-all plugin to exclude `.DS_Store` files by default, update to `1.0.1-dev.20180801T111658Z.da9c3ba` to pick that up.  I then discovered a few things about the pattern matching, which ultimately relies on [minimatch](https://www.npmjs.com/package/minimatch).

First, pattern precedence is left to right, and you can reintroduce excluded material further to the right.  They seem to think of this as a feature, which you can use to twiddle the order of files, by excluding them out and then bringing them back in in a particular order.  To deal with this, I made sure the new ignores were the right-most material in my rollup definitions.

However, it seems that it's also possible to break out of negated patterns.  So, here's what you had:

```
        lintAll: {
            sources: {
                md: [ "**/*.md"],
                js: ["**/*.js"],
                json: ["**/*.json"],
                other: ["./.*", "!./**/.DS_Store"]
            }
        }
```
The pattern `**/*.md` seems to bypass the inherited `!./node_modules/**/*` negation, and bring undesirable content from `node_modules` into scope.  Although we might be able to track this down further and report a bug against minimatch, for now the solution is to recognise the dangers of broad patterns like `**/*.js` and use more targeted settings like the following, which I tested with your branch:

```
        lintAll: {
            sources: {
                md: [ "./*.md"],
                js: ["./tests/**/*.js", "./*.js"],
                json: ["./tests/**/*.json", "./*.json"],
                other: ["./.*"]
            }
        }
```

That configuration should lint everything that needs to be linted, and ignore everything that should be ignored.

While I was trying out your branch, I also noted that [the bit of the file that loads the package.json](https://github.com/jobara/fluid-publish/blob/test-lint-all/Gruntfile.js#L19) is no longer used, and can be removed.

jobara:
@the-t-in-rtf thanks for looking into this. I've filed a new JIRA to updated gpii-grunt-lint-all in fluid-publish. https://issues.fluidproject.org/browse/FLUID-6313.

PR #915, opened 2018-07-24
FLUID-6297: Adding low contrast themes

jobara:
Added two more low contrast themes: black on brown, grey on white.

https://issues.fluidproject.org/browse/FLUID-6297


demos/prefsFramework/js/injectSVG.js
```diff
@@ -0,0 +1,61 @@
...
+    demo.injectSVG.fetch = function (that) {
+        $.ajax(that.options.svgSource, {
+            dataType: "html",
+            type: "GET",
+            success: that.events.onSuccess.fire,
+            error: that.events.onError.fire
+        });
+    };
+
+    demo.injectSVG.inject = function (that, svgMarkup) {
```
amb26: It looks like this could productively be added to the NewViewSupport.js that was added as part of the orator work. Note that there is a work-alike in https://github.com/fluid-project/fluid-authoring/blob/FLUID-4884/src/js/ComponentGraph.js#L567 - it is worth investigating whether these achieve the same effect, or whether this approach might even allow us to do without a wrapper. In any case we should move this component over to https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/ResourceLoader.js#L27 which would eliminate a lot of boilerplate.

jobara:
@amb26 I added a new view support that renders a fetched template. This PR is ready for another round of review.

PR #916, opened 2018-07-25
FLUID-6309 - Use npm prepublishOnly instead of prepublish

gtirloni:
This effectively ensures that "grunt BuildDist && grunt BuildStylus" only run during "npm publish" and not at "npm install" time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "repository": "git://github.com/fluid-project/infusion.git",
     "main": "./src/module/fluid.js",
     "scripts": {
-        "prepublish": "npm run buildDists && npm run buildStylus",
+        "prepublishOnly": "npm run buildDists && npm run buildStylus",
         "pretest": "node node_modules/rimraf/bin.js reports/* coverage/*",
         "test": "npm run test:browser && npm run test:node",
         "test:browser": "node node_modules/testem/testem.js ci --file tests/testem.js",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "repository": "git://github.com/fluid-project/infusion.git",
     "main": "./src/module/fluid.js",
     "scripts": {
-	"postinstall": "npm run buildStylus",
+        "postinstall": "npm run buildStylus",
         "prepublishOnly": "npm run buildDists",
         "pretest": "node node_modules/rimraf/bin.js reports/* coverage/*",
         "test": "npm run test:browser && npm run test:node",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "repository": "git://github.com/fluid-project/infusion.git",
     "main": "./src/module/fluid.js",
     "scripts": {
-        "prepublishOnly": "npm run buildDists && npm run buildStylus",
+	"postinstall": "npm run buildStylus",
+        "prepublishOnly": "npm run buildDists",
         "pretest": "node node_modules/rimraf/bin.js reports/* coverage/*",
         "test": "npm run test:browser && npm run test:node",
         "test:browser": "node node_modules/testem/testem.js ci --file tests/testem.js",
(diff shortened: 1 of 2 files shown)


gtirloni:
Looks like this shaves 1-3min from `npm install`.

amb26:
I approve of these changes in theory but I am a bit suspicious since I recall that some of the built CSS files are necessary for some tests to pass. @jobara - do you recall what these are? I wonder why this is getting through CI.

jobara:
@amb26 and @gtirloni I believe it's the responsive tests for the the prefs framework that require the stylus files built. Also @gtirloni how are the demos deployed? They would also need built versions of the stylus files.

jobara:
@gtirloni that's interesting. I guess we could skip that step if we move buildStylus to run with install.

gtirloni:
@jobara do you mean in the CI configure as-is today or in this PR? I've tested with a clean checkout and only renamed to "prepublishOnly" -- all tests pass. Shouldn't they fail if buildStylus is required?

jobara:
@gtirloni i guess you'd be able to skip them in both cases. Although I suppose it is good to make sure the build actually works when we are running the tests. However, we don't have many tests for it.

Regarding the tests passing. Did you have that line you mentioned above still present?

gtirloni:
I had moved "prepublish" to "prepublishOnly" in my tests locally.

I just realized that the Infusion PR's are running in BuildKite and the .buildkite/pipeline.yml configuration doesn't have the line I mentioned. That one is for Jenkins and it's currently only running when something is merged. So CI is passing in this PR's build with just the change I proposed here (and the usual steps defined [here](https://github.com/fluid-project/infusion/blob/master/.buildkite/pipeline.yml)). Sorry for make it confusing.

So am I right to assume that, besides the renaming of prepublish to prepublishOnly, we just need to keep "buildStylus" running before `npm install` at all cases then? In practice that will remove the "buildDist" from running on install and that's all, I think.

jobara:
@gtirloni assuming I'm not missing anything again, I think that's correct. The buildStylus task is required for the dists but should be run as part of that task. It is also required for demos and tests which I don't believe are part of the dist but are used for development, deploying to the build site, and running tests.

gtirloni:
Could it be because grunt is running during `vagrant up`? 
https://github.com/fluid-project/infusion/blob/master/provisioning/vars.yml#L17

CI does this in sequence:
* `vagrant up`
* `grunt lint`
* `npm run test:vagrantNode`
* `npm run test:vagrantBrowser`

I'd like to propose a simplified Vagrantfile that doesn't use Ansible but I'm afraid I won't have time to pursue it. In any case, it's [here](https://github.com/gtirloni-fluid/infusion/pull/1) for future reference.

ok to test

amb26:
@gtirloni - I can't see that affects the issue. All vagrant up does is npm install, and then it expects to run the tests immediately after that. If the checkout directory is clean, they should fail, as they do locally.

gtirloni:
I meant that because `vagrant up` is running Ansible which is running the list commands [here](https://github.com/fluid-project/infusion/blob/master/provisioning/vars.yml#L15-L17) which includes calling `grunt` which, if my understanding of the Gruntfile is right, is going to run the `default` tasks which runs `build:all`. I think that last bit runs `grunt stylus:compile` which could be satisfying the dependency.

amb26:
Oh I see sorry. I see you've been trying to tell me this for two rounds of comments now :)
That's pretty wierd - can we just eliminate the grunt step? It causes CI to not be a realistic test of usage.

gtirloni:
OK, @amb26 looks like this is starting to make sense :)

CI failed without that step and it worked when that step was added to `postinstall`.

gtirloni:
@jobara @amb26 thanks for helping sort this out! 🤖

amb26:
No worries @gtirloni , thanks for your patience with it!

PR #913, opened 2018-07-23
FLUID-6305: fixes / test coverage for locale bug in fluid.fetchResources

waharnum:
This PR (based on investigation by @BlueSlug and myself) fixes a bug in `fluid.fetchResources` that emerged in working on adding several translated message bundles for UIO (https://issues.fluidproject.org/browse/FLUID-6299), and adds test coverage for confirming it fixed in the tests for both `fluid.fetchResources` and `fluid.resourceLoader`.

The underlying bug is described in https://issues.fluidproject.org/browse/FLUID-6305 and its impact on `fluid.resourceLoader` in https://issues.fluidproject.org/browse/FLUID-6306

Tagging @amb26 for review per recent email exchange.

PR #917, opened 2018-07-27
FLUID-6299: added multilingual message bundles and defaultLocale option

BlueSlug:
This work depends on outstanding unmerged work for FLUID-6305 and FLUID-6306, see pull request #913 

Added multilingual message bundles for UIO for the following languages:
- English
- Farsi
- French
- Spanish

English is simply the existing bundles that have been renamed to have "_en" at the end, so there are no longer any "non-localized" message bundles.

We (@waharnum and I) have added a new top-level option on UIO to set the default locale (`defaultLocale`) similar to the existing options for the TOC templates and lazyload.


BlueSlug:
@jobara can you please review? thanks!

BlueSlug:
Also, some changes for FLUID-6300 have been based on the work in this branch, and further refactoring has been done. I haven't yet opened a pull request for that branch, but it's possible that some of the things you point out in review will already have been addressed.

tests/component-tests/uiOptions/js/UIOptionsTests.js
```diff
@@ -15,30 +15,150 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-        jqUnit.module("UIOptions Tests");
+    fluid.defaults("fluid.uiOptions.testPrefsEditorCustomToc", {
+        gradeNames: ["fluid.uiOptions.testPrefsEditorBase"],
+        tocTemplate: customizedTocTemplate
+    });
+
+    fluid.defaults("fluid.uiOptions.prefsEditorCustomTocTester", {
+        gradeNames: ["fluid.test.testCaseHolder"],
+        modules: [{
+            name: "UIOptions Tests",
```
jobara: Should probably be "UI Options Tests".

tests/component-tests/uiOptions/js/UIOptionsTests.js
```diff
@@ -15,30 +15,150 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+                name: "Pass in customized toc template",
+                expect: 2,
+                sequence: [{
+                    event: "{prefsEditorCustomTocTest prefsEditor}.events.onCreate",
+                    listener: "jqUnit.assertEquals",
+                    args: ["The toc template is applied properly to the pageEnhancer", customizedTocTemplate, "{prefsEditor}.enhancer.uiEnhancer.fluid_prefs_enactor_tableOfContents.options.tocTemplate"]
+                },
+                {
+                    funcName: "fluid.identity"
+                },
```
jobara: I'm assuming this call to `fluid.identity` is to try to get around having the two events be back-to-back. Rather than do it this way, can you just do all the tests after the last event. If you really need to listen and run tests to the different events independently and you are not able to just specify both events, please make sure to leave a comment and link to the relevant JIRA issue for why this is necessary.

jobara: Also, it appears that this test could just use the plain jqUnit.async test instead. You'd have separate listeners to each event and run jqUnit.start() in the listener that executes last.

tests/component-tests/uiOptions/js/UIOptionsTests.js
```diff
@@ -15,30 +15,150 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
-                listeners: {
-                    onReady: function (that) {
-                        jqUnit.assertEquals("The toc template is applied properly to the pageEnhancer", customizedTocTemplate, that.enhancer.uiEnhancer.fluid_prefs_enactor_tableOfContents.options.tocTemplate);
-                        jqUnit.assertEquals("FLUID-5474: The toc template is applied properly to iframeEnhancer", customizedTocTemplate, that.prefsEditorLoader.iframeRenderer.iframeEnhancer.fluid_prefs_enactor_tableOfContents.options.tocTemplate);
-                        jqUnit.start();
-                    }
+    fluid.defaults("fluid.uiOptions.prefsEditorLocalizedTester", {
+        gradeNames: ["fluid.test.testCaseHolder"],
+        modules: [{
+            name: "UIOptions Locale Tests",
```
jobara: Similar to above, rename to "UI Options"

tests/component-tests/uiOptions/js/UIOptionsTests.js
```diff
@@ -15,30 +15,150 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 (function ($) {
     "use strict";
 
-    $(document).ready(function () {
+    var customizedTocTemplate = "../../../../src/components/tableOfContents/html/TableOfContents.html";
+
+    /* Mixin grade for UIO test component */
+    fluid.defaults("fluid.uiOptions.testPrefsEditorBase", {
```
jobara: Better to put all of the test grades, functions, and etc in the "fluid.tests" namespace. This should ensure that there are no accidental collisions with the actual component and functions under test, and keeps things semantically grouped together in the global object.

tests/component-tests/uiOptions/js/UIOptionsTests.js
```diff
@@ -15,30 +15,150 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        prefsEditor: {
+            path: "slidingPanelPanelLabel",
+            expected: "Préférences de l'utilisateur"
+        }
+    };
+
+    fluid.uiOptions.prefsEditorLocalizedTester.verifyLocalizedMessages = function (prefsEditor) {
+        fluid.each(prefsEditor.prefsEditorLoader.messageLoader.resources, function (panel, key) {
+            var actualValue = panel.resourceText[localizedValuesToVerify[key].path];
+            jqUnit.assertEquals("Panel " + key + " localized message loaded correctly", localizedValuesToVerify[key].expected, actualValue);
```
jobara: This doesn't seem to test that the localized values are actually rendered.

jobara:
Could you also localize the captions.json message bundle

jobara:
The prefsFramework demo and some of the examples have their own message bundles. Should these be specified with the "en" locale?

jobara:
I'm thinking a bit more about there no longer being any "non-localized" fall backs. If an integrator supplies a defaultLocale that doesn't exist, the component will break. Is this what we want? Let me know what you think. Also should we specify the specific locales for the language, will it still get picked up if the integrator requests a different version of that locale? Also maybe we should go ahead and have separate localizations for en-us and en-ca, in particular for things like "colours vs colors".

BlueSlug:
Thanks for your review so far, @jobara. I'm making changes to address the items you've mentioned. So far I've:
- updated the test names to "UI Options" instead of "UIOptions"
- added non-localized message files back into the project
- added en_CA and en_US files to the project, where en_CA is simply a duplicate of the en and the non-localized bundles (the only differences I noticed were in the spelling of "grey/gray" and "colour/color")
- added new bundles for the captions panel in French, Farsi, and machine translated Spanish
- added all variables in the test file to the "fluid.tests.uiOptions" namespace

I still have to:
- verify that the rendered strings match the stored strings after a locale change
- check out how to use jqUnit.asyncTest in this context. I need to do a bit of reading to fully understand that and why using fluid.identity is not ideal where I've put it

Other comments:
- before adding the non-localized bundle back in, I confirmed that the panel broke when the locale was set to a value that wasn't present in the existing bundles. For the sake of avoiding regressions or issues with legacy projects, I've taken your advice.
- an integrator could specify a locale with region included (like "fr_CA"), and if that particular locale doesn't exist the fallback rules will load the next best thing ("fr" in this case)
- perhaps an en_GB bundle could be added as well (differences in words like "Emphasize/Emphasise")

BlueSlug:
@jobara Thanks for your help in moving this along. I've made the remaining changes you requested

src/framework/preferences/messages/contrast.json
```diff
@@ -5,7 +5,7 @@
     "contrast-by": "Black on yellow",
     "contrast-yb": "Yellow on black",
     "contrast-lgdg": "Light grey on dark grey",
-    "contrast-gw": "Grey on white: ",
+    "contrast-gw": "Grey on white",
```
jobara: thanks for catching this typo

src/framework/preferences/messages/letterSpace_fa.json
```diff
@@ -0,0 +1,6 @@
+{
+    "label": "",
+    "description": "",
+    "increaseLabel": "",
+    "decreaseLabel": ""
+}
```
jobara: It seems that the translations for this adjuster are missing.

src/framework/preferences/messages/enhanceInputs_en_US.json
```diff
@@ -0,0 +1,6 @@
+{
+    "label": "Enhance Inputs",
+    "description": "Emphasize links, buttons, menus, textfields, and other inputs",
+    "switchOn": "on",
+    "switchOff": "off"
```
jobara: There seems to be some inconsistency for adjusters that have a the on/off switch. In some message bundles the values ore ON/OFF and others on/off. Could you please update them to be consistent. The rendered casing is probably handled by CSS, but it's worthwhile to double check.

src/framework/preferences/messages/letterSpace_fa.json
BlueSlug: I spoke with Sepi about this one, and while I initially thought it was an oversight, I'd done it intentionally. It turns out that the idea of letter spacing in Farsi doesn't really apply, and I didn't want it to fall back to another localization, so I added the blank strings with the anticipation that we would be working on the ability to activate/deactivate certain preferences by localization at some point (I believe there is a Jira for this but I can't find it at the moment). I wanted to add a comment in the file but Atom didn't like it, so I didn't.

What do you think is the best solution for this issue?

jobara: It's a good question. I think it's probably better to have the localized text, otherwise the UI will just look broken. And the adjustments would still apply if they tried to use it. Probably better that they know what it does. Also, if originally they set the letter spacing for a different language at least they'd know where to go to change it back. That's just my thoughts.. I think long term we'd also need to know what languages not to apply settings too, as the letter spacing could already be part of their prefs set.

BlueSlug:
@jobara ready for further review! :)

jobara:
merged at 780f34d7b213d43a9d2bd4ba561f23c74a47aaa9

PR #922, opened 2018-08-20
FLUID-6314: Fixing versioned infusion namespace.

jobara:
https://issues.fluidproject.org/browse/FLUID-6314

### DIFF
--- a/src/framework/core/js/ModelTransformationTransforms.js
+++ b/src/framework/core/js/ModelTransformationTransforms.js
@@ -13,7 +13,6 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 */
 
 var fluid_3_0_0 = fluid_3_0_0 || {};
-var fluid = fluid || fluid_3_0_0;
 
 (function ($, fluid) {
     "use strict";
--- a/src/framework/core/js/ModelTransformationTransforms.js
+++ b/src/framework/core/js/ModelTransformationTransforms.js
@@ -12,8 +12,8 @@ You may obtain a copy of the ECL 2.0 License and BSD License at
 https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 */
 
-var fluid_2_0_0 = fluid_2_0_0 || {};
-var fluid = fluid || fluid_2_0_0;
+var fluid_3_0_0 = fluid_3_0_0 || {};
+var fluid = fluid || fluid_3_0_0;
 
 (function ($, fluid) {
     "use strict";
--- a/src/framework/core/js/FluidDebugging.js
+++ b/src/framework/core/js/FluidDebugging.js
@@ -16,7 +16,6 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 */
 
 var fluid_3_0_0 = fluid_3_0_0 || {};
-var fluid = fluid || fluid_3_0_0;
 
 (function ($, fluid) {
     "use strict";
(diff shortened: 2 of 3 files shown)


jobara:
@cindyli could you please review this PR?

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -12,8 +12,8 @@ You may obtain a copy of the ECL 2.0 License and BSD License at
 https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
 */
 
-var fluid_2_0_0 = fluid_2_0_0 || {};
-var fluid = fluid || fluid_2_0_0;
+var fluid_3_0_0 = fluid_3_0_0 || {};
+var fluid = fluid || fluid_3_0_0;
```
cindyli: This line seems unnecessary as `fluid_3_0_0` is passed as an argument at line 992, which is accepted as a name `fluid` at line 18.

Looking thru the code base, currently there are 2 ways to scope a fluid version:
1. [Pre-define 2 global vars](https://github.com/fluid-project/infusion/blob/dbe793d88d23ef8e8cad1032875a39b2d512c0db/src/framework/core/js/ModelTransformationTransforms.js#L15-L18) of `fluid_3_0_0` and `fluid` as in this file;
2. [Only define a global var `fluid_3_0_0`](https://github.com/fluid-project/infusion/blob/dbe793d88d23ef8e8cad1032875a39b2d512c0db/src/components/inlineEdit/js/InlineEdit.js#L17-L19) which is re-scoped by passing in as a function parameter when being used.

I wonder if we should take the chance to sync up the 2 usage? The option 2 looks simpler and more reasonable to me.

What do you think?

jobara: I believe the reason for `var fluid = fluid || fluid_3_0_0` is if "fluid" itself isn't defined. In that case it becomes the fluid_3_0_0 object. I'm not sure it is necessary outside of fluid.js though.

cindyli: Why does `fluid` need to be defined as it [gets passed in as an argument](https://github.com/fluid-project/infusion/blob/dbe793d88d23ef8e8cad1032875a39b2d512c0db/src/framework/core/js/ModelTransformationTransforms.js#L18) with [the value `fluid_3_0_0`](https://github.com/fluid-project/infusion/blob/dbe793d88d23ef8e8cad1032875a39b2d512c0db/src/framework/core/js/ModelTransformationTransforms.js#L992)?

amb26: One definition is outside the closure, and the other one is inside. 

Yes, historically there were two styles in order to reflect the different requirements of components which consumed the framework, but were versioned along with it (dubious in any case - since it should be possible to version components separately from the framework) such as InlineEdit, and on the other hand files like ModelTransformationTransforms.js which actually implemented the framework.

Things became a little further confused because for a while there were multiple profiles of the framework - e.g. "Fluid.js without FluidIoC.js" etc. I believe it is now the case that it is impossible to consume the framework without having first loaded Fluid.js and so we should be able to revert the more complex (two-line) boilerplate in every case except for that one.

jobara:
I've removed the `var fluid = fluid || fluid_3_0_0` statement from files that require Fluid.js. There are however a few files that can be loaded independently: loadScripts.js, jquery.keyboard-a11y.js, and jquery.standalone.js. And, of course, Fluid.js still requires that statement.

cindyli:
Merged at db4f4fb1fb06df0003e11fc90b237e54fb310841

PR #923, opened 2018-08-20
FLUID-6326 - Use node 10.x and npm 6.x

gtirloni:
Please don't merge this until the ansible-nodejs repository points the LTS branch to node 10.x as well, which should happen only when node 10.x becomes a LTS branch and 11.x is released too.


gtirloni:
ok to test

waharnum:
@amb26 - you asked about the status of this PR a few weeks back and I've now had a chance to look into it (sorry for slowness).

The Ansible role depends upon packages published by NodeSource, which hasn't yet released a package for the 11.x line. [According to this issue they will do so shortly](https://github.com/nodesource/distributions/issues/744). I will calendar a reminder to check for the 11.x release from NodeSource on a regular basis until they've done so, and update the Ansible role at that time. This PR should then be mergeable.

amb26:
It's the 10.x line we need, rather than 11.x

waharnum:
@amb26 Understood - to be clearer, the underlying Ansible role that manages this has options of `current` and `lts` that are supposed to map to Node's terminology. There are two outstanding issues with updating the Ansible role:

1. Node still has the `8.12.0` release designated as `lts`, while `11.0` is current, per https://nodejs.org/en/download/
2. Our CentOS package source (NodeSource) does not yet have a package released for `11.0`

If it's not urgent I'd like to wait for both of those issues to be resolved before updating the role. [It appears 10.x will become LTS on October 31, by the chart here](https://github.com/nodejs/Release), and NodeSource will ideally have packages for 11.x at around the same time.

gtirloni:
I've reverted the commits in this PR now that LTS 10.x has been released.

See https://github.com/idi-ops/ansible-nodejs/pull/42

PR #927, opened 2018-08-22
Add SKIP_NODE_MODULES_BIND_MOUNT env var to Vagrantfile

gtirloni:
In certain situations, it's perfectly fine to have the node_modules shared between host and a vagrant box:

* When host/VM architecture are the same
* When all work happens inside the VM

That is the situation in CI Linux build nodes.

This has the advantage that certain operations like publishing the build artifacts can happen outside the VM (as they need access to credentials that would be hard to import into the VM) without having to run `npm install` a second time.

This PR adds an environment variable that allow CI (and Linux users) to avoid the bind mount during the `vagrant up` phase by specifying SKIP_NODE_MODULES_BIND_MOUNT=1.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -30,11 +30,16 @@ Vagrant.configure(2) do |config|
   config.vm.synced_folder ".", "#{app_directory}"
 
   # Mounts node_modules in /var/tmp to work around issues in the VirtualBox shared folders
-  config.vm.provision "shell", run: "always", inline: <<-SHELL
-    sudo mkdir -p /var/tmp/#{app_name}/node_modules #{app_directory}/node_modules
-    sudo chown vagrant:vagrant -R /var/tmp/#{app_name}/node_modules #{app_directory}/node_modules
-    sudo mount -o bind /var/tmp/#{app_name}/node_modules #{app_directory}/node_modules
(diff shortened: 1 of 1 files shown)


gtirloni:
@amatas a suggestion: it would be nice if the vagrant-ci-plugin could have something like this so we don't need to change all Vagrantfiles. Maybe a boolean to decide if it should be done or not and some additional vars for specifying where exactly it should be mounted. Thanks.

jobara:
@gtirloni should we merge this PR or wait for changes to vagrant-ci-plugin. Also, is there a JIRA for this?

amatas:
Thanks @gtirloni for the suggestion. It seems that the bind mount of node_modules directory is a very specific hack to avoid problems with FS operations made by npm only on Linux, so I don't know how to implement this taking into account all the possible parameters.
It seems a similar hack to the use of the samba mounted "v:" drive that we use in the Windows boxes. That hack is implemented in the box itself.

@jobara I would merge this PR, the code seems right and the improvement of the vagrant plugin will delay some more time.

jobara:
merged at c62b79db0ebe0eb669f1e4011cbee0f2e90b8727

PR #928, opened 2018-08-22
FLUID-6328: Add missing files to lint checks and fix linting errors.

the-t-in-rtf:
See [FLUID-6328](https://issues.fluidproject.org/browse/FLUID-6328) for details.

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,16 +1,16 @@
 {
-  "include": [
-     "**/*.js"
-  ],
-  "exclude": [
-    "tests",
-    "reports",
-    "coverage",
-    "examples",
-    "Gruntfile.js"
-  ],
-  "reporter": "none",
-  "report-dir": "reports",
-  "temp-directory": "coverage",
-  "clean": false
+    "include": [
+        "**/*.js"
+    ],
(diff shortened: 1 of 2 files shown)


the-t-in-rtf:
@jobara, I figured out the discrepancy with the JSON files, a few of the JSON files included with jQuery were being linted.  I adjusted the rules to fix this.

jobara:
@the-t-in-rtf thanks for the update. I used `grunt lint --verbose`, as you mentioned in the channel yesterday, to manually verify which files were linted. The only thing that looks to be missing is the [.nycrc](https://github.com/fluid-project/infusion/blob/master/.nycrc) file which appears to be JSON although without the .json extension. I think we should still add it to the json linting though.

There are some of our json dependency files and version files in the lib directories that were written by us and should probably be listed. However, I'm likely going to make some changes in this area for [FLUID-6253](https://issues.fluidproject.org/browse/FLUID-6253) so I think we can hold off on adding them in this PR.

the-t-in-rtf:
I have added the `.nycrc` file to the JSON linting and fixed its indentation.

PR #931, opened 2018-09-10
FLUID-6338: Moving buildStylus from postInstall to prepare.

jobara:
This should prevent it from trying to run buildStylus when installed as 
a dependency of another package.

Should also using Infusion as a dependency in another project using 3.0.0-dev.20180913T134630Z.1dc9abaf4.FLUID-6338

https://issues.fluidproject.org/browse/FLUID-6338

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -455,7 +455,8 @@ module.exports = function (grunt) {
             "distributions" + ( target ? ":" + target : "" ),
             "cleanForDist",
             "verifyDistJS",
-            "verifyDistCSS"
+            "verifyDistCSS",
+            "buildStylus" // put back stylus files needed for development
         ];
         grunt.task.run(tasks);
     });
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "repository": "git://github.com/fluid-project/infusion.git",
     "main": "./src/module/fluid.js",
     "scripts": {
-        "postinstall": "npm run buildStylus",
+        "prepare": "npm run buildStylus",
         "prepublishOnly": "npm run buildDists",
         "pretest": "node node_modules/rimraf/bin.js reports/* coverage/*",
         "test": "npm run test:browser && npm run test:node",
--- a/provisioning/vars.yml
+++ b/provisioning/vars.yml
@@ -6,8 +6,6 @@ nodejs_app_name: infusion
 
 nodejs_branch: lts
 
-nodejs_npm_version: 3.10.10
-
 nodejs_app_npm_packages:
   - testem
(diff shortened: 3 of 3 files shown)


jobara:
ok to test

waharnum:
ok to test

waharnum:
ok to test

waharnum:
@jobara The separated panel tests appear to be failing with this change (https://ci.fluidproject.org/job/infusion-pr-browser-tests/37/console) - is `prepare` expected to run following an NPM install automatically?

jobara:
@waharnum yes, it should be the same as the old prepublish script "prepare: Run both BEFORE the package is packed and published, and on local npm install without any arguments (See below). This is run AFTER prepublish, but BEFORE prepublishOnly." ( https://docs.npmjs.com/misc/scripts )

jobara:
ok to test

jobara:
@waharnum I did a fresh clone on my own machine and ran `npm install`. The stylus files were built. I ran the tests as well using `npm test` and they passed.

jobara:
ok to test

jobara:
ok to test

jobara:
@waharnum, @avtar I wonder if the issue is related to this https://github.com/npm/npm/issues/19684

waharnum:
@jobara @avtar I'll have to look into this more closely at the VM level - I believe the PR Builder plugin uses commit hashes rather than branches, which does work according to the NPM issue as I'm reading it. Will try to investigate today.

waharnum:
ok to test

waharnum:
I've jumped into the VM on the current build (https://ci.fluidproject.org/job/infusion-pull-request/47/) and manually run `grunt buildStylus` - will see if this makes a difference.

waharnum:
Doesn't seem to have made a difference, investigating the Vagrant setup locally now.

waharnum:
Tests on this branch are passing from Vagrant locally, both within the VM directly and from outside it via `npm run test:vagrantBrowser` - doing a wipe on the Jenkins workspace to make sure it's starting from fresh and trying again.

waharnum:
ok to test

jobara:
@waharnum I think I've found the problem. I used the vagrant setup that comes with the repo and found that it also wasn't working. Took a few tries because I had to remember to clear out the existing built css before running vagrant up. What I've found is the VM is running NPM v3.10.10. However, the prepare script wasn't added till NPM v4. How can we update NPM in the VM?

cindyli:
Merged at 4a66b154e7728fe09d012cb2327b2d4d23eb5d35

PR #932, opened 2018-09-11
FLUID-6339: Expanded range for adjuster values to decrease size/space.

jobara:
Also updated the overview panel css to work with the decreased 
letter-spacing.

https://issues.fluidproject.org/browse/FLUID-6339


jobara:
ok to test

PR #933, opened 2018-09-13
FLUID-6257: Successful writes update remote model path from local model path

jobara:
Removes the temporary fix, and now requires that writes resolve/reject its promise after the write operation has been handled and not simply after the request has been sent.

https://issues.fluidproject.org/browse/FLUID-6257


jobara:
@amb26 or @cindyli would you be able to review this PR

src/framework/core/js/RemoteModel.js
```diff
@@ -68,9 +68,14 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                 listener: "fluid.remoteModelComponent.unblockFetchReq",
                 args: ["{that}"]
             },
+            "afterWrite.updateRemoteModel": {
+                changePath: "remote",
+                value: "{that}.model.local"
```
amb26: Seems awkward - what if the change included a DELETE?

PR #935, opened 2018-09-18
FLUID-5633: Fix and test case for further deregistration failures, and for fluid.queryIoCSelector

amb26:
Hi @BlueSlug - this should fix the issue(s) you encountered when writing your test case attached recently to https://issues.fluidproject.org/browse/FLUID-5633


BlueSlug:
This fix does seem to solve the error I was seeing. There are still issues with the test sequence I was attempting to use, but these are beyond the scope of this Jira. Thanks! :)

PR #938, opened 2018-10-10
FLUID-6298: Updating and expanding contrast themes.

jobara:
- Added a grey on dark theme
- expanded grey on white and black on brown themes to include additional 
colours for links, buttons, and selection. This is to match Windows 
Contrast themes.

https://issues.fluidproject.org/browse/FLUID-6298


jobara:
@jhung could you please take a look at this PR? There are a lot of Styling changes.

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -125,19 +151,45 @@ build-themes-Enactors(contrastThemes) {
...
 
+            if (selectedFColor && selectedBColor) {
+                &::selection,
+                ::selection {
+                    color: selectedFColor !important;
+                    background-color: selectedBColor !important;
+                }
+            }
+
+            if (disabledColor) {
```
simonbates: Minor thing, but I think I would swap the order here of the `if (selectedFColor...` and `if (disabledColor...` blocks so that they match the order of specification of the values above beginning line 136.

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -301,17 +350,44 @@ build-themes-Enactors(contrastThemes) {
                     border-left-color: bColor !important;
                 }
 
+                if (selectedFColor && selectedBColor) {
```
simonbates: Would we want an entry in the `.fl-inverted-color` block for `disabledColor`?

demos/prefsFramework/index.html
```diff
@@ -418,12 +418,14 @@ <h1 class="ocean">Climate Change Impacts</h1>
...
                         <img class="icon" style="padding-right: 5px; float: right;" src="images/theme_climate_90.gif" alt="Oceans and Coasts" height="90" width="90">
                         <aside id="left_box">
-
-                            <div class="left_tab"></div>
-                            <div class="middle_tab">
-                                <h2 class="h1_middle_text fl-inverted-color">Resources</h2>
+                            <div class="fl-inverted-color">
+                                <div class="left_tab"></div>
+                                <div class="middle_tab">
+                                    <h2 class="h1_middle_text fl-inverted-color">Resources</h2>
```
simonbates: With the new `fl-inverted-color` container `div`, did we also want to keep the `fl-inverted-color` class here on the `h2`? We now have that class in 2 places: on a container and on the element.

src/framework/preferences/css/stylus/utils/Themes.styl
jobara: My initial thought is that it isn't needed, but I'm open to changing things. If I remember correctly the `fl-inverted-color` class was provided as a means to make certain portions of the content to stick out. This is particularly necessary for our binary colour schemes. In the prefs framework demo we use this for a header. If you were to have a whole section marked as `fl-inverted-color`, which I can't think of a reason for at the moment, I think you'd still want the disabled inputs to look the same for consistency.

Now that i'm thinking more about this, i'm not sure that it's necessary to invert the selection styles either. 

Please let me know what you think about these.

tests/manual-tests/framework/preferences/assortedContent/index.html
```diff
@@ -291,6 +291,119 @@ <h4>Mid-term Exam</h4>
                             <textarea>Please enter a simple JavaScript function to find the smallest value of an array.</textarea>
                             <button>Run Script</button>
                         </fieldset>
+                        <fieldset class="disabled-inputs">
+                            <legend>HTML Property</legend>
```
simonbates: I think that we are either missing a layer of `fieldset` here or we need to change this `legend`. At the moment it looks like the whole disabled inputs `fieldset` is called "HTML Property" rather than something like "Disabled Inputs".

jobara:
@simonbates I've addressed the issues we talked about earlier today. This PR is ready for more review.

PR #939, opened 2018-10-15
FLUID-6330: Improvements to logging system

amb26:


src/framework/core/js/Fluid.js
```diff
@@ -1479,8 +1477,14 @@ var fluid = fluid || fluid_3_0_0;
         var that;
 
         var lazyInit = function () { // Lazy init function to economise on object references for events which are never listened to
+            // The authoritative list of all listeners, a a hash indexed by namespace, looking up to a stack (array) of
```
jobara: typo "a a hash"

src/module/fluid.js
```diff
@@ -94,12 +94,20 @@ fluid.renderLoggingArg = function (arg) {
     return togo;
 };
 
-// Monkey-patch the built-in fluid.doLog utility to improve its behaviour within node.js - see FLUID-5475
-fluid.doLog = function (args) {
-    args = fluid.transform(args, fluid.renderLoggingArg);
+// Improve the behaviour of onkey-patch the built-in fluid.doLog utility to improve its behaviour within node.js - see FLUID-5475
```
jobara: looks like a typo "onkey-patch" but wording in general seems to be a bit off; may be missing other words or changes.

src/framework/core/js/Fluid.js
```diff
@@ -1741,6 +1744,24 @@ var fluid = fluid || fluid_3_0_0;
...
+    fluid.loggingEvent = fluid.makeEventFirer({name: "logging event"});
+
+    fluid.addTimestampArg = function (args) {
+        var arg0 = fluid.renderTimestamp(new Date()) + ":  ";
+        args.unshift(arg0);
+    };
+
+    fluid.loggingEvent.addListener(fluid.doBrowserLog, "log");
+    // A placeholder for any actions filtering the log arguments before dispatching
+    fluid.loggingEvent.addListener(fluid.identity, "filterArgs", "before:log");
```
jobara: I assume that you're meant to replace this for the centralized logging. I think it would probably be useful to have a test to verify that it works and to demonstrate how it would be replaced.

amb26: No, the filterArgs just represents a position with the chain so that one can write after:filterArgs rather than writing before:log which might be more ambiguous. I'll improve the comment wording

src/module/fluid.js
amb26: Sorry about this, some junk got inserted here as the result of a partial "undo" : P

tests/framework-tests/core/js/FluidJSTests.js
```diff
@@ -718,6 +718,31 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+        fluid.log("Zis guy");
+        // Slice to remove the timestamp argument unshifted by the standard interceptor
+        jqUnit.assertDeepEq("Logged to memory with interception", ["Zis guy", 42],
+            fluid.tests.memoryLog[0].slice(1));
+        fluid.loggingEvent.removeListener(fluid.tests.doMemoryLog);
+        fluid.loggingEvent.removeListener("42");
+        var listeners = fluid.getMembers(fluid.loggingEvent.sortedListeners, "listener");
+        jqUnit.assertFalse("Intercepting listener removed", listeners.includes(fluid.tests.insert42));
+        jqUnit.assertFalse("Memory log listener removed", listeners.includes(fluid.tests.doMemoryLog));
+        jqUnit.assertTrue("Browser log listener restored", listeners.includes(fluid.doBrowserLog));
```
jobara: These tests fail in IE 11 because `includes` isn't supported in that browser. ( https://caniuse.com/#feat=array-includes ).

PR #940, opened 2018-10-17
FLUID-6346: Fixes for fluid.isArrayable

amb26:


src/framework/core/js/Fluid.js
```diff
@@ -352,8 +354,11 @@ var fluid = fluid || fluid_3_0_0;
      * @param {Any} totest - The value to be tested
      * @return {Boolean} `true` if the supplied value is an array
      */
+    // Note: The primary place jQuery->Array conversion is used in the framework is in dynamic components with a jQuery source.
     fluid.isArrayable = function (totest) {
-        return totest && (totest.jquery || Object.prototype.toString.call(totest) === "[object Array]");
+        return (Boolean)(totest && (Object.prototype.toString.call(totest) === "[object Array]"
+            || totest.jquery && totest.constructor && totest.constructor.prototype
+               && totest.constructor.prototype.jquery));
```
jobara: It's a bit difficult to read. I wonder if we could either just use [`instanceOf`](https://stackoverflow.com/a/1853246/10198026) or make a new function that determines if something is a jQuery and make use of that here.

amb26: Sure. We can't use instanceOf because jQuery might not be loaded at all (if we are on the server)

amb26: As well as, of course, the possibility that the jQuery object originates from a different iframe

src/framework/core/js/Fluid.js
```diff
@@ -345,15 +347,26 @@ var fluid = fluid || fluid_3_0_0;
...
 
+    /** Determines whether the supplied object is an jQuery object. The strategy uses optimised inspection of the
+     * constructor prototype since jQuery may not actually be loaded
+     * @param {Any} totest - The value to be tested
+     * @return {Boolean} `true` if the supplied value is a jQuery object
+     */
+    fluid.isJQuery = function (totest) {
+        return Boolean(totest && totest.jquery && totest.constructor && totest.constructor.prototype
+               && totest.constructor.prototype.jquery);
+    };
```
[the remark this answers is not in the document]
amb26: We won't encounter the object, but we still need to make the fluid.isArrayable check.

PR #942, opened 2018-10-24
FLUID-6348: Placing the play button at the top of the stack.

jobara:
https://issues.fluidproject.org/browse/FLUID-6348

### DIFF
--- a/src/components/orator/css/Orator.css
+++ b/src/components/orator/css/Orator.css
@@ -73,6 +73,9 @@
     border-radius: 0.2rem;
     /* Align down arrow to the beginning of the character that is first selected. */
     transform: translate(-49%);
+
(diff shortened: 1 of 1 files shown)


src/components/orator/css/Orator.css
```diff
@@ -73,6 +73,9 @@
     border-radius: 0.2rem;
     /* Align down arrow to the beginning of the character that is first selected. */
     transform: translate(-49%);
+
+    /* force the play button to the top of the stacking order (see: https://issues.fluidproject.org/browse/FLUID-6348) */
+    z-index: 9007199254740992;
```
jhung: This isn't an arbitrary number, so we should probably define where the value is defined / comes from.

PR #944, opened 2018-11-16
FLUID-6240: Syllabification preference

jobara:
https://issues.fluidproject.org/browse/FLUID-6240

also includes fixes for https://issues.fluidproject.org/browse/FLUID-6354


jobara:
@amb26 would you be able to review this PR?

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -0,0 +1,386 @@
...
+        that.hyphenators[langSegs[0]] = hyphenatorPromise;
+        return promise;
+    };
+
+    fluid.prefs.enactor.syllabification.syllabify = function (that, node, lang) {
+        var hyphenatorPromise = that.getHyphenator(lang);
+        hyphenatorPromise.then(function (hyphenator) {
+            if (hyphenator) {
+                // remove \u200B characters added hyphenateText
+                var hyphenated = hyphenator.hyphenateText(node.textContent).replace(/\u200B/gi, "");
```
amb26: This is too vital a part of policy to be hidden away inside a promise callback. The body should be a public function, possibly with a bit of paramterisation

package.json
```diff
@@ -53,20 +53,22 @@
         "grunt-contrib-watch": "1.1.0",
         "grunt-modulefiles": "0.4.0",
         "grunt-shell": "2.1.0",
+        "hypher": "0.2.5",
```
amb26: When you get back from your leave we should start thinking about modularization of Infusion quite urgently since these are far too many dependencies for those who just to get access to the core framework

README.md
```diff
@@ -204,6 +204,8 @@ grunt custom --name="myPackage"
 #### External Libraries
 
 * fastXmlPull
+* fonts
```
amb26: Which fonts? What library does this refer to? We should really add links to these list elements and also revisit the text at the base which says "All of these libraries are already bundled within the Infusion image."

demos/prefsFramework/js/prefsEditorDemo.js
```diff
@@ -102,6 +102,20 @@ var demo = demo || {};
...
     });
 
+    // Fine-tune the syllabification schema
+    fluid.defaults("demo.prefsEditor.auxSchema.syllabification", {
+        gradeNames: ["fluid.prefs.auxSchema.syllabification"],
+        auxiliarySchema: {
+            syllabification: {
+                enactor: {
+                    terms: {
+                        patternPrefix: "../../src/lib/hypher/patterns"
```
amb26: Remind me if we have any scheme for managing these kinds of relative file paths or whether the user is expected to override them for each enactor separately

ReleaseNotes.md
```diff
@@ -57,8 +57,13 @@ categorized by license:
 
 ### BSD-3
 
+* [Hypher v0.2.5](https://github.com/bramstein/Hypher)
 * [Sinon v6.1.3](http://sinonjs.org)
 
+### LGPL
```
amb26: @colinbdclark - can we manage this?

colinbdclark: Yup, LGPL is not a problem—especially if we're managing this dependency via npm, rather than including it directly in our source tree.

demos/prefsFramework/js/prefsEditorDemo.js
jobara: Sadly it is the integrators responsibility to override them.

README.md
jobara: The fonts are referring to [open-sans](https://github.com/fluid-project/infusion/blob/master/package.json#L62). 

When you say add links, what would you like to point at? The same place we point links to for the libraries in the release notes? or somewhere else?

For "All of these libraries are already bundled within the Infusion image." Maybe we should just remove that line. I'm not sure it's really adding anything anyways.

package.json
jobara: Yes, we should look more into modularization. Even being able to break up the prefs framework into smaller modules would be useful. We should try to sketch out what we want for this again, or add to what we already have if there are more/new requirements. 

Regarding the above dependencies, they are all listed as dev dependencies so they aren't pulled in with NPM installs into other projects. Also, they can be omitted from various builds of infusion via the custom builds. For example, hypher wouldn't be included in the "framework" only builds.

README.md
amb26: Well we should remove the line since it is false :)
We can just rewrite the line "fonts" to read opensans-webkit since then readers will have some guess what this refers to

package.json
amb26: I can understand the reasoning for listing them as dev dependencies, but it is essentially misleading since they are expected to be used by the implementation. For example, in use cases such as chart authoring - https://github.com/fluid-project/chartAuthoring/blob/master/package.json#L11 - Infusion is consumed as a client-side dependency. Whilst the consumer could just bind to "dist", they might want to bind at the source level for debugging purposes. In that case they'd have to replicate the setup and package.json entries for hypher etc. Whilst this is an edge use case, it is part of the set of issues we'll need to consider when modularising.

src/framework/core/js/MutationObserver.js
```diff
@@ -0,0 +1,124 @@
...
+            createObserver: {
+                funcName: "fluid.mutationObserver.createObserver",
+                args: ["{that}"]
+            }
+        }
+    });
+
+    /**
+     * A Mutation Observer; allows for tracking changes to the DOM.
+     * A mutation observer is created with a callback function, configured through its observe` method.
```
amb26: Stray backtick

src/framework/core/js/MutationObserver.js
```diff
@@ -0,0 +1,124 @@
...
+     * Instantiates a mutation observer, defining the callback function which relays the observations to component
+     * events. The configuration passed to the observe function will determine what mutations are observed and what
+     * information is returned. See `fluid.mutationObserver.observe` about configuring the mutation observer.
+     *
+     * Event Mapping:
+     * onNodeAdded - fired for each added node, includes the node and mutation record
+     * onNodeRemoved - fired for each removed node, includes the node and mutation record
+     * onAttributeChanged - fired for each attribute change, includes the node and mutation record
+     *
+     * @param {Component} that - an instance of `fluid.mutationObserver
```
amb26: Stray backtick

package.json
jobara: The lib files are actually included in the NPM package ( https://unpkg.com/infusion@3.0.0-dev.20181120T213100Z.8ef92ceac.FLUID-6240/src/lib/ ). They are just ignored from git. This, for example, is how we get access to the language pattern files in UIO+.

README.md
jobara: I guess it was fonts because we were expecting it to possibly have more, but i can change that up.

jobara: I found at least one more font that should be pulled in by NPM, Roboto, and will also update them to use woff instead of ttf ( see: https://issues.fluidproject.org/browse/FLUID-6354 ).

jobara: @amb26 I've made a bunch of improvements to the sourcing of fonts, including icon fonts. I've also switched, where possible, fonts to woff as these are intended to be more performant on the web. Please not that my update to the overview panel icons is based off of changes I've made in https://github.com/fluid-project/infusion-icons/pull/7

jobara:
@amb26 this is ready for more review.

amb26:
Test failure on IE11:
````
not ok 118 IE 11.0 - [943 ms] - Framework Tests: /tests/framework-tests/preferences/html/SyllabificationEnactor-test.html
    ---
        actual: >
            2
        expected: >
            1
        stack: >
               at runLoggingCallbacks (http://localhost:7357/tests/lib/qunit/js/qunit.js:1610:4)
               at push (http://localhost:7357/tests/lib/qunit/js/qunit.js:959:3)
               at assert.strictEqual (http://localhost:7357/tests/lib/qunit/js/qunit.js:614:3)
               at jsUnitCompat.assertEquals (http://localhost:7357/tests/test-core/jqUnit/js/jqUnit.js:137:13)
               at Anonymous function (http://localhost:7357/tests/framework-tests/preferences/js/SyllabificationEnactorTests.js:294:13)
               at fluid.each (http://localhost:7357/src/framework/core/js/Fluid.js:1:230580)
               at fluid.tests.syllabificationTester.verifyUnsyllabified (http://localhost:7357/tests/framework-tests/preferences/js/SyllabificationEnactorTests.js:292:9)
        message: >
            fluid.prefs.enactor.syllabification: Add/Remove syllabification: Syllabification Removed: .flc-syllabification-parentLang should
 only have one child node - at sequence position 4 of 4
        Log: |
    ...
````

jobara:
@amb26 thanks for catching this. It seems that Node.normalize() doesn't work properly in IE 11. https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8727426/ 

I'll investigate an alternative approach for IE 11.

jobara:
@amb26 I've written my own normalize function which should now work across browsers.

PR #945, opened 2018-11-27
FLUID-6318: Updating the selection reader control to play and stop.

jobara:
https://issues.fluidproject.org/browse/FLUID-6318

### DIFF
--- a/src/components/orator/js/Orator.js
+++ b/src/components/orator/js/Orator.js
@@ -160,17 +160,13 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             },
             toggle: {
                 funcName: "fluid.orator.controller.toggleState",
-                // Not providing an option for the explicit state value because
-                // when called through the jQuery click event the event object
-                // is passed in.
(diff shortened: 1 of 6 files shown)


src/components/orator/js/Orator.js
```diff
@@ -847,6 +856,10 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             stop: {
                 funcName: "fluid.orator.selectionReader.stopSpeech",
                 args: ["{that}.model.play", "{fluid.textToSpeech}.cancel"]
+            },
+            toggle: {
+                funcName: "fluid.orator.selectionReader.togglePlay",
```
cindyli: Understand the `state` argument is implicitly passed in via jQuery click event. But it would be more readable and reusable to include the pass-in of `{arguments}.0` on the argument list.

The same comment applies to the other `toggle()` function at line 161.

jobara:
@cindyli I've addressed your comment. Ready for more review.

PR #946, opened 2018-11-28
FLUID-6240: linting

jobara:
Fixes a small linting issue that was breaking the builds.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -365,7 +365,7 @@ module.exports = function (grunt) {
             sources: {
                 md: [ "*.md", "!./src/**/lib/**/*.md", "!./demos/**/lib/**/*.md", "!./tests/**/lib/**/*.md"],
                 js: ["*.js", "!./src/**/lib/**/*.js", "!./demos/**/lib/**/*.js", "!./dist/**/*.js", "!./tests/**/lib/**/*.js", "!./tests/**/infusion-1.5.js"],
-                json: ["*.json", "./.nycrc", "!./src/**/lib/**/*.json", "!./dist/**/*.json"],
+                json: ["*.json", "./.nycrc", "./src/thirdPartyDependencies.json", "!./src/lib/**/*.json", "!./dist/**/*.json"],
                 other: ["./.*"]
             }
         },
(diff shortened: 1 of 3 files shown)


jobara:
@amb26 could you please take a look at this PR. It seems a linting issue snuck in, and is preventing publishing to npm.

Gruntfile.js
```diff
@@ -365,7 +365,7 @@ module.exports = function (grunt) {
             sources: {
                 md: [ "*.md", "!./src/**/lib/**/*.md", "!./demos/**/lib/**/*.md", "!./tests/**/lib/**/*.md"],
                 js: ["*.js", "!./src/**/lib/**/*.js", "!./demos/**/lib/**/*.js", "!./dist/**/*.js", "!./tests/**/lib/**/*.js", "!./tests/**/infusion-1.5.js"],
-                json: ["*.json", "./.nycrc", "!./src/**/lib/**/*.json", "!./dist/**/*.json"],
+                json: ["*.json", "./.nycrc", "./src/thirdPartyDependencies.json", "!./src/lib/**/*.json", "!./dist/**/*.json"],
```
cindyli: Would it be more sensible to have `thirdPartyDependencies.json` sitting in `src/lib/`directory rather than `src/` directory?

Does `"!./src/lib/**/*.json"` mean to exclude json files in `./src/lib/` from linting? It doesn't seem to work. I experimented with having a wrong formatted json file in `src/lib/fastXmlPull/` and running `grunt buildDists` failed at reporting linting error for this file.

jobara: `"!./src/lib/**/*.json"` means that the JSON files under "lib" will not be linted. Because of this, the `thirdPartyDependencies.json` file was also not being linted. The only two ways I could see to get this to work would be 1) list out each lib subdirectory to excluded from linting 2) move the thirdPartyDependencies.json file out of lib. While it is nice to have the dependency file grouped with the lib directories, it seemed the better trade off than having to update the linting config every time we added or removed third party dependencies.

cindyli: After a chat with @jobara, `"!./src/lib/**/*.json"` does exclude jsons files from "src/lib/" directory from linting with running `grunt lint`. Running `grunt buildDists` fails is because the file I copied into `src/lib/fastXmlPull/` is named "thirdPartyDependencies.json" that is in dependency file naming convention and the build script tries to read dependencies from it and fails.

cindyli:
Merged at beaa1c76305d89d2f30b0c4ed3f537689cf0b056

PR #948, opened 2018-12-03
FLUID-6357: Moving aria-expanded to the button instead of the panel.

jobara:
This is to conform with the aria spec. See: 
https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element

https://issues.fluidproject.org/browse/FLUID-6357


jobara:
@waharnum could you please verify that this PR addresses the JIRA?

waharnum:
@jobara Checked with Lisa, she says this looks correct to her.

cindyli:
Merged at 2ca90f6e6db043849826dd09418824aac24394f6

PR #949, opened 2018-12-20
FLUID-6303: Adding a localization preference

jobara:
https://issues.fluidproject.org/browse/FLUID-6303

Adding an adjuster and enactor for localization. The enactor supports applying localization by modifying the URL path or by hooking into the `lang` model path. An enactment using query parameter has not been implemented yet as currently there isn't a use case for it. An example, that enacts with URL path change, has been provided.


examples/framework/preferences/localizationPreference/urlPath/fr/index.html
```diff
@@ -0,0 +1,105 @@
...
+
+        <!-- BEGIN markup for Preferences Editor -->
+        <div class="flc-prefsEditor-separatedPanel fl-prefsEditor-separatedPanel">
+            <!--
+                This div is for the sliding panel bar that shows and hides the Preference Editor controls in the mobile view.
+                A separate panel bar for mobile displays is needed to preserver the correct tab order.
+            -->
+            <div class="fl-panelBar fl-panelBar-smallScreen">
+                <span class="fl-prefsEditor-buttons">
+                    <button class="flc-slidingPanel-toggleButton fl-prefsEditor-showHide"> Show/Hide</button>
```
BlueSlug: "Show/Hide" should be "Montrer/Cacher"

examples/framework/preferences/localizationPreference/urlPath/fr/index.html
```diff
@@ -0,0 +1,105 @@
...
+                    <button class="flc-prefsEditor-reset fl-prefsEditor-reset"><span class="fl-icon-undo"></span> Reset</button>
```
BlueSlug: "Reset" should be "Réinitialiser" or "Remettre"

examples/framework/preferences/localizationPreference/urlPath/fr/index.html
```diff
@@ -0,0 +1,105 @@
...
+            <!-- This is the div that will contain the Preference Editor component -->
+            <div class="flc-slidingPanel-panel flc-prefsEditor-iframe"></div>
+
+            <!--
+                This div is for the sliding panel bar that shows and hides the Preference Editor controls in the desktop view.
+                A separate panel bar for desktop displays is needed to preserver the correct tab order.
+            -->
+            <div class="fl-panelBar fl-panelBar-wideScreen">
+                <span class="fl-prefsEditor-buttons">
+                    <button class="flc-slidingPanel-toggleButton fl-prefsEditor-showHide"> Show/Hide</button>
```
BlueSlug: See previous comments for this and the following line

examples/framework/preferences/localizationPreference/urlPath/fr/index.html
```diff
@@ -0,0 +1,105 @@
...
+                    <button class="flc-prefsEditor-reset fl-prefsEditor-reset"><span class="fl-icon-undo"></span> Reset</button>
```
BlueSlug: ditto

examples/framework/preferences/localizationPreference/urlPath/fr/index.html
jobara: @BlueSlug what do you think would be better to use in this case?

examples/framework/preferences/localizationPreference/urlPath/js/localization.js
```diff
@@ -0,0 +1,96 @@
...
+    "use strict";
+
+    fluid.prefs.builder({
+        gradeNames: ["fluid.prefs.auxSchema.localization"],
+        auxiliarySchema: {
+            "namespace": "example.prefs.localization"
+        }
+    });
+
+    fluid.defaults("example.prefsEditor", {
```
cindyli: I'm not sure how useful this examplary code will help integrators. With all these uses of aux schema namespace, distributeOptions targets etc, writing this level of grade requires a good understanding of the detailed implementation of the prefs framework schema version. But of course, it's great to demonstrate how the localization can be done.

examples/framework/preferences/localizationPreference/urlPath/js/localization.js
```diff
@@ -0,0 +1,96 @@
...
+            },
+            "example.localization.template": {
+                source: "{that}.options.template",
+                target: "{that prefsEditorLoader > templateLoader}.options.resources.prefsEditor"
+            }
+        }
+    });
+
+    /**
+     * Used as an expander to find the langSegIndex. For most implementations this would actually refer to 1, which
```
cindyli: This comment is confusing. The function is to calculate the index of the language code on the url path segments. But why would the index value 1 refers to the default value in most implementations? The default lang code in this example is `null`. I wonder what default lang code you expect in most implementations.

src/framework/preferences/js/LocalizationEnactor.js
```diff
@@ -0,0 +1,152 @@
...
+                "model.lang": "value"
+            }
+        },
+        contextAwareness: {
+            localeChange: {
+                checks: {
+                    inPanel: {
+                        contextValue: "{iframeRenderer}"
+                    },
+                    urlPath: {
```
cindyli: It seems the only localization method that the prefs framework supports is via the url path. If so, is it necessary to detect this case with the context awareness? Also, the default grade `fluid.prefs.enactor.localization.no-op` is not implemented. I'm curious how the localization would behave with this non-implemented grade.

src/framework/preferences/js/LocalizationEnactor.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+    });
+
+    /**
+     * A simple wrapper around the location.pathname setter/getter property.
+     *
+     * @param {String} pathname - (Optional) pathname to set. If not supplied the current pathname will be returned
+     * @return {String} - If the `pathname` argument is not provided, the current pathname is returned
+     */
+    fluid.prefs.enactor.localization.urlPathLocale.pathname = function (pathname) {
```
cindyli: It might be clearer to define setPathName() and getPathName() rather than sharing these 2 functionalities via this single function. It seems the getter is used for setting the initial value at line 70 and the setter is for responding to path change requests at line 83.

examples/framework/preferences/localizationPreference/urlPath/js/localization.js
```diff
@@ -0,0 +1,96 @@
...
+        }
+    });
+
+    fluid.defaults("example.prefsEditor", {
+        gradeNames: ["example.prefs.localization.prefsEditor"],
+        localizationOpts: {
+            enactor: {
+                localizationScheme: "urlPath",
+                langMap: {
+                    "default": null,
```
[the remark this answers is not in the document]
jobara: Please feel free to suggest an alternative. The issue this attempts to address is when no language preference has been made. We don't want the default to be English for example, if that's the case when a user explicitly navigates to a URL like `www.example.com/fr/about/` they would get redirected to the `www.example.com/en/about/`. By having it a no-op, it just stays where it is. Of course the downside of this is that if you had selected a localization preference and then changed to default, nothing would happen.

examples/framework/preferences/localizationPreference/urlPath/js/localization.js
jobara: hmm.. that's interesting.. perhaps there is a helper grade missing that should include many of those things and could be added directly to the schema. I'll have to investigate that.

src/framework/preferences/js/LocalizationEnactor.js
jobara: There are effectively two implemented methods, although an integrator could support additional ones. For example using a query parameter. The two methods supported are URL path and model. The no-op is essentially the model method as there is no additional functionality needed on our part. The integrator would have to hook into the model from the enactor, which is why it uses the "fluid.resolveRoot" grade to make it more easily discoverable outside of the context of the prefs framework's component tree structure.

jobara: I think i just got what yo meant about the no-op grade. I believe that was something i was supposed to have removed from previous work. I've taken it out now.

examples/framework/preferences/localizationPreference/urlPath/js/localization.js
jobara: `default` is a bit of an overloaded term. The "default" preference value is `null`. The "default" configured langSegIndex is `1`. I've tried to modify the comment to make that clearer.

jobara:
@cindyli, I believe I've addressed all of your comments. Ready for more review.

examples/framework/preferences/localizationPreference/urlPath/js/localization.js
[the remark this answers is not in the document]
jobara: From what we discussed at our impromptu mini design crit, it seems that the default option is in fact as confusing as it seems. The suggestions are to change the panel description to be something more like "set your preferred language", change the "default" option to be something like "no selection", and possibly to have a discrete english option and a true no selection default. 

We may also want to consider adding to this example or creating a new one that provides a language selector directly on the page in addition to the one in the panel.

jobara:
@cindyli this is ready for more review.

src/framework/preferences/js/LocalizationEnactor.js
```diff
@@ -35,7 +35,10 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             localeChange: {
                 checks: {
                     inPanel: {
```
cindyli: What's the "inPanel" check for?

src/framework/preferences/js/LocalizationSchemas.js
```diff
@@ -0,0 +1,66 @@
...
+                "templatePrefix": "../../framework/preferences/html/",
+                "messagePrefix": "../../framework/preferences/messages/"
+            },
+            "template": "%templatePrefix/SeparatedPanelPrefsEditor.html",
+            "message": "%messagePrefix/prefsEditor.json",
+            localization: {
+                "type": "fluid.prefs.localization",
+                "enactor": {
+                    "type": "fluid.prefs.enactor.localization",
+                    "classes": "@localization.classes"
```
cindyli: There is not a `localization.classes` block defined in the aux schema. Can this line be removed?

src/framework/preferences/js/LocalizationEnactor.js
jobara: It's to see if we are running in the separated panel's iframe. I'm going to add a comment about that.

src/framework/preferences/js/LocalizationEnactor.js
```diff
@@ -0,0 +1,156 @@
...
+                if (urlLangSeg) {
+                    pathSegs[langSegIndex] = urlLangSeg;
+                } else {
+                    if (langSegIndex === pathSegs.length - 1) {
+                        pathSegs.pop();
+                    } else {
+                        pathSegs.splice(langSegIndex, 1);
+                    }
+                }
+            } else if (urlLangSeg) {
```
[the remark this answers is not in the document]
jobara: This one is for the case where `hasLang` is false and `urlLangSeg` is truthy.

src/framework/preferences/js/LocalizationSchemas.js
jobara: thanks for catching that, I had made changes form some prior work and forgot to remove this.

jobara:
@cindyli thanks for the feedback. Ready for more review.

jobara:
@cindyli when this PR is merged, I think we should file another JIRA to take more time to examine the user experience of setting languages through the panel/adjuster. From our mini design crit there were a lot of view points expressed that made we wonder if the approach taken here is the best. For example there was mention that usually language happens through a toggle directly visible on the page and/or on a screen before hand. Which makes me wonder if a user would ever find it in the panel. We also do have a paradigm at the moment for exposing adjusters outside of the panel. Although I don't think it would be too much work to make that happen. At any rate, I think it warrants more design thought, maybe user testing and/or co-design and a more extensive exploration of how sites currently set language.

demos/prefsFramework/index.html
```diff
@@ -61,11 +61,13 @@
         <script type="text/javascript" src="../../src/framework/preferences/js/Panels.js"></script>
         <script type="text/javascript" src="../../src/framework/preferences/js/SelfVoicingPanel.js"></script>
         <script type="text/javascript" src="../../src/framework/preferences/js/LetterSpacePanel.js"></script>
+        <script type="text/javascript" src="../../src/framework/preferences/js/LocalizationPanel.js"></script>
```
cindyli: The manual test shows the localization panel added for this demo doesn't perform a language change on the page when switching to a non-English localization option.

cindyli:
Merged at 35e8a5a9c958cca65ca403108f784668ff42148d

cindyli:
The pull request https://github.com/fluid-project/infusion/pull/951 is issued based on this pull request to address the last minor code review comment. https://github.com/fluid-project/infusion/pull/951 is merged at 35e8a5a9c958cca65ca403108f784668ff42148d

PR #951, opened 2019-01-23
FLUID-6303: Add a localization preference

cindyli:
This pull request is forked from @jobara's https://github.com/fluid-project/infusion/pull/949 pull request to address [the last minor comment](https://github.com/fluid-project/infusion/pull/949#discussion_r249489976) as @jobara is on vacation.

The change is to remove the localization preference from the prefsFramework demo because the selection change on this preference doesn't perform the language change on the demo page.

This change should be re-visited when @jobara comes back from the vacation.


cindyli:
Merged at 35e8a5a9c958cca65ca403108f784668ff42148d

PR #956, opened 2019-07-17
FLUID-5525: add OpenDyslexic as an option to text-font panel

waharnum:
Add the OpenDyslexic font as an option to the preferences framework text-font panel, based on work to add it to CISL's version.

Sources OpenDyslexic from the NPM package, pinned to latest version.

Resolves (adding, rather than replacing) this JIRA issue: https://issues.fluidproject.org/browse/FLUID-5525


waharnum:
@jobara, can you review this for merge?

jobara:
@waharnum could you also update the release with mention of the license and which version of Open Dyslexic we are using. The 3rd party software is all organized by license in there.

waharnum:
There seems to be an ongoing problem with the CI failing that isn't necessarily related to this PR. I am investigating now.

waharnum:
ok to test

waharnum:
CI tests pass when Headless Chrome is excluded as a browser to use and consistently fail when it is present (hangs forever). Appears to occur on the `SyllabificationEnactor-test` run.

Have opened separate PR to address what I think is an underlying issue in the testem config: https://github.com/fluid-project/infusion/pull/957

src/framework/preferences/js/PrefsEditor.js
```diff
@@ -441,6 +441,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
 
     // called once markup is applied to the document containing tab component roots
     fluid.prefs.prefsEditor.finishInit = function (that) {
+        console.log(that);
```
jobara: stray console.log

jobara:
@waharnum with the OpenDyslexic font being pulled in via NPM we shouldn't need to commit the font files to git. Could you please update any ignore (e.g. .gitignore) as needed. Also we need to makes sure that the fonts are properly added to the dist directory. Perhaps you could also address https://issues.fluidproject.org/browse/FLUID-6361 at the same time. A more general fix for these will be done in https://issues.fluidproject.org/browse/FLUID-6359 but if you can get fonts fixed up ahead of time that would be helpful.

jobara:
@waharnum seems this last build failure was related to linting

waharnum:
@jobara I've done the following re the `dist` directory:

- added OpenDyslexic and OpenSans to what is copied over (so this PR should also address https://issues.fluidproject.org/browse/FLUID-6361 now)
- added tasks to grunt to verify the fonts are present in `dist`, along the lines of the existing JS and CSS verification tasks
- refactored the grunt tasks for file verification to eliminate logic duplication

Gruntfile.js
```diff
@@ -166,7 +167,7 @@ module.exports = function (grunt) {
                 files: [{
                     expand: true,
                     cwd: "build/",
-                    src: ["src/lib/fonts/**", "src/framework/preferences/fonts/**", "src/framework/preferences/images/**"],
+                    src: ["src/lib/fonts/**", "src/framework/preferences/fonts/**", "src/framework/preferences/images/**", "src/lib/open-dyslexic/**", "src/lib/opensans/**"],
```
jobara: I don't think the "src/lib/fonts" directory exists anymore.

jobara:
@waharnum I noticed that the full page instances are having problems. The previews do not display in openDyslexic and the minimal footprint content page doesn't either. 

- infusion/tests/manual-tests/framework/preferences/fullPage/
- infusion/tests/manual-tests/framework/preferences/fullPage-schema/
- infusion/examples/framework/preferences/minimalFootprint/

waharnum:
@jobara This should now be fixed - the `font-face` definitions should have been in the `Enactors.stylus` file rather than the `PrefsEditor.stylus` file - OpenDyslexic was working on the demo page because that includes `PrefsEditor.css` in the page, but those other contexts don't.

src/framework/preferences/css/stylus/Enactors.styl
```diff
@@ -32,3 +32,32 @@ build-themes-Enactors(themes);
...
+    font-weight: normal;
+    font-style: italic;
+}
+
+@font-face {
+    font-family: 'OpenDyslexic';
+    src:url('../../../lib/open-dyslexic/fonts/OpenDyslexic-BoldItalic.woff');
+    font-weight: 600;
+    font-style: italic;
+}
```
jobara: I wonder if this font-face definition should be part of build-fonts-Enactors method, or at least somehow part of Fonts.styl. It seems that if one were to use that stylus helper but not have manually added the font definition, that it wouldn't work properly.

waharnum: This seems reasonable. It could probably set up as a generator-style function as well (though OpenDyslexic would be the only example).

Gruntfile.js
```diff
@@ -652,17 +669,36 @@ module.exports = function (grunt) {
...
+                message: "Verify opendyslexic font files are present in /dist/ directory",
+                distDirectory: "dist/assets/src/lib/open-dyslexic/fonts"
+            },
+            openSans: {
+                cwd: "src/lib/opensans/fonts",
+                pattern: "*.woff",
+                message: "Verify opensans font files are present in /dist/ directory",
+                distDirectory: "dist/assets/src/lib/opensans/fonts"
+            }
+        };
```
jobara: Looking at this, seems a bit odd that the configuration for which files to check isn't in the typical grunt config. I think ideally that would be the case for all the verification steps. Is there a reason why the configuration is hardcoded in each task?

waharnum: Is your suggestion simply to move the functions etc that generate the file lists into the Grunt config, or to move to manually maintaining the list of files to verify? (I'm unclear as to which)

jobara: I mean would it be possible to make use of grunts own [files](https://gruntjs.com/configuring-tasks#files) mechanics.

waharnum: I've finished initial refactoring to use Grunt's file mechanics where possible. Where not (the generated JS files), I've made the multitask accept a custom `verifyFilesListFunc`. 

So task configuration either uses the standard Grunt file mechanics where there's a one-to-one mapping (possibly with renaming) between source files and destination files, or specifies a function to return an array of files to check. I think this should have the flexibility required for future additions - most of them will fall into the first case, but the second is available if the verification logic is not as straightforward as "these source files are compiled into these destination files".

Let me know your thoughts.

Gruntfile.js
```diff
@@ -469,6 +475,48 @@ module.exports = function (grunt) {
...
+                    return expectedFilenames;
+                }
+            },
+            css: {
+                files: [
+                    {
+                        expand: true,
+                        src: ["*.styl"],
+                        cwd: "src/framework/preferences/css/stylus/",
+                        dest: "dist/assets/src/framework/preferences/css/stylus/",
```
jobara: There is also a stylus file from the Switch component that gets compiled and placed in the Assets directory. We should probably test for this one too.

Gruntfile.js
```diff
@@ -618,46 +663,48 @@ module.exports = function (grunt) {
...
+        var message = "Verifying all \"" + this.target + "\" files are in /dist";
+        var expectedFiles = [];
 
-        displayVerifyFilesReport(report);
-        processVerifyFilesReport(report);
+        // If the target uses the standard Grunt file options
+        if (this.files.length > 1) {
+            _.forEach(this.files, function (file) {
+                expectedFiles.push(file.dest);
+            });
```
jobara: Minor change, but I think you can just use either [_.map](https://lodash.com/docs/4.17.15#map) or JavaScript's array [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) method.

Gruntfile.js
```diff
@@ -469,6 +475,48 @@ module.exports = function (grunt) {
...
+                verifyFilesListFunc: function () {
+                    var expectedFilenames = [];
+                    var distributions = grunt.config.get("distributions");
+                    _.forEach(distributions, function (value, distribution) {
+                        var jsFilename = "infusion-" + distribution + ".js";
+                        var mapFilename = jsFilename + ".map";
+                        var distDirectory = "dist/";
+                        expectedFilenames.push(distDirectory + jsFilename, distDirectory + mapFilename);
+                    });
+                    return expectedFilenames;
```
jobara: Similar to previous comment, can use the map method instead of forEach.

jobara: May want to consider defining this function outside of the config and just passing in the reference to it here.

jobara:
@waharnum could you please merge master, and maybe file a JIRA about moving the file verification into a grunt module?

waharnum:
Have created Grunt module JIRA as https://issues.fluidproject.org/browse/FLUID-6387

src/framework/preferences/css/stylus/Enactors.styl
jobara: I believe this PR is just waiting on addressing this comment.

src/framework/preferences/css/stylus/utils/Fonts.styl
```diff
@@ -27,5 +60,21 @@ build-fonts-Enactors(textFont) {
                 font-family: unquote(font) !important;
             }
         }
-    }
+    }    
+
+    // Build any necessary @font-face definitions
+    for fontFamilyName in textFontFaces {
```
jobara: Things seem to be working, so I'm not sure if this really matters, but as set up, the `@font-face` declaration comes after using the fonts in the CSS file.

jobara:
@waharnum for the list verification we don't seem to be testing the minified css files nor the PrefsFramework-Icons.woff file.

jobara:
@waharnum when the OpenDyslexic font is applied to the Prefs Editor panel, the contrast selection styling is off. The options look like lego heads, which is kind of cool, but maybe not what we want.
contrast

jobara:
In Firefox the text style options are displayed in their own font, except for OpenDyslexic. I'm not sure if this is a limitation of the browser only showing the system fonts, or if there is something not setup correctly. 
textStyleOptions

In MS Edge, it shows up fine in the original list, but after applying the OpenDyslexic font, the whole lists shows up in OpenDyslexic instead of displaying in their own font types. Again, could be related to OpenDyslexic not being installed on the system.
MS_Edge_textStyle

src/framework/preferences/css/stylus/utils/Fonts.styl
waharnum: It's probably more intuitive to have it come before - I'll make the change.

jobara:
@waharnum maybe just file that in a separate JIRA then, seems suitably different from this work if it's already an issue.

waharnum:
> 
> 
> @waharnum maybe just file that in a separate JIRA then, seems suitably different from this work if it's already an issue.

Filed as https://issues.fluidproject.org/browse/FLUID-6388

waharnum:
> 
> 
> @waharnum for the list verification we don't seem to be testing the minified css files nor the PrefsFramework-Icons.woff file.

I've added the minified CSS files and the icon font file to the verification config.

waharnum:
> @waharnum when the OpenDyslexic font is applied to the Prefs Editor panel, the contrast selection styling is off. The options look like lego heads, which is kind of cool, but maybe not what we want.

Open Dyslexic seems to have unusually tall character blocks, which were extending outside the container. Adding `overflow:hidden` to their style seems to have fixed the issue.

jobara:
merged at 058483bcca139731896ad11e4f734034e1d5cc0b

PR #957, opened 2019-07-19
FLUID-6383: exclude Headless Chrome from testem browsers used.

waharnum:
This PR can be considered an "open for discussion" change, as is the issue described in JIRA: https://issues.fluidproject.org/browse/FLUID-6383 - see the JIRA issue for a fuller explanation.

### DIFF
--- a/tests/testem.js
+++ b/tests/testem.js
@@ -25,7 +25,7 @@ fluid.defaults("fluid.tests.testem", {
         tests:   "%infusion/tests"
     },
     testemOptions: {
-        skip: "PhantomJS,Opera,Safari",
+        skip: "PhantomJS,Opera,Safari,Headless Chrome",
         disable_watching: true,
         tap_quiet_logs: true
     }
--- a/tests/testem.js
+++ b/tests/testem.js
@@ -25,7 +25,7 @@ fluid.defaults("fluid.tests.testem", {
         tests:   "%infusion/tests"
     },
     testemOptions: {
-        skip: "PhantomJS,Opera,Safari,Headless Chrome",
+        skip: "PhantomJS,Opera,Safari,Chrome",
         disable_watching: true,
         tap_quiet_logs: true
     }
--- a/tests/testem.js
+++ b/tests/testem.js
@@ -27,7 +27,9 @@ fluid.defaults("fluid.tests.testem", {
     testemOptions: {
         skip: "PhantomJS,Opera,Safari,Chrome",
         disable_watching: true,
-        tap_quiet_logs: true
+        tap_quiet_logs: true,
+        // Set to headless browser mode without need to set environment variable in CLI
+        browser_args: "@expand:gpii.testem.constructBrowserArgs({that}.options.headlessBrowserArgs)"
     }
 });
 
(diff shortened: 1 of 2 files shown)


waharnum:
@jobara Can you review this PR? I think it may be connected to the problems you & I are both having with the Syllabification tests hanging or failing in CI.

tests/testem.js
```diff
@@ -25,9 +25,11 @@ fluid.defaults("fluid.tests.testem", {
         tests:   "%infusion/tests"
     },
     testemOptions: {
-        skip: "PhantomJS,Opera,Safari",
+        skip: "PhantomJS,Opera,Safari,Chrome",
```
jobara: Is skipping "Chrome" required with the headless browser args specified on line 32 below?

waharnum: I believe so, since Chrome and Headless Chrome are separate launchers - otherwise I think testem will launch both.
The l32 argument forces Firefox to run in headless mode as well, which is otherwise not possible without setting an environment variable. May be temporary, but testing that Firefox tests will pass in headless mode under CI.

jobara:
@the-t-in-rtf would you mind to review this PR. I can merge it after you're happy with the changes.

waharnum:
@the-t-in-rtf @jobara We may want to address the assymetry in setting headless mode for testem - what I've done here is a bit of a hack for testing purposes. I think we should ideally be able to set headless mode with an obvious boolean in the config or an environment variable.

waharnum:
Headless Firefox seemed to fail in CI, or I possibly did the configuration wrong ("worked on local" tho 😢). So at the moment this just removes non-headless Chrome.

the-t-in-rtf:
@waharnum, you should just need to set the environment variable HEADLESS and then make sure both browsers are included in your "launch" option.   I'd suggest adding `HEADLESS=true ` at the beginning of the definition for the npm `test:vagrant` script, that should be picked up by both [the master CI multijob](https://github.com/fluid-project/ci-service/blob/master/jenkins_jobs/infusion-master.yml#L83) and [the PR CI multijob](https://github.com/fluid-project/ci-service/blob/master/jenkins_jobs/infusion-pull-request.yml#L84).

waharnum:
> 
> 
> @waharnum, you should just need to set the environment variable HEADLESS and then make sure both browsers are included in your `launch` option. I'd suggest adding `HEADLESS=true ` at the beginning of the definition for the npm `test:vagrant` script, that should be picked up by both [the master CI multijob](https://github.com/fluid-project/ci-service/blob/master/jenkins_jobs/infusion-master.yml#L83) and [the PR CI multijob](https://github.com/fluid-project/ci-service/blob/master/jenkins_jobs/infusion-pull-request.yml#L84).

@the-t-in-rtf  I believe that will cause the Vagrant scripts to fail on Windows from Powershell or cmd, which isn't desirable. It sounds like a common way of dealing with this is https://github.com/kentcdodds/cross-env, which is heavily used by other projects and under active maintenance. I will test how well it works (I'm on Windows now).

jobara:
We've switched to using headless chrome in 28cfbe44c5d20f23df5b75bb880b337f2156ac94

PR #958, opened 2019-07-22
FLUID-6382: Preventing loading the same pattern file multiple times.

jobara:
In the case where more than one language code points to the same pattern 
file, there could be multiple attempts to load the same pattern if it 
hadn't already been loaded before a new request was submitted. To 
prevent this we keep track of which sources are requested and share the 
promise related to them.

https://issues.fluidproject.org/browse/FLUID-6382


jobara:
@cindyli or @amb26 can you take a look at this PR. It should address those random test failures we were seeing related to Syllabification Enactor tests.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -222,6 +222,13 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         }
 
         var src = fluid.stringTemplate(pattern, that.options.terms);
+        var existingLang = fluid.get(that, ["hyphenatorSRCs", src]);
```
amb26: Rather than code around this problem by creating an extra piece of "back channel state" in the only function we have which has a hard dependency on Hypher, I think we should try to factor "fluid.prefs.enactor.syllabification.createHyphenator" a bit better.
I think we really have two problems in one - as we "smelled " there's indeed a race condition in luid.prefs.enactor.syllabification.waitForHyphenators that would very likely make the hard failure go away if we fixed it - that is, if it continued to iterate over freshly arrived promises . However, this would still end up with an inefficient component which made multiple network requests to the same file.

So what I suggest is, rather than keeping records of hyphenators at multiple paths in "that.hyphenators" that we simply mirror the real state at global Hypher.languages.xxx in our own. This means factoring out the fallback logic within getHyphenator into a self-contained function, such that we only get around to calling createHyphenator once we have already resolved the fallback.
This would result in a much better getHyphenator that 
i) Had only one site for calling createHyphenator
ii) Had only one site for assigning to that.hyphenators, which made only one assignment per call of getHyphenator/createHyphenator
iii) Maintained hyphenators keyed by "resolved language" rather than "requested language" meaning that we only had one entry in our cache for each pattern.

jobara:
An additional comment from @amb26 in the [fluid-work channel on 23 July 2019](http://irc-logs.fluidproject.org/%23fluid-tech/%23fluid-tech.2019-07-23.log) asked:

> I'm just wondering now whether fluid.prefs.enactor.syllabification.waitForHyphenators is even necessary at all

> It seems much cleaner to just stick the event escalation logic inside fluid.prefs.enactor.syllabification.syllabify

> And then we don't need to keep squinting at this promise sequence wondering whose handlers are going to fire first ....

jobara:
@amb26 this is ready for more review. I didn't tackle your additional comments from the [channel](https://github.com/fluid-project/infusion/pull/958#issuecomment-515049663) yet because I'm not sure how we'd know when the syllabification would be complete. As mentioned in the [fluid-tech channel yesterday](http://irc-logs.fluidproject.org/%23fluid-tech/%23fluid-tech.2019-07-25.log) I'm not sure of another way were we could know that the syllabification has finished because there could be multiple languages on the page and the injection of the pattern files is asynchronous.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -235,66 +242,78 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
         }
 
-        // Attempt to create an appropriate hyphenator
+        return {
+            lang: lang,
+            src: src
+        };
+    };
+
+    // /**
```
amb26: This has mysteriously become a commented-out doc comment?

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -235,66 +242,78 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+     * requested language code cannot be located, it will attempt to locate a fall back, by looking for a pattern
+     * supporting the generic language code. If no pattern can be found, `undefined` is returned as the `src` value.
      *
-     * When creating a hyphenator, it first checks if there is configuration for the specified `lang`. If that fails,
-     * it attempts to fall back to a less specific localization.
      *
-     * @param {Component} that - an instance of `fluid.prefs.enactor.syllabification`
      * @param {String} lang - a valid BCP 47 language code. (NOTE: supported lang codes are defined in the
      *                        `patterns`) option.
+     * @param {String} patterns - an object mapping language codes to file paths for the pattern files
```
amb26: Looks like this doc type is incorrect - perhaps Object.<String, String> ?

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -205,7 +210,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
      * if it hasn't already been loaded. If the pattern file cannot be loaded, the onError event is fired.
      *
      * @param {Component} that - an instance of `fluid.prefs.enactor.syllabification`
-     * @param {Object} pattern - the `file` path to source the pattern file from.
+     * @param {Object} pattern - the `file path` to source the pattern file from.
```
amb26: Looks like this is actually a String holding a fluid.stringTemplate pattern

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -131,10 +131,15 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                 funcName: "fluid.prefs.enactor.syllabification.parse",
                 args: ["{that}", "{arguments}.0"]
             },
+            createHyphenator: {
```
amb26: Please add a comment in the "members" area defining "hyphenators" explaining that it is a map of Strings representing sources to Promises, etc.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -218,12 +223,14 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
         // If the pattern file has already been loaded, return the hyphenator.
+        // This could happen if the pattern file is statically linked to the page.
         if (hyphenator) {
             promise.resolve(hyphenator);
             return promise;
         }
 
         var src = fluid.stringTemplate(pattern, that.options.terms);
+
         var injectPromise = that.injectScript(src);
```
amb26: This is a comment for the impl of injectScript - note that the return value of $.ajax is already a promise so there is no need to construct a further fluid.promise

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -235,66 +242,78 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                 textStatus,
                 errorThrown
             );
-            // If the pattern file could not be loaded resolve the promise with out a hyphenator (undefined).
+            // If the pattern file could not be loaded, resolve the promise without a hyphenator (undefined).
```
amb26: This is a comment for the error block on line 239 - note that the signature you attempt to use of (jqXHR, textStatus, errorThrown) cannot be received since the promise rejection API only supplies a single error payload as a value.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -235,66 +242,78 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+    //  * to create one and assign the related promise to the `hyphenators` member for future retrieval.
+    //  *
+    //  * When creating a hyphenator, it first checks if there is configuration for the specified `lang`. If that fails,
+    //  * it attempts to fall back to a less specific localization.
+    //  *
+    //  * @param {Component} that - an instance of `fluid.prefs.enactor.syllabification`
+    //  * @param {String} lang - a valid BCP 47 language code. (NOTE: supported lang codes are defined in the
+    //  *                        `patterns`) option.
+    //  *
+    //  * @return {Promise} - returns a promise. If a hyphenator is successfully created, it is resolved with it.
```
amb26: It seems more appropriate to reject rather than resolve with undefined if the hyphenator can't be created

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -235,66 +242,78 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+        var promise = fluid.promise();
         var hyphenatorPromise;
-        var pattern = that.options.patterns[lang];
 
-        if (pattern) {
-            hyphenatorPromise = fluid.prefs.enactor.syllabification.createHyphenator(that, pattern, lang);
-            fluid.promise.follow(hyphenatorPromise, promise);
-            that.hyphenators[lang] = hyphenatorPromise;
+        if (!lang) {
+            promise.resolve();
```
amb26: reject seems more appropriate here

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -235,66 +242,78 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
-        if (pattern) {
-            hyphenatorPromise = fluid.prefs.enactor.syllabification.createHyphenator(that, pattern, langSegs[0]);
-            fluid.promise.follow(hyphenatorPromise, promise);
-        } else {
+        if (!pattern.src) {
             hyphenatorPromise = promise;
-            // If there no available patterns to match the specified language, resolve the promise with out a
-            // hyphenator (undefined).
             promise.resolve();
+            return promise;
```
amb26: Again reject seems more appropriate here

src/framework/preferences/js/SyllabificationEnactor.js
jobara: This is actually related to a conversation I had with you when the syllabification code was first being implemented. Originally this was rejected, but it was suggested to resolve with an empty value instead. I believe for the use of promise sequences to trigger the afterSyllabification event.

see: http://irc-logs.fluidproject.org/%23fluid-work/%23fluid-work.2018-11-02.log

amb26: @jobara Thanks for locating this ancient conversation. It seems that the main reason for the choice was the unsuitability of our fluid.promise.sequence algorithm for this use case and several others. It will be ok if you leave an implementation comment here explaining that a promise rejection would be more appropriate here and that we can improve the implementation once https://issues.fluidproject.org/browse/FLUID-5938 is resolved

jobara:

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -218,29 +215,64 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
 
+    /**
+     * Assembles an Object containing the information for locating the pattern file. If a pattern for the specific
+     * requested language code cannot be located, it will attempt to locate a fall back, by looking for a pattern
+     * supporting the generic language code. If no pattern can be found, `undefined` is returned as the `src` value.
+     *
+     *
+     * @param {String} lang - a valid BCP 47 language code. (NOTE: supported lang codes are defined in the
+     *                        `patterns`) option.
+     * @param {Object} patterns - an object mapping language codes to file paths for the pattern files. For example:
```
amb26: Please refine the types a bit here - e.g. {Object.<String, String>} and we could have a small @typedef for the "pattern" object value type in the returned structure
https://jsdoc.app/tags-type.html

jobara: @amb26 I've made the change to the comment regarding the types. Please let me know if I have addressed it appropriately.

jobara:
@amb26 ready for more review.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -218,29 +215,72 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
         });
+
         return promise;
     };
 
+    /**
+     * Information about a pattern, including the resolved language code and the file path to the pattern file.
+     *  {"lang": "resolvedLanguageCode", src: "pattern/file/path"}
+     *
+     * @typedef {Object} PatternInfo
```
amb26: It's possible and desirable to document the members structurally using @property rather than in a comment - see examples at https://jsdoc.app/tags-typedef.html

jobara: I modified this to use `@property`. I also fixed up a couple other typedefs that had a similar issue.

jobara:
@amb26 ready for more review.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -218,29 +215,74 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
-                "The pattern file \"" + src + "\" could not be loaded.",
-                jqXHR,
-                textStatus,
-                errorThrown
-            );
-            // If the pattern file could not be loaded resolve the promise with out a hyphenator (undefined).
+        }, function () {
+            var errorMessage = fluid.stringTemplate(that.options.strings.patternLoadError, {src: src});
+            // Pass along arguments fired for the rejection, but with our errorMessage as the first argument
+            var args = [errorMessage].concat(fluid.values(arguments));
```
amb26: fluid.values is an odd choice given arguments is arraylike - we would typically use fluid.makeArray here. But in fact there is no choice about the signature - the promise API is fixed and always provides a single-argument error payload. So you can just say function (error) and push that into the argument list. As a courtesy it might be worth checking whether it is a string and if so simply appending it to the message rather than dangling it off the argument list where it might not be seen (especially since this argument list is not documented). Another useful courtesy might be to send the full error message to fluid.log here in addition to firing the event.

jobara:
@amb26 ready for more review.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -227,11 +227,15 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
             promise.resolve(hyphenator);
-        }, function () {
-            var errorMessage = fluid.stringTemplate(that.options.strings.patternLoadError, {src: src});
-            // Pass along arguments fired for the rejection, but with our errorMessage as the first argument
-            var args = [errorMessage].concat(fluid.values(arguments));
-            that.events.onError.fire.apply(null, args);
+        }, function (error) {
+            var errorInfo = {
+                src: src,
+                errorMsg: typeof(error) === "string" ? error : ""
```
amb26: Still not quite right - now you throw away the "error" object completely when is it not a string

jobara: Thanks, I had meant to add it to the fluid.log call as well.

jobara:
@amb26 ready for more review.

PR #960, opened 2019-08-14
FLUID-6394: Addressing positioning of TTS selection reader control 

jobara:
Taking into consideration margin of container when positioning TTS selection reader control.

https://issues.fluidproject.org/browse/FLUID-6394

dev release for testing: 3.0.0-dev.20190820T145526Z.389bd6937.FLUID-6394

Needed for https://issues.gpii.net/browse/GPII-3767

### DIFF
--- a/src/components/orator/js/Orator.js
+++ b/src/components/orator/js/Orator.js
@@ -889,28 +889,52 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
     };
 
     /**
+     * An object containing specified offset scales: "edge" and "pointer" offsets to account for the room needed for a
+     * control element to be correctly positioned.
+     *
+     * @typedef {Object} OffsetScale
+     * @property {Float} edge - The minimum distance between the button and the viewport's edges
+     * @property {Float} pointer - The distance between the button and the coordinates the DOMRect refers too. This
+     *                                provides space for an arrow to point from the button.
+     */
+
+    /**
+     * An object containing the sizes of the top and left margins.
+     *
+     * @typedef {Object} MarginInfo
+     * @property {Float} top - The size of margin-top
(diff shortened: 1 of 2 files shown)


jobara:
@amb26 or @cindyli are you able to review this PR?

cindyli:
This fix does offer a better user experience. Here are a few other issues I noticed:

1. Issues with selecting the top line on the webpage:
* When dragging mouse from right to left to select the top line text, the display of "Play" button covers some of the selected text:
image

* When dragging mouse from left to right to select the top line text, the display of "Play" button shows much ahead of the start of the selected text (this might because the Morphic logo image is considered as selected so the button is displayed there):
image

2. When a chunk of text is selected, the "play" button is displayed closed to the first line of selected text. When continuing to drag the mouse to select more text, or stop dragging but scroll the page to see the content below, "play" button stays at where it was rendered and disappears from the view. It would be nice to keep the "play" button always showing in the viewport but this might be an extra feature.
image

jobara:
@cindyli thanks for your feedback. Below is my response to your points above.

1. Selecting top line
  a. will need to investigate. Probably something to do with collision detection with the top of the screen.
  b. the location of the play button is based off of where selected range indicates is the start position. I don't think we can adjust this without modifying the users selection, which may not be what the user wants.
2. We could really do either or, but I find it becomes very confusing when the play button starts moving around. That can still happen in some cases. You can see that with the overview panel in our demos because it is fixed positioned. Maybe a question for a design crit?

jobara:
@cindyli ready for more review. There was a small regression with the positioning when the play button should have been below. I've updated the info in the comment with the latest dev release for testing in UIO+.

cindyli:
Merged at b024bff871a7d4a73a3fc8f609d8dfd08149f209

PR #963, opened 2019-10-28
FLUID-6394: Position of play button 

jobara:
Fixes the position of the selectionReader play button.

https://issues.fluidproject.org/browse/FLUID-6394

### DIFF
--- a/tests/component-tests/orator/js/OratorTests-Utils.js
+++ b/tests/component-tests/orator/js/OratorTests-Utils.js
@@ -125,6 +125,16 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
      * Assertions
      *******************************************************************************/
 
+    // assert that the element is actually in the document, not detached
+    fluid.tests.orator.assertNodeInDOM = function (msg, elm) {
+        jqUnit.okWithPrefix($.contains(document, fluid.unwrap(elm)), msg);
+    };
+
+    // assert that the element is not in the document, may be detached
+    fluid.tests.orator.assertNodeNotInDOM = function (msg, elm) {
+        jqUnit.okWithPrefix(!$.contains(document, fluid.unwrap(elm)), msg);
+    };
+
     fluid.tests.orator.verifyControllerState = function (controller, state) {
         var toggleButton = controller.locate("playToggle");
         jqUnit.assertEquals("The model state should be set correctly", state, controller.model.playing);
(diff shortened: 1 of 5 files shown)


jobara:
@amb26 and @cindyli would one of you be able to review this PR?

cindyli:
There could be collision to the right too:

Screen Shot 2019-10-28 at 2 40 15 PM

jobara:
@cindyli it does check for collisions with the right hand side. Is that the scrollbar that it is colliding with in the picture? It might also be related to your previous comment.

cindyli:
The collision was with the right edge, not the scrollbar. I've moved the cursor out of the browser area to ensure the scrollbar fades away.

jobara:
@cindyli I think the issue was with the other comment you left regarding `viewPortWidth || document.body.clientWidth;`. I've addressed this. Please let me know if it fixes the issue you were seeing with the right hand side collision.

cindyli:
Yes, it does fix the right hand side collision. Thanks.

tests/component-tests/orator/js/OratorTests-Utils.js
```diff
@@ -125,6 +125,16 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
      * Assertions
      *******************************************************************************/
 
+    // assert that the element is actually in the document, not detached
+    fluid.tests.orator.assertNodeInDOM = function (msg, elm) {
```
cindyli: Functions at line 129 & 135 are pretty much the same. It might be simpler to define just one `isNodeInDOM` function with `$.contains(document, fluid.unwrap(elm));`. This function returns a boolean that can be verified in caller functions.

jobara: I was trying to keep some symmetry with the `assertNodeExists` and `assertNodeNotExists` functions from jqUnit-browser.js

https://github.com/fluid-project/infusion/blob/master/tests/test-core/jqUnit/js/jqUnit-browser.js#L45-L51 

What do you think?

cindyli: I'm not a big fan of writing duplicate code to keep symmetry with somewhere else, but this is your choice and I respect. ;)

Let's keep as is.

cindyli:
I hit this error when testing this pull request with UIO+ master:

Screen Shot 2019-10-28 at 4 11 24 PM

Resource loader issue?

jobara:
@cindyli, I forgot to warn you. UIO+ master is now based off of the FLUID-6148 branch. Where as this PR is based off of Infusion master. The easiest way to test it would be to use a slightly older version of UIO+ from before the FLUID-6148 PR was merged. The plan to get this particular bit of code into UIO+ will be after @amb26 merges Infusion master into his FLUID-6148 branch and creates a new dev release, I'll be able to update that in UIO+.

cindyli:
I tested this pull request with the UIO+ commit hash `9406c628a38c0851e7b521868a22e4476bdd8ac8`. When selecting the text on the banner, the "play" button is display below the selected text instead of above:

Screen Shot 2019-10-28 at 5 07 27 PM

jobara:
According to the [spec](https://drafts.csswg.org/cssom-view/#dom-htmlelement-offsettop):

> Return the result of subtracting the y-coordinate of the top padding edge of the first CSS layout box associated with the offsetParent of the element from the y-coordinate of the top border edge of the first CSS layout box associated with the element, relative to the initial containing block origin, ignoring any transforms that apply to the element and its ancestors.

So is it a bug in one of the browsers or is just a matter of interpretation of what they consider to be the border edge.

jobara:
Ok to test

jobara:
@cindyli this PR is ready for more review.

PR #964, opened 2019-10-28
FLUDI-6400: Prefs Framework primary schema updates

jobara:
First steps for Prefs Framework refactoring. Updating the primary schema entries to use `multipleOf` instead of `divisibleBy` and to have a property for `enumLabels`.

https://issues.fluidproject.org/browse/FLUID-6400


jobara:
@amb26 would you be able to review this PR?

PR #967, opened 2019-11-06
FLUID-6420: Splitting starter schema and refactoring fluid.prefs.create

jobara:
Split the starter schema into independent schemas per preferences. Still retained a starter schema to aggregate the original base set as a quick option. Also refactored the fluid.prefs.create function to be a grade that calls the builder and supplies the generated grade to the gradeNames via an invoker.

https://issues.fluidproject.org/browse/FLUID-6420


jobara:
@amb26 could you please review this PR?

src/framework/preferences/js/Builder.js
```diff
@@ -244,10 +276,27 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
     /*
      * A one-stop-shop function to build and instantiate a prefsEditor from a schema.
      */
-    fluid.prefs.create = function (container, options) {
-        options = options || {};
-        var builder = fluid.prefs.builder(options.build);
-        return fluid.invokeGlobalFunction(builder.options.assembledPrefsEditorGrade, [container, options.prefsEditor]);
+    fluid.defaults("fluid.prefs.create", {
+        gradeNames: ["fluid.viewComponent", "{that}.getGradeFromSchema"],
```
amb26: It looks like this uses an unblessed technique of using a component's grade constructor as a free function - we will have better error checking in the future for this but this must never happen. Component creator functions must only ever be used by top-level code kicking off a standalone component tree construction - all other construction must occur declaratively.
It looks like we should return fluid.prefs.create into being a function, and then alter the action of fluid.prefs.builder so that it instead constructs some kind of linkage targetting fluid.uiOptions.prefsEditor rather than rewriting the contents of a statically named grade like fluid.prefs.constructed.prefsEditor - we should remove the latter from fluid.uiOptions.prefsEditor to avoid accidents. But fluid.prefs.create should be clear that it only advises the particular prefs editor that it creates itself, by running fluid.prefs.builder in a particular mode that doesn't cause the grade side-effects. The invocation of it in UIOptions.js should be provided some special option to indicate that it is intended to have the "spooky action at a distance" effect on fluid.uiOptions.prefsEditor

jobara: Spoke with @amb26 about this in the [fluid-work channel today](http://irc-logs.fluidproject.org/%23fluid-work/%23fluid-work.2019-11-08.log).

jobara: @amb26 what do you think of this idea.

- Have fluid.prefs.create (or rename to something more appropriate) be the top level component 
  - has a grade of fluid.primary.builder that merges the primary schemas
  - has a grade of something like fluid.primary.auxiliarybuilder that merges the auxiliary schemas
  - takes an array of preferences which are used to determine which primary and auxiliary schemas are added
  - has an invoker grade that generates and returns the synthesized grades to create the prefseditor, enhancer, and store sub components. Possibly an option to only create the store, prefseditor/store, enhancer/store, or all three.

This would keep a similar setup to the current PR where fluid.prefs.create is actually a component grade, the primary builder is likely not touched, the builder and auxbuilder are refactored (hopefully not too big a change), and the user of fluid.prefs.create wouldn't have to write anything that looked like a schema and could just request the preferences they want based on the preference type.

jobara:
@amb26 in my latest push, I've made use of the mergePolicy to try to merge the supplied preferences with those from context awareness. I'm able to expand the preferences, but the one from the context awareness is never contributed. However, the grade is present in the grade names. You can test this out by running the prefs framework demo.

src/framework/preferences/js/PrimaryBuilder.js
```diff
@@ -77,22 +79,22 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
      * An invoker method that builds a list of grades that comprise a final version of the primary schema.
      * @param {JSON} schemaIndex - A global index of all schema grades registered with the framework.
-     * @param {Array} typeFilter   - A list of all necessarry top level preference names.
+     * @param {Array} preferences   - A list of all necessary top level preference names.
      * @param {JSON} primarySchema - Primary schema provided as an option to the primary builder.
      * @return {Array} - A list of schema grades.
      */
-    fluid.prefs.primaryBuilder.buildPrimary = function (schemaIndex, typeFilter, primarySchema) {
+    fluid.prefs.primaryBuilder.buildPrimary = function (schemaIndex, preferences, primarySchema) {
         var suppliedPrimaryGradeName = "fluid.prefs.schemas.suppliedPrimary" + fluid.allocateGuid();
```
amb26: Whilst this is in progress, we should note that "supplied" is a quite misleading name for this since it is not supplied at all. Better to just use the name of this function itself as a clue to its origin, e.g. "buildPrimary"

jobara:
@amb26 I believe this is ready for more review now. However, it is now based off of https://github.com/fluid-project/infusion/pull/972. Would you prefer that I close this PR and submit a PR directly against that branch?

amb26:
Just as a reminder to ourselves - this pull is only for our internal reference and its contents have already been merged into the FLUID-6145 - it is also cross-contaminated with that branch and so will be merged together with it.

jobara:
These changes are now tracked in https://github.com/fluid-project/infusion/pull/1052

PR #974, opened 2019-12-05
FLUID-6431: Updating selection button rendering on text change.

jobara:
Right Clicks in some browsers also perform a selection. In Safari and 
Chrome those selectionchange events fire twice but with the same text 
selection so there is no time where the UI would have been removed. By 
rendering the text on all selection changes, the UI is always at the 
beginning of the text and will follow right click selections.

https://issues.fluidproject.org/browse/FLUID-6431

### DIFF
--- a/src/components/orator/js/Orator.js
+++ b/src/components/orator/js/Orator.js
@@ -841,7 +841,6 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         },
         model: {
             enabled: true,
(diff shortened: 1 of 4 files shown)


jobara:
@cindyli would you mind to review this PR?

cindyli:
Merged at 46a1b887dcc4d78a78b3c3f8faaab498afc782cd

PR #975, opened 2020-01-31
FLUID-6421: provides the ability to generate panel containers

jobara:
This allows for greater flexibility when specifying which preferences to display, as you do not need to create a new template for each variation of preferences. However, this only works for standard cases like those in the current UIO. More complex variations such as composite panels or the categorized view in the full page examples, are not yet supported. For those complex cases a full template is required.

https://issues.fluidproject.org/browse/FLUID-6421

Includes change from https://github.com/fluid-project/infusion/pull/967


jobara:
@amb26 could you please review these changes?

jobara:
These changes are now tracked in https://github.com/fluid-project/infusion/pull/1052

PR #980, opened 2020-03-26
FLUID-6482: add guard to fluid.getCallerInfo if atDepth exceeds stack size

waharnum:
See https://issues.fluidproject.org/browse/FLUID-6482 for details of this issue (seen on Safari).

### DIFF
--- a/src/framework/core/js/FluidDebugging.js
+++ b/src/framework/core/js/FluidDebugging.js
@@ -138,7 +138,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
     fluid.getCallerInfo = function (atDepth) {
         atDepth = (atDepth || 3) - stackStyle.offset;
         var stack = fluid.decodeStack();
-        var element = stack && stack[atDepth][0];
+        var element = stack && stack[atDepth] && stack[atDepth][0];
         if (element) {
             var lastslash = element.lastIndexOf("/");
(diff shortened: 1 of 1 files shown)


amb26:
If we're receiving the wrong stack depth, this implies that Safari is producing a stack trace that fluid.decodeStack can't decode. Rather than just putting in a guard to fix the symptoms, we should fix the implementation of fluid.decodeStack.

waharnum:
@amb26 Agreed - the immediate term, could we merge this guard and create a separate issue to fix `fluid.decodeStack`? I ask because this is currently a critical blocker in our use of Infusion on the Clusive project, and the full fix may take some time to implement.

amb26:
I'm happy to merge this, although equally happy to just roll you a dev release from this branch for your own use, which would probably be more direct help. Also, if you can attach a stack trace as issued by Safari to fluid.decodeStack I can work it up to a proper fix.

waharnum:
@amb26 rolling a dev release sounds great, thanks; I'll do some investigation of the stack trace in Safari this afternoon and see if I can find a path to replication. Will update issue and PR as I find out more.

amb26:
Cheers @waharnum - I've rolled 3.0.0-dev.20200326T173810Z.24ddb2718.FLUID-6482 containing the contents of this branch

PR #984, opened 2020-04-16
FLUID-6493: Update node.js dependencies

gtirloni:
I need to update a dependency that was causing builds inside containers to break. So I've updated all dependencies that had newer versions, except for jQuery (because it breaks our tests).


cindyli:
ok to test

gtirloni:
@jobara @amb26 @colinbdclark @BlueSlug @cindyli and others, I can't add reviewers to PR's so if you could take a look at this I'd appreciate.

PR #988, opened 2020-05-13
FLUID-6497: Can now set any cookie attribute.

jobara:
Updates defaults to set samesite attribute.

https://issues.fluidproject.org/browse/FLUID-6497

### DIFF
--- a/src/framework/preferences/js/Store.js
+++ b/src/framework/preferences/js/Store.js
@@ -42,8 +42,41 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
      ****************/
 
     /**
-     * SettingsStore Subcomponent that uses a cookie for persistence.
-     * @param options {Object}
+     * Options for reading and/or writing a cookie using the `Document.cookie` web api through a {fluid.dataSource}.
+     * In most cases these are used directly for setting attributes of the same name on the cookie. However, the `name`
+     * and `directModel` properties are used for reading/writing the specific cookie. The `directModel` is typically
+     * included via calls to the `dataSource`'s `get` and `set` methods. The `directModel` is typically prioritized for
+     * sourcing the cookie's name, followed by the `name` property.
+     *
+     * Note: Other than `name` and `directModel` all property names are case insensitive.
+     * See: https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie for more details on `Document.cookie`
+     *
+     * @typedef {Object} CookieOpts
+     *
+     * @property {Object} [directModel] - An object defining the coordinates to the cookie
+     *                                    see: `fluid.dataSource` for more information.
+     * @property {String} [directModel.cookieName] - The name of the cookie to read/write
+     * @property {String} [domain] - The domain to scope the cookie's permissions to.
+     *                               (e.g. "example.com" or "subdomain.example.com"). Defaults to the host portion of
+     *                               the current `document.location`.
+     * @property {GMTString} [expires] - The cookie expiry date.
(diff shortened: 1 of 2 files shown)


src/framework/preferences/js/Store.js
```diff
@@ -128,13 +131,14 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
-        if (options.expires) {
-            cookieStr += "; expires=" + options.expires;
-        }
-
-        if (options.path) {
-            cookieStr += "; path=" + options.path;
-        }
+        fluid.each(options, function (value, setting) {
+            // skip name because it was already set as the first portion of the cookieStr
+            if (setting !== "name" && setting !== "secure") {
```
cindyli, line 136: `HttpOnly` is another no-value cookie setting, similar as `secure`. See [this documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Syntax). It's good to handle it too.

jobara: The cookie store is meant to be used in the browser, as it uses [document.cookie](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie). If you look at the JavaScript section [Creating Cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Creating_cookies) it says:

> Cookies created via JavaScript cannot include the HttpOnly flag.

src/framework/preferences/js/Store.js
```diff
@@ -128,13 +131,14 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
         var cookieStr = cookieName + "=" + data;
 
-        if (options.expires) {
-            cookieStr += "; expires=" + options.expires;
-        }
-
-        if (options.path) {
-            cookieStr += "; path=" + options.path;
-        }
+        fluid.each(options, function (value, setting) {
```
amb26, line 134: While we are here, could we take the opportunity to improve the docs of fluid.prefs.cookieStore.assembleCookie - e.g. more accurate type for "options" - perhaps Object.<String, String>, or perhaps with a typedef given this argument type appears in more than one function, and a list of at least the settings that we know we support - perhaps with a link to the MDN documentation that you quoted.
Also perhaps fix the minor anomaly that the argument is called "payload" in writeCookie but "data" in assembleCookie. It looks like the argument type in writeCookie must be wrong

amb26: Other comment issues include the use of "SettingsStore" where this has now been migrated over to DataSources, and the use of "Subcomponent" is anachronistic

jobara:
@cindyli and @amb26 this is ready for more review.

tests/framework-tests/preferences/js/StoreTests.js
```diff
@@ -48,31 +48,56 @@ https://github.com/fluid-project/infusion/raw/master/Infusion-LICENSE.txt
...
+            },
+            expected: "cookieName=cookieValue"
         }, {
-            testName: "no path or expiry date set",
+            testName: "only name set",
             cookieName: "cookieName",
             payload: "cookieValue",
             expected: "cookieName=cookieValue"
+        }, {
+            testName: "name attribute in cookie options",
```
cindyli, line 94: Probably add a test case that the name attribute in cookie options is applied when `cookieName` is not provided.

jobara: `fluid.prefs.cookieStore.assembleCookie` doesn't handle the name calculations. That is handled by `fluid.prefs.cookieStore.writeCookie`. There already tests in place against that function. However, I've done a bit of refactoring to make the test reporting a bit clearer, and simplified some other parts of the test code.

PR #990, opened 2020-05-15
FLUID-6454: Removed title attribute from navbar

sachin10101998:
There's a title attribute on the "Previous" link and the "Next" link. The issue here is that the link has accessible text that will result in screen readers announcing previous twice (e.g. "link, Previous, Previous page" in VoiceOver. It's not recommended that the title attribute be used on links because of inconsistent browser support and the potential for duplicate content being announced by screen readers. 
Reference from MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#Accessibility_concerns

### DIFF
--- a/demos/pager/css/pager.css
+++ b/demos/pager/css/pager.css
@@ -138,4 +138,3 @@
  .demo-pager-next-character::after {
      content: ">";
  }
- 
--- a/demos/pager/css/pager.css
+++ b/demos/pager/css/pager.css
@@ -127,3 +127,14 @@
  .demo-pager-page-size  label {
     margin-right: 2px;
  }
+
+ /***********************************************
+ * Styles for pager previous and next page symbol
+ */
+ .demo-pager-previous-character::before {
+     content: "<";
+ }
+
+ .demo-pager-next-character::after {
+     content: ">";
+ }
--- a/demos/pager/css/pager.css
+++ b/demos/pager/css/pager.css
@@ -138,3 +138,4 @@
  .demo-pager-next-character::after {
      content: ">";
  }
+ 
--- a/demos/pager/css/pager.css
+++ b/demos/pager/css/pager.css
@@ -127,14 +127,3 @@
  .demo-pager-page-size  label {
     margin-right: 2px;
  }
-
- /***********************************************
- * Styles for pager previous and next page symbol
- */
- .demo-pager-previous-character::before {
-     content: "<";
- }
-
- .demo-pager-next-character::after {
-     content: ">";
- }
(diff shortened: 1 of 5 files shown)


gtirloni:
ok to test

BlueSlug:
Thanks for filing this, @sachin10101998. I've left a comment on the [Jira issue](https://issues.fluidproject.org/browse/FLUID-6454?focusedCommentId=41433&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-41433) detailing some limitations with this particular fix. I recommend using `aria-label` instead, but it would be good to wait for confirmation from @jobara or @greatislander.

demos/pager/css/pager.css
```diff
@@ -127,3 +127,14 @@
...
+ /***********************************************
+ * Styles for pager previous and next page symbol
+ */
+ .demo-pager-previous-character::before {
+     content: "<";
+ }
+
+ .demo-pager-next-character::after {
+     content: ">";
+ }
```
jobara, line 140: please add an empty new line character at the end of the file.

sachin10101998: Done

sachin10101998: Done

jobara:
@sachin10101998 after testing with VoiceOver on macOS I've found that my original suggestion didn't work. I've left an updated [comment](https://issues.fluidproject.org/browse/FLUID-6454?focusedCommentId=41534&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-41534) on the issue. Also, when you make the update, could you please also updated the [Pager Examples](https://github.com/fluid-project/infusion/tree/master/examples/components/pager)?

demos/pager/index.html
```diff
@@ -62,15 +62,15 @@ <h1>School Roster</h1>
 
             <div class="flc-pager-top">
                 <ul class="demo-pager-bar fl-force-left fl-focus">
-                    <li class="flc-pager-previous demo-pager-previous"><a href="#" title="Previous Page">&lt; Previous</a></li>
+                    <li class="flc-pager-previous demo-pager-previous"><span class="demo-pager-previous-character" aria-hidden="true">&lt;</span><a href="#"> Previous</a></li>
```
jobara, line 65: I don't think you need the class names anymore with the characters added in directly.

jobara: the `<` and `>` symbols should be contained within the `<a>`. If that isn't workable with the `aria-hidden="true"` we may need to find an alternative solution.

sachin10101998: Yes @jobara , with aria-hidden true, we can't put things that are focusable inside the span element. We'll have to go with icons as discussed after making changes in the pager component.

amb26:
Closing since the pager component will be removed from the next Infusion release

PR #992, opened 2020-05-26
FLUID-6500: Improve usability of arrayToSetMembership and setMembershipToArray transforms

dmahajan980:
This pull request improves the `arrayToSetMembership` and `setMembershipToArray` transformations by removing the 3rd argument `transformer` and making the 2nd argument `transformSpec` optional. It also includes a detailed description in the form of JSDoc comments and a set of test cases for both the transformations.

### DIFF
--- a/src/framework/core/js/ModelTransformationTransforms.js
+++ b/src/framework/core/js/ModelTransformationTransforms.js
@@ -388,7 +388,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
      * https://docs.fluidproject.org/infusion/development/ModelTransformationAPI.html#fluidtransformsarraytosetmembership
      *
      * @param {Array} value - The Array to be transformed into an object.
-     * @param {Object} [transformSpec] - The options provided to the transformation rule.
+     * @param {Object} [transformSpec] - (optional) The options provided to the transformation rule.
      * @return {Object} - The transformed object
      *
      */
@@ -402,18 +402,21 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             missingValue = fluid.get(transformSpec, "missingValue") ? transformSpec.missingValue : false,
             options = fluid.get(transformSpec, "options");
 
-        // When <options> are provided in the transformation rule, use them for keys in the output object.
-        fluid.each(options, function (outPath, key) {
-            /**
-             * Write to output object the value <presentValue> or <missingValue> depending on whether key is
-             * found in user input.
-             */
-            var outVal = (value.indexOf(key) !== -1) ? presentValue : missingValue;
-            fluid.set(output, outPath, outVal);
-        });
-
-        // Otherwise, when no <options> are provided, use items from the <value> array as keys.
-        if (!options) {
+        /**
+         * When <options> are provided in the transformation rule, use them for keys in the output object.
+         * Otherwise, use items from the <value> array as keys.
+         */
+        if (options !== undefined) {
+            fluid.each(options, function (outPath, key) {
+                /**
+                 * Write to output object the value <presentValue> or <missingValue> depending on whether key is
+                 * found in user input.
+                 */
+                var outVal = (value.indexOf(key) !== -1) ? presentValue : missingValue;
+                fluid.set(output, outPath, outVal);
+            });
+        }
+        else {
             fluid.each(value, function (outPath) {
                 // Write <presentValue> for all the items present in the <value> array.
                 fluid.set(output, outPath, presentValue);
@@ -489,29 +492,32 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
      * https://docs.fluidproject.org/infusion/development/ModelTransformationAPI.html#fluidtransformssetmembershiptoarray
      *
      * @param {Object} input - The object to be transformed into an array.
-     * @param {Object} [transformSpec] - The options provided to the transformation rule.
+     * @param {Object} [transformSpec] - (optional) The options provided to the transformation rule.
      * @return {Array} - The transformed array
      *
      */
     fluid.transforms.setMembershipToArray = function (input, transformSpec) {
         // <input> should be an object.
-        if (!input || !fluid.isPlainObject(input, true)) {
+        if (!fluid.isPlainObject(input, true)) {
             fluid.fail("setMembershipToArray didn't find object at inputPath nor passed as value.");
         }
         var outputArr = [];
         var presentValue = fluid.get(transformSpec, "presentValue") ? transformSpec.presentValue : true,
             options = fluid.get(transformSpec, "options");
 
-        // When <options> are provided in the transformation rule, use them for entries in the output array.
-        fluid.each(options, function (outputVal, key) {
-            // Write to output array the key of the <input> object if it's value matches the <presentValue>
-            if (fluid.get(input, key) === presentValue) {
-                outputArr.push(outputVal);
-            }
-        });
-
-        // Otherwise, when no <options> are provided, use the <input> object for entries in the output array.
-        if (!options) {
+        /**
+         * When <options> are provided in the transformation rule, use them for entries in the output array.
+         * Otherwise, use the <input> object for entries in the output array.
+         */
+        if (options !== undefined) {
+            fluid.each(options, function (outputVal, key) {
+                // Write to output array the key of the <input> object if it's value matches the <presentValue>
+                if (fluid.get(input, key) === presentValue) {
+                    outputArr.push(outputVal);
+                }
+            });
+        }
+        else {
             fluid.each(input, function (keyValue, outputVal) {
                 // Write to output array the key of the <input> object if it's value matches the <presentValue>
                 if (keyValue === presentValue) {
--- a/src/framework/core/js/ModelTransformationTransforms.js
+++ b/src/framework/core/js/ModelTransformationTransforms.js
@@ -349,27 +349,71 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         invertConfiguration: "fluid.transforms.arrayToSetMembership.invert"
     });
 
+    /**
+     *
+     * Transform the input array into an object based upon the options provided to the specification. Accepts
+     * an array as its first argument and an optional second argument (an object) which contains the
+     * specifications for the resulting output.
+     *
+     * An error will be thrown if the input to be transformed is not an array.
+     *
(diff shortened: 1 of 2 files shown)


the-t-in-rtf:
ok to test.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,182 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+     * https://docs.fluidproject.org/infusion/development/ModelTransformationAPI.html#fluidtransformssetmembershiptoarray
+     *
+     * @param {Object} input - The object to be transformed into an array.
+     * @param {Object} [transformSpec] - The options provided to the transformation rule.
+     * @return {Array} - The transformed array
+     *
+     */
+    fluid.transforms.setMembershipToArray = function (input, transformSpec) {
+        // <input> should be an object.
+        if (!input || !fluid.isPlainObject(input, true)) {
```
amb26, line 498: First branch is unnecessary, no falsy object can pass isPlainObject

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,182 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+     * }
+     *
+     * The output array will be:
+     *
+     * ["keyboard", "mouse"]
+     *
+     * https://docs.fluidproject.org/infusion/development/ModelTransformationAPI.html#fluidtransformssetmembershiptoarray
+     *
+     * @param {Object} input - The object to be transformed into an array.
+     * @param {Object} [transformSpec] - The options provided to the transformation rule.
```
amb26, line 492: Whilst it isn't necessary, just for clarity for those unused to the JSDoc convention we tend to write (optional) after the hyphen for optional arguments, in addition to placing their name in square brackets

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,182 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
-            var value = fluid.get(input, key, transformer.resolverGetConfig);
-            if (value === transformSpec.presentValue) {
+            // Write to output array the key of the <input> object if it's value matches the <presentValue>
+            if (fluid.get(input, key) === presentValue) {
                 outputArr.push(outputVal);
             }
         });
+
+        // Otherwise, when no <options> are provided, use the <input> object for entries in the output array.
+        if (!options) {
```
amb26, line 514: As above, I think a full if () {} else {} structure is more logically clear

dmahajan980: Hi @amb26, thanks for pointing this out. Although the current code is logically correct, I agree that it would make more sense to use an if-else statement and be more apparent to someone trying to read it.

dmahajan980:
@amb26 I have made the changes as you suggested. Please go through it once.

Thanks

the-t-in-rtf:
> Sorry to pass back on this, but I feel this suggestion has slightly reduced the clarity of the code, in that what was an else block has now received an isolated and negated test test - could I suggest putting the code back the way it was? : P

Sounds good.  We'll have to talk about the pattern in my own work next time I use it.

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,188 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            options = fluid.get(transformSpec, "options");
+
+        /**
+         * When <options> are provided in the transformation rule, use them for entries in the output array.
+         * Otherwise, use the <input> object for entries in the output array.
+         */
+        if (options !== undefined) {
+            fluid.each(options, function (outputVal, key) {
+                // Write to output array the key of the <input> object if it's value matches the <presentValue>
+                if (fluid.get(input, key) === presentValue) {
```
amb26, line 515: fluid.get is unnecessary here since you have already verified that input is an Object - and may even malfunction in the case the key name contains periods. input[key] will be more performant and correct

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,188 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+         * Otherwise, use items from the <value> array as keys.
+         */
+        if (options !== undefined) {
+            fluid.each(options, function (outPath, key) {
+                /**
+                 * Write to output object the value <presentValue> or <missingValue> depending on whether key is
+                 * found in user input.
+                 */
+                var outVal = (value.indexOf(key) !== -1) ? presentValue : missingValue;
+                fluid.set(output, outPath, outVal);
```
amb26, line 416: Similarly to below, output[outPath] = outVal is more performant and correct

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,188 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+    fluid.transforms.setMembershipToArray = function (input, transformSpec) {
+        // <input> should be an object.
+        if (!fluid.isPlainObject(input, true)) {
+            fluid.fail("setMembershipToArray didn't find object at inputPath nor passed as value.");
         }
-
-        if (transformSpec.presentValue === undefined) {
-            transformSpec.presentValue = true;
+        var outputArr = [];
+        var presentValue = fluid.get(transformSpec, "presentValue") ? transformSpec.presentValue : true,
```
amb26, line 505: This code is certainly an improvement over the previous implementation which modified the caller's arguments but my personal preference is for a transformSpec = transformSpec || {} at the start and then returning all of these to being transformSpec.options, and presentValue = transformSpec.presentValue || true etc. which are considerably more readable and performant. fluid.get is an extremely heavyweight piece of machinery and not suitable for use in low-level framework code (in practice, when the framework really needs something like the full functionality it uses various cheaper, semi-private alternatives such as fluid.getImmediate, fluid.model.getSimple etc.)

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,188 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            fluid.fail("arrayToSetMembership didn't find array at inputPath nor passed as value.");
         }
-        if (!options) {
-            fluid.fail("arrayToSetMembership requires an options block set");
-        }
-
-        if (transformSpec.presentValue === undefined) {
-            transformSpec.presentValue = true;
+        var output = {};
+        var presentValue = fluid.get(transformSpec, "presentValue") ? transformSpec.presentValue : true,
```
amb26, line 401: As below, rewrite these using direct JS logic on a defaulted transformSpec without fluid.get - since this is low-level framework code we can generally use both the performance and the readability

src/framework/core/js/ModelTransformationTransforms.js
```diff
@@ -349,86 +349,188 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            missingValue = fluid.get(transformSpec, "missingValue") ? transformSpec.missingValue : false,
+            options = fluid.get(transformSpec, "options");
+
+        /**
+         * When <options> are provided in the transformation rule, use them for keys in the output object.
+         * Otherwise, use items from the <value> array as keys.
+         */
+        if (options !== undefined) {
```
amb26, line 409: Feel totally free to ignore this suggestion, but my personal quite mild preference is to always position the positive branch of a condition first - if (option === undefined) { ... even if there is a fairly clear sense, as in this case, that the dominant logical pathway, is the other branch. I wouldn't go so far as the developer who advised this stackexchange questioner - https://softwareengineering.stackexchange.com/questions/350472/developer-insists-if-statements-shouldnt-have-negated-conditions-and-should-al - and create an entire empty branch to satisfy the preference. But if both branches have to be there, I would tend to order them the other way round. If you think the code is clearer as it stands I am perfectly happy to merge it as it is.

src/framework/core/js/ModelTransformationTransforms.js
dmahajan980: Hi @amb26, thanks for the feedback. I'll remove the usage of `fluid.get` and use the normal JavaScript pattern. I have one doubt though

> my personal preference is for a transformSpec = transformSpec || {} at the start and then returning all of these to being transformSpec.options, and presentValue = transformSpec.presentValue || true etc. which are considerably more readable and performant.

Wouldn't it be more effective if we directly access it from the argument instead of `transformSpec = transformSpec || {}` at the beginning?

amb26: What do you mean by access it directly from the argument, given that the argument might be undefined?

dmahajan980: I don't think that the `transformSpec` will be `undefined` in cases when it is used for model transformation (not sure if it is used somewhere else too). The `transformSpec` will always have at least one key namely `type`. So `transformSpec` shouldn't be `undefined` and we can use it safely.

Coming to our case, if we try to access it by `transformSpec.presentValue` and `transformSpec.missingValue`, it would simply return `undefined` in cases where it wasn't provided to the transformation rule. But `transformSpec` will never be undefined (in case of model transformations).

To verify this, I set up a model component (with subcomponents) for both the transformations and passed nothing else other than the transformation `type` to the transformation rule. I tried logging the value to the console from inside the transformation functions. It showed the following results:
transformSpec

amb26: One of the use cases for this pull was to allow the transformSpec to be omitted so that the function is useful as a standalone utility - and you have documented it as optional in your JSDoc!

dmahajan980: Sorry, I almost forgot about that use case. I now understand why we should use `transformSpec = transformSpec || {}` here Thank you for clarifying.

amb26:
Hi @dmahajan980 - thanks so much for this excellent, careful work and your great patience with the review process. This is a great improvement to Infusion which should benefit many users in future. Best of luck with your GSoC project with @the-t-in-rtf and I look forward to staying in touch!

dmahajan980:
Thank you @amb26 for your kind words and help throughout the review process. I look forward to more contributions to the Fluid Community and working with you and @the-t-in-rtf.

PR #994, opened 2020-06-11
FLUID-6288 Add Dockerfile

gtirloni:
How to test:

```
$ docker build -t infusion .
$ docker run --rm -ti -p 8000:80 infusion
$ open http://localhost:8000
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,21 @@ we use a `pretest` script to clean up previous coverage data before we run the t
 compile the actual report.  You should not need to run the `pretest` scripts manually before running either the node or
 browser tests, or to run the `posttest` scripts afterward.
 
+### Run Tests In a Docker Container
+
+You can also run the tests from a [Docker](https://docs.docker.com/get-docker) container.
+
+Once you have Docker installed, run the following commands to build a Docker image and start a container:
+
+* Build the image: `docker build -t infusion .`
+* Run the container: `docker run --name infusion -p 8000:80 infusion`
+
+Infusion will be available at [http://localhost:8000](http://localhost:8000)
+
+* To stop and remove the container: `docker rm -f infusion`
+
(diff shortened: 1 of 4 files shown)


jobara:
@gtirloni everything seems to be working. Could you please add some information to the README.md about using Docker? From time-to-time we have questions about how to actually run our code when someone does a checkout. This will be another option someone could use locally. Will the docker image update changes made, or will they need to re-run the build step?

gtirloni:
@jobara Thanks for reviewing it! I've added some instructions to the README now.

README.md
```diff
@@ -265,6 +265,21 @@ When this VM is first created, Chrome and Firefox will be upgraded to the latest
...
+* Build the image: `docker build -t infusion .`
+* Run the container: `docker run --name infusion -p 8000:80 infusion`
+
+Infusion will be available at [http://localhost:8000](http://localhost:8000)
+
+* To stop and remove the container: `docker rm -f infusion`
+
+If you make changes to Infusion, repeat the steps to build the image and start a new container.
+
 #### Coverage Reporting
```
jobara, line 283: Coverage reporting was a sub section under the VM instructions. It should either be moved ahead of the docker instructions or made as a level 3 heading so that it sits at the same level as the sections about how to run the tests.

gtirloni: Fixed, thanks. I didn't notice the extra `#` in the heading, sorry.

jobara:
@gtirloni I don't think we want the Dockerfile in the node module. What do you think? Could you please add anything that should be ignored to the .npmignore. There might be some other existing items missing from there too. If you think that there is too much cleanup to be included in this JIRA please file a new one for the task.

gtirloni:
@jobara good catch, add those!

jobara:
merged at e6d51d139d6a694bdf94bc7ec0406fa3e2bf69a6

PR #995, opened 2020-06-16
FLUID-6520 Specify browsers to run tests on

gtirloni:
Instead of specifying a growing number of browsers that we do not want to run tests on, specify the list of browsers we actually care about.

### DIFF
--- a/tests/testem.js
+++ b/tests/testem.js
@@ -38,7 +38,7 @@ fluid.defaults("fluid.tests.testem", {
         tests:   "%infusion/tests"
     },
     testemOptions: {
-        launch: "Chrome,Edge,Firefox",
+        launch: "Chrome,Firefox",
         ignore_missing_launchers: true,
         disable_watching: true,
         tap_quiet_logs: true
--- a/tests/testem.js
+++ b/tests/testem.js
@@ -38,7 +38,8 @@ fluid.defaults("fluid.tests.testem", {
         tests:   "%infusion/tests"
     },
     testemOptions: {
-        skip: "PhantomJS,Opera,Safari",
+        launch: "Chrome,Firefox",
+        ignore_missing_launchers: true,
         disable_watching: true,
         tap_quiet_logs: true
     }
--- a/tests/testem.js
+++ b/tests/testem.js
@@ -38,7 +38,7 @@ fluid.defaults("fluid.tests.testem", {
         tests:   "%infusion/tests"
     },
     testemOptions: {
-        launch: "Chrome,Firefox",
+        launch: "Chrome,Edge,Firefox",
         ignore_missing_launchers: true,
         disable_watching: true,
         tap_quiet_logs: true


amb26:
I assume we would want to run on Edge, assuming the VM was on Windows, which I assume it isn't.... but I think when we can we do actually want to start running our tests on Windows VMs since the browsers can often behave differently

gtirloni:
I'm working on adding support for running tests on Windows/macOS in addition to Linux in CI. I'll add Edge to the list, good point.

We should also add Safari but we know it won't work for now (https://github.com/testem/testem/issues/1387)

colinbdclark:
I suggested in IRC that we should simplify our browser support to omit the various Chrome forks (i.e. Edge, Opera, Brave). Do we really need to test them all, do you think, @amb26?

colinbdclark:
And is there even a workable Edge launcher for Testem yet? There wasn't last time I looked.

gtirloni:
@colinbdclark There's one now (https://github.com/testem/testem/pull/662/files) but that's a good point

colinbdclark:
Finally! It only took, what, five years? :)

gtirloni:
@colinbdclark @amb26 I was mistaken. The code I linked to only detects Edge but there's no launcher for Edge yet. See https://github.com/testem/testem/issues/996

So for now we only get Linux and Windows (no macOS) and Chrome and Firefox (no Safari, no Edge).

gtirloni:
Any concerns with merging this?

jobara:
> @colinbdclark @amb26 I was mistaken. The code I linked to only detects Edge but there's no launcher for Edge yet. See [testem/testem#996](https://github.com/testem/testem/issues/996)
> 
> So for now we only get Linux and Windows (no macOS) and Chrome and Firefox (no Safari, no Edge).

@gtirloni why not run on macOS as well? It looks like the virtual environment comes with all the major browsers now. Safari, Chrome, Firefox and Edge. I know that we are only testing on Chrome and Firefox for now, but if we are covering those on both Windows and Linux, we might as well test it on macOS too. And hopefully we'll be able to expand to run against Safari and Edge in the future.

(see: https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md)

Also looks like SafariDriver, MSEdgeDriver, ChromeDriver, and geckodriver are installed. We may want to look at those in the future for the test automation.

gtirloni:
> @gtirloni why not run on macOS as well?

My understanding was that testem is broken on Safari/macOS because it requires user interaction to continue. Is that still the case?

jobara:
> > @gtirloni why not run on macOS as well?
> 
> My understanding was that testem is broken on Safari/macOS because it requires user interaction to continue. Is that still the case?

As far as I know yes. I guess I'm asking because we could also test Chrome and FF on macOS like we're testing them on Windows.

gtirloni:
Ah, got it! Sorry, I'm a bit slow. Yeah, we could test on macOS as well. I'll update the other PR with it. Since we're specifying only Chrome/Firefox here, it should be no problem.

gtirloni:
Once this is merged, I'll pull the changes into the FLUID-6519 branch.

jobara:
Merged at 742275f3a7e0a7ba3e0fce35009096c3410eaba4

PR #997, opened 2020-06-22
FLUID-6522: Newer portions of jQuery UI plugin are not bundled correctly.

dmahajan980:
Added jQuery UI [focusable](https://api.jqueryui.com/focusable-selector/) and [tabbable](https://api.jqueryui.com/tabbable-selector/) pseudo-selectors for use in Infusion.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -326,7 +326,9 @@ module.exports = function (grunt) {
                         "node_modules/jquery-ui/ui/unique-id.js",
                         "node_modules/jquery-ui/ui/widgets/mouse.js",
                         "node_modules/jquery-ui/ui/widgets/draggable.js",
-                        "node_modules/jquery-ui/ui/widgets/tooltip.js"
+                        "node_modules/jquery-ui/ui/widgets/tooltip.js",
+                        "node_modules/jquery-ui/ui/focusable.js",
+                        "node_modules/jquery-ui/ui/tabbable.js"
                     ],
                     dest: "src/lib/jquery/ui/js/",
                     expand: true,
--- a/src/thirdPartyDependencies.json
+++ b/src/thirdPartyDependencies.json
@@ -60,7 +60,9 @@
             "./lib/jquery/ui/js/unique-id.js",
             "./lib/jquery/ui/js/mouse.js",
             "./lib/jquery/ui/js/draggable.js",
-            "./lib/jquery/ui/js/tooltip.js"
+            "./lib/jquery/ui/js/tooltip.js",
+            "./lib/jquery/ui/js/focusable.js",
+            "./lib/jquery/ui/js/tabbable.js"
         ],
         "dependencies": [
             "jQuery"


the-t-in-rtf:
@amb26 or @jobara, would either of you have time to look at this small change?  We need the missing jQuery pseudo-selectors for our gamepad navigation project.

the-t-in-rtf:
ok to test.

jobara:
@dmahajan980 and @the-t-in-rtf, I'm worried about adding more to the Infusion package. Especially for something that may not be used in Infusion itself. They aren't that large so maybe it's okay, or maybe you could make the case why it should be include with Infusion in general. As an alternative, have you tried to pull it into your package form jQuery UI directly, rather than through Infusion?

Another, longer term solution, could be to have a better build system to more specifically picked only the parts that are needed to be included in a specific build of Infusion.

dmahajan980:
Hi @jobara, when @the-t-in-rtf and I discovered the issue, we also thought of loading jQuery UI separately. But, as we discussed, we felt that it would be better if this comes with Infusion. It is because these are some of the selectors which people tend to use more often, especially if they are building a navigation tool (similar to our case) or something which relies on obtaining DOM elements that are tabbable/focusable.

I agree with your concern about space limitations and Infusion is already a large framework. One strategy I could think of is having three packages:

- Infusion without jQuery and jQuery UI
- Infusion with jQuery, but without jQuery UI
- Infusion with both jQuery and jQuery UI

I think this will provide more options to users in regards to what packages they want to include with Infusion. I would like to know yours and the @the-t-in-rtf's thoughts on the same.

jobara:
@dmahajan980 you can actually already create custom builds of the types that you are suggesting. Please see: https://github.com/fluid-project/infusion#how-do-i-create-an-infusion-package So that could be an option to do a custom build of infusion without jQuery and jQuery UI and to add those in yourself. Although it's best to at least use the same versions that would normally ship with Infusion to make sure that everything works. 

My long term suggestion was that you could be even more granular in what you choose. So you could pick and choose individual pieces of jQuery UI for example.

the-t-in-rtf:
@jobara, I know it's possible to roll custom builds, but the instructions leave a bit to be desired.  I suppose we're meant to add an npm postinstall script that takes care of the custom build?

I also think that this is a specifically useful feature in building accessible components using Infusion.  The tabbable and focusable pseudo-selectors are directly useful in building non-traditional navigation schemes.  My suspicion is also that we didn't choose not to include them so much as they showed up in an update that we missed.

the-t-in-rtf:
We are currently investigating shortcomings with tabbable and exploring options, will report back here, but it seems like we may withdraw this.

jobara:
> We are currently investigating shortcomings with tabbable and exploring options, will report back here, but it seems like we may withdraw this.

Please report on the reasoning if this is the case. It would be useful to document.

jobara:
> @jobara, I know it's possible to roll custom builds, but the instructions leave a bit to be desired. I suppose we're meant to add an npm postinstall script that takes care of the custom build?

@the-t-in-rtf and @dmahajan980 if you're pulling infusion in through NPM you can just use one of the "no-jquery" distributions. https://unpkg.com/browse/infusion@3.0.0-dev.20200326T173810Z.24ddb2718/dist/

dmahajan980:
> Please report on the reasoning if this is the case. It would be useful to document.

@jobara the problem that we noticed with both these selectors is that they didn't return the complete list of elements that are focusable/tabbable. For example, the `summary` tag and the `details` tag are not present in the array returned by them, even if it's on the webpage. It would create accessibility issues and is counterproductive to what we actually want. 

As an alternative, we found [ally.js](https://allyjs.io/) which is more accurate and offers a broad range of options and flexibility. WIth this, we would like to withdraw this pull request and close it.

PR #998, opened 2020-06-23
FLUID-6518: Add missing styles for textFieldStepper focus.

jobara:
The textFieldStepper adds focus to a parent element using a class name 
added via JavaScript. Because of the special nature, the focus styling 
needed to be explicitly added for the contrast themes.

https://issues.fluidproject.org/browse/FLUID-6518


jobara:
@cindyli could you please review this PR?

PR #999, opened 2020-07-08
FLUID-6525: Style placeholder text in contrast themes.

jobara:
https://issues.fluidproject.org/browse/FLUID-6525

### DIFF
--- a/demos/uiOptions/index.html
+++ b/demos/uiOptions/index.html
@@ -238,11 +238,11 @@
                     <form class="feedback-form">
                         <div>
                             <label for="email">Email address <span class="optional">(optional)</span></label>
-                            <input type="text" id="email" value="This form is non-functional" />
+                            <input type="text" id="email" value="" placeholder="e-mail@example.com" />
                         </div>
                         <div>
                             <label for="comments">Your comments</label>
-                            <textarea id="comments">This form is non-functional, and exists only for demonstrating Preferences Editor functionality</textarea>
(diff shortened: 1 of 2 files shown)


jobara:
@cindyli could you please review this PR. You may also want to run the results by Dana or another designer to verify that the themed placeholders look okay.

cindyli:
Based on the discussion with Justin, Dana, Gregor last Friday, the decision was to check with @lliskovoi regarding the contrast ratio between the placeholder text and the background color.

jobara:
I spoke with @lliskovoi about the contrast ratios for the placeholder text. If the placeholder text is important, i.e. not a repeat of the label, the contrast needs to pass WCAG AA. Because we don't know the nature of the placeholder text, we will need to treat it as though it is important and ensure the contrast ratios, at least for the high contrast themes. The suggestion to differentiate the text is to have the placeholder text be italicized.

jobara:
@cindyli ready for more review.

cindyli:
> Because we don't know the nature of the placeholder text, we will need to treat it as though it is important and ensure the contrast ratios, at least for the high contrast themes.
>

Thanks for the update, @jobara. Do you know what's the contrast ratio criteria for the placeholder text and the background? Is it same as the contrast ratio between the text and the background:  at least 4.5:1 but 3:1 for large text? If so, contrast ratios for some UIO themes don't meet the criteria, for example, the grey on white theme:

Screen Shot 2020-07-14 at 4 20 18 PM

jobara:
@cindyli they would need to meet the same contrast specs to pass wcag AA; however, some of the contrast themes are low contrast so they don't really follow those guidelines at all.

PR #1000, opened 2020-07-08
FLUID-6523: Prevent whitespace selection from showing play button.

jobara:
https://issues.fluidproject.org/browse/FLUID-6523


jobara:
@cindyli could you please review this PR? @BlueSlug could you please confirm that this addresses the issues you found? You should be able to test with the Orator demo.

jobara:
@cindyli I updated renamed `getSelection` to `setTextFromSelection`

BlueSlug:
> @cindyli could you please review this PR? @BlueSlug could you please confirm that this addresses the issues you found? You should be able to test with the Orator demo.

This does not seem to solve the issue in either the Orator demo page or the current `master` version of the Storytelling Tool (tested by directly editing the files in my `node_modules` dir)

jobara:
> This does not seem to solve the issue in either the Orator demo page or the current `master` version of the Storytelling Tool (tested by directly editing the files in my `node_modules` dir)

@BlueSlug It looks like the story telling tool uses infusion-all (see: https://github.com/fluid-project/sjrk-story-telling/blob/master/themes/base/storyEdit.html#L7 ). Did you modify the infusion-all file or the individual Orator files in `node_modules`?

BlueSlug:
> @BlueSlug It looks like the story telling tool uses infusion-all (see: https://github.com/fluid-project/sjrk-story-telling/blob/master/themes/base/storyEdit.html#L7 ). Did you modify the infusion-all file or the individual Orator files in `node_modules`?

We paired and tested this to make sure my approach was correct, and we determined that the source of the issue in the Storytelling Tool is the way the "Add block" buttons are defined in the markup; they're `button` elements contianing `span`s with text, so the Selection Reader is behaving properly by appearing when they're selected. The solution for the Storytelling Tool will be to find a new way to define those buttons. The biggest issue in the ST, which was triggering the FLUID-6523 bug while attempting to drag-and-drop a block, has been solved.

We also discussed the Selection Reader's behaviour on the Orator demo page, which is better but not 100% solved with this fix. It no longer shows up when selecting the Orator buttons in the "main" part of the page, but it will still appear when selecting the text in the side pane:
2020-07-15 Orator FLUID-6523 fix oddness

Some progress has been made, either way :)

jobara:
@cindyli  and @BlueSlug I did some more research on the last remaining portion of this issue related to the selection of the "Play" button itself. I've left a [comment](https://issues.fluidproject.org/browse/FLUID-6523?focusedCommentId=41762&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-41762) on the JIRA with details. In general there are no straightforward solutions. I've looked into two potential approaches; however neither worked. I mention a third option we could explore, but it would both be complex to implement and may result in other user issues. I wonder if we can merge in this PR as is, and leave the JIRA open for future work.

src/components/orator/js/Orator.js
```diff
@@ -909,7 +909,8 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             toggle: {
                 funcName: "fluid.orator.selectionReader.togglePlay",
                 args: ["{that}", "{arguments}.0"]
-            }
+            },
+            isWord: "fluid.textNodeParser.isWord"
```
cindyli, line 913: The regex `\S` in `isWord()` checks the existence of a character rather than a word. It's probably better to name it `containChars()`.

cindyli:
> @cindyli and @BlueSlug I did some more research on the last remaining portion of this issue related to the selection of the "Play" button itself. I've left a [comment](https://issues.fluidproject.org/browse/FLUID-6523?focusedCommentId=41762&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-41762) on the JIRA with details. In general there are no straightforward solutions. I've looked into two potential approaches; however neither worked. I mention a third option we could explore, but it would both be complex to implement and may result in other user issues. I wonder if we can merge in this PR as is, and leave the JIRA open for future work.
>

Does this remaining issue only happen in Edge as @gmoss mentioned yesterday? I tried but cannot reproduce it in chrome, firefox, safari. If so, can you update the JIRA with this information? Thanks.

jobara:
@cindyli ready for more review. I revisited the isWord name throughout and updated to make them clearer for what it was actually used for in context.

BlueSlug:
@cindyli, @jobara: I have added a comment on the Jira detailing exactly which browser versions are currently still affected.

PR #1006, opened 2020-08-04
FLUID-6537 Limit CI workflows to 20 minutes

gtirloni:
Sometimes CI jobs will hang for a while and the default timeout in GitHub Actions is 360 minutes. This consumes the total of 2000-3000 minutes/month assigned to our project and should be limited.

Most jobs take between 10-14min to run so proposing we limit them at 20min and increase when needed.


jobara:
Merged at 7e422ecd6d064f0fca8de4f9c5ee4e6619fd1e23

PR #1007, opened 2020-09-14
FLUID-6549: Rebuild contrast themes using CSS custom properties

greatislander:
This PR rebuilds UI Options contrast themes using CSS custom properties. This reduces the unminified Enactors.css file size from 90.6kB to 16.6kB.

## Implementation Notes

Instead of generating a full stylesheet for each contrast theme, [`build-themes-Enactors`](https://github.com/greatislander/infusion/blob/fc15184f29da2ebcf774454bdf819d1369c548d3/src/framework/preferences/css/stylus/utils/Themes.styl#L63) now generates a generic stylesheet for a document with a class which contains `fl-theme` using CSS custom properties for all of the values that can be modified by a contrast theme.

The selector for a document with a contrast theme applied is as follows:

```css
[class*="fl-theme"]:not(.fl-theme-prefsEditor-default)
```

This uses a CSS [attribute selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#Syntax) to select elements with a class which contains the string `fl-theme` and uses the [negation pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:not) (`:not`) to exclude elements with the `fl-theme-prefsEditor-default` class.

Contrast theme stylesheets themselves now consist of [the following](
https://github.com/greatislander/infusion/blob/fc15184f29da2ebcf774454bdf819d1369c548d3/src/framework/preferences/css/stylus/utils/Themes.styl#L29-L38) (for example):

```css
.fl-theme-gd {
    --fColor: #888888;
    --bColor: #222222;
    --linkColor: #8080ff;
    --disabledColor: #cc0000;
    --selectedForegroundColor: #ffffff;
    --selectedBackgroundColor: #008000;
    --buttonForegroundColor: #ffffff;
    --buttonBackgroundColor: #000000;
}
```

There are some special cases that impact how these custom properties are applied.

1. Link colours, button foreground and background colours:

Because not all themes supply values for link colours and button foreground/background colours, these properties are set using CSS custom properties ability to fall back to a default value. For example, a link colour is set like this:

```css
a {
    color: var(--linkColor, var(--fColor));
}
```

This syntax tells the browser to fall back to the foreground colour if a link colour is not supplied by the current theme.

2. Disabled colour, selected foreground and background colours:

These properties are used without a fallback value, which means that if a given theme does not provide a colour for disabled elements or selected element foregrounds/backgrounds, these elements' colours will not be modified (custom properties without values are ignored by the browser).


amb26:
Hey there @greatislander - this is really great work, I'd hadn't read about CSS custom properties (caveman me) and they look like a great way to organise our styling better. I'm wondering, given as I seem to understand, CSS custom properties are allocated from a global pot of names, whether we should adopt some policy on namespacing them.

greatislander:
Hi @amb26, that makes lots of sense to me. I'll adjust the property names to include a `--fl` prefix for consistency.

greatislander:
@amb26 @jobara I've made this change and also fixed an issue where the updated styles where incorrectly targeting documents with the `fl-theme-prefsEditor-default` class.

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -1,250 +1,374 @@
...
-        selectedBColor = contrastThemes[themeName].selectedBackgroundColor;
-
-
-        {themeName} {
-            // General Styling
-            &, *, .fl-preview-theme&,
+        &::selection,
+        ::selection {
+            color: var(--fl-selectedFColor) !important;
+            background-color: var(--fl-selectedBColor) !important;
```
jobara, line 91: Is there a typo here, should it be:
```suggestion
            background-color: var(--fl-selectedBgColor) !important;
```

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -1,250 +1,374 @@
...
-            if (selectedFColor && selectedBColor) {
-                &::selection,
-                ::selection {
-                    color: selectedFColor !important;
-                    background-color: selectedBColor !important;
-                }
+            &::selection,
+            ::selection {
+                color: var(--fl-selectedBColor) !important;
+                background-color: var(--fl-selectedFColor) !important;
```
jobara, line 294:
```suggestion
                color: var(--fl-selectedBgColor) !important;
                background-color: var(--fl-selectedFgColor) !important;
```

src/framework/preferences/css/stylus/utils/Themes.styl
jobara, line 91:
```suggestion
            color: var(--fl-selectedFgColor) !important;
            background-color: var(--fl-selectedBgColor) !important;
```

src/framework/preferences/css/stylus/SeparatedPanelPrefsEditor.styl
```diff
@@ -63,6 +62,8 @@
     }
 
     .fl-prefsEditor-showHide {
+        background-color: #fff;
+        color: #2f2b2a;
```
jobara, line 66: should these be using the variables? The styling for the reset button makes use of the variables.

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -1,250 +1,374 @@
...
+    --fl-buttonBgColor: #ffd18a;
 }
 
-build-themes-separatedPanelPrefsEditor(contrastThemes) {
-    for themeName in contrastThemes {
-        fColor = contrastThemes[themeName].foregroundColor;
-        bColor = contrastThemes[themeName].backgroundColor;
-
-        {themeName} {
-            background-image: none !important;
```
jobara, line 103: I believe this is still needed. If you look at the UIOptions demo the background image for the page is still in effect after the contrast theme is applied. See the [demo](https://build.fluidproject.org/infusion/demos/uiOptions/) from the build site as a comparison.

jobara:
@greatislander I noticed that your more recent commit logs are not following the Infusion convention of "FLUID-6549: ...". Would you be able to update your previous commit messages?

src/framework/preferences/css/stylus/utils/Themes.styl
greatislander: @jobara Resolved in 4345c44b50466de62f05d47ccdb696365ea50c3c.

greatislander:
> @greatislander I noticed that your more recent commit logs are not following the Infusion convention of "FLUID-6549: ...". Would you be able to update your previous commit messages?

My bad @jobara, this is done now.

src/framework/preferences/css/stylus/SeparatedPanelPrefsEditor.styl
greatislander: Maybe, I just stopped at recreating the stylus logic from here with custom properties: https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/css/stylus/utils/Themes.styl#L90-L116

greatislander: Resolved in 289c3f6d1.

greatislander:
@jobara I think I've addressed all of your review comments now.

jobara:
@greatislander could you please review the imports of the stylus utility files. It appears that Helpers.styl is not actually used for anything and could be removed completely. The Themes.styl file may only be needed by Enactors.styl.

jobara:
@greatislander could you update the [README.md](https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/css/stylus/README.md) file located in the stylus directory. It has information about how to build themes. I wonder if it also makes sense to push this up to the main README or maybe link to it from there.

src/framework/preferences/css/stylus/utils/Themes.styl
```diff
@@ -1,250 +1,376 @@
...
-            .fl-prefsEditor-separatedPanel {
-                .fl-panelBar, .fl-prefsEditor-reset {
-                    get-colors-for-separatedPanelPrefsEditor(fColor, bColor);
-                }
-                .fl-prefsEditor-buttons { //FLUID-4991: The shadow effect around the hide/show preferences bug fix.
-                    box-shadow:none;
-                }
-            }
+// For the general theme styling that applies to html elements and infusion components
+build-themes-Enactors() {
```
jobara, line 63: It doesn't appear that this actually needs to be a function any more. If it is included as function just to provide a way to import the colour variables independently, we should split them into a separate file. If the reason is about importing this particular bit of styling to a specific location within the enactors file, I believe that Stylus files can be imported into any location, only CSS files have to be imported at the top.

jobara: Speaking of the Enactors.styl file, we actually may want to provide options for using the styling for enactors independently. That is, there may be cases where the themes aren't used, but the rest of the enactors are. We don't want to have to include all those styles unnecessarily. Similar for the other preferences. I've written up [FLUID-6552](https://issues.fluidproject.org/browse/FLUID-6552) for that.

greatislander: Thanks, @jobara. Resolved your first comment in a1b6fba.

greatislander:
> @greatislander could you please review the imports of the stylus utility files. It appears that Helpers.styl is not actually used for anything and could be removed completely. The Themes.styl file may only be needed by Enactors.styl.

I removed `Helpers.styl` in 0a66481.

I tested and removed unnecessary imports of `utils/Themes.styl` in e6b0ab0.

greatislander:
> @greatislander could you update the [README.md](https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/css/stylus/README.md) file located in the stylus directory. It has information about how to build themes. I wonder if it also makes sense to push this up to the main README or maybe link to it from there.

Updated the readme and linked it from the main README in 16cd410.

src/framework/preferences/css/stylus/Enactors.styl
```diff
@@ -27,8 +25,395 @@
     content: "·";
 }
 
-// Theming
-build-themes-Enactors(themes);
+// General theme styling that applies to html elements and infusion components.
+// !important is used for all styles to ensure that Preferences Framework's enactors are able to override the pages default styles.
+[class*="fl-theme"]:not(.fl-theme-prefsEditor-default) {
```
jobara, line 30: What I was suggesting with [FLUID-6552](https://issues.fluidproject.org/browse/FLUID-6552) was to allow for the composition of enactor styles into this single file or output as individual CSS files. In that sense I'm not sure moving this styling form the Themes.style buys much. Unless you think we'll have more cases where we want to include the variables without related styling.

src/framework/preferences/css/stylus/README.md
```diff
@@ -4,25 +4,21 @@ The "[stylus](./)" directory contains Stylus files for generating the Preference
...
 ```stylus
-    contrastThemes = {
-        "theme-selector": {
-            foregroundColor: "#000000",
-            backgroundColor: "#ffffff"
-        }
-        // ...
+    .fl-theme-selector { // The CSS class that the Preferences Framework uses to enable this contrast theme.
+        --fl-fgColor: #000000;
+        --fl-bgColor: #ffffff;
```
jobara, line 12: Should probably add a comment next to these to clarify that they are for foreground and background colours respectively.

src/framework/preferences/css/stylus/README.md
```diff
@@ -4,25 +4,21 @@ The "[stylus](./)" directory contains Stylus files for generating the Preference
...
+Define your own theme variables in `utils/Themes.styl` using the following approach.
 
 ```stylus
-    contrastThemes = {
-        "theme-selector": {
-            foregroundColor: "#000000",
-            backgroundColor: "#ffffff"
-        }
-        // ...
+    .fl-theme-selector { // The CSS class that the Preferences Framework uses to enable this contrast theme.
```
jobara, line 10: I'm not sure that it's clear here that `.fl-theme-selector` should be replaced with the name of your own theme, and that `fl-theme-` is the conventional prefix with the `selector` part being replaced with the specific theme name.

README.md
```diff
@@ -312,3 +312,5 @@ them when they are changed. This task can be run using the following command:
 ```bash
 grunt watch:buildStylus
 ```
+
+For more information on styling the Preferences Framework, [see here](./src/framework/preferences/css/stylus/README.md).
```
jobara, line 316: Instead of "see here" I wonder if there is more descriptive text that can be used. Possibly the following:

```suggestion
For more information on styling the Preferences Framework, please review the [using Stylus with the Preferences Framework](./src/framework/preferences/css/stylus/README.md) documentation.
```

src/framework/preferences/css/stylus/Enactors.styl
greatislander: My thinking here was that if we do split enactors styling into individual stylesheets we'd need the ability to import the variables into individual enactor stylesheets as well as a combined stylesheet like the current `Enactors.css`. I think it makes sense to address those structural changes in work on [FLUID-6552](https://issues.fluidproject.org/browse/FLUID-6552) rather than here.

jobara: @greatislander I think at this point it might be better to keep the contrast theme styles in Themes.styl file that way at least there is an option to build the contrast styles exclusive of the other enactors. Which is a step towards the goals of [FLUID-6552](https://issues.fluidproject.org/browse/FLUID-6552). I get your point about wanting more flexibility with the colour variables. I think eventually we'll also want those in a separate file too, that way you could import them into other styl files or use a completely different set of colour variables if you wanted. Anyways, I think in general not moving the styles out of Themes.styl minimizes the changes in this PR and the broader structural changes can take place in FLUID-6552 or other future JIRAs as needed.

greatislander: Makes sense! I'll move them back.

src/framework/preferences/css/stylus/PrefsEditor.styl
```diff
@@ -334,4 +329,26 @@
...
+    .fl-prefsEditor {
+        // ON/OFF switch
+        .fl-switchUI-control:focus {
+            box-shadow: 0 0 0 0.2em var(--fl-bgColor),
+                0 0 0 0.35em var(--fl-fgColor);
+        }
+
+        .fl-switchUI-controlKnob {
+            background-color: var(--fl-fgColor) !important;
+        }
```
jobara, line 342: @greatislander the switch component is actually an Infusion component that can be used outside of the prefs editor. ([Switch Demo](https://build.fluidproject.org/infusion/demos/switch/)). With this change of moving the theming into the PrefsEditor styles, we'd need to have a prefs editor on the same page for the switches to be themed. However, there is a possibility that the page uses the UI Enhancer to style the page but does not include the prefs editor on the page. An example of that can be seen in the full page prefs editors ([example](https://build.fluidproject.org/infusion/examples/framework/preferences/minimalFootprint/)), where the preference editor is located on a separate page. This might be for performance reasons, moving the preferences to a user account page, and etc. I think for now we should move these styles back into the Themes.styl file, that way they are always available with the enactors. However, this may still not be the best factoring and we should consider how to improve things, with the work on [FLUID-6552](https://issues.fluidproject.org/browse/FLUID-6552).

greatislander: @jobara If I add fallbacks (`#fff` for `--fl-bgColor` and `#000` for `--fl-fgColor`), the switch component will work in its default styling without any theme being applied. Let me know if that would work to address this issue.

jobara: @greatislander I don't think so. I believe the issue is a bit different. It's the case of theming the switch components in the absence of the prefs editor. The enacting and setting of preferences can happen independently.

greatislander: Resolved this in https://github.com/fluid-project/infusion/pull/1007/commits/94eacc6cb898acf9b3af6f619dd38bad63e683c6.

greatislander: Added this: https://issues.fluidproject.org/browse/FLUID-6553

jobara:
Merged at 4e0124e9af46da914f1365d2aed117b749195cde

greatislander:
Thanks @jobara!

PR #1008, opened 2020-09-17
Fluid-6538: Update to jQuery 3.5.1

jobara:
Will update to the jQuery 3.5.1 and address breakages such as with fluid.focus and fluid.blur.

https://issues.fluidproject.org/browse/FLUID-6538

Requires: https://github.com/fluid-project/fluid-grunt-lint-all/pull/15


jobara:
@amb26 I believe I've updated all of the deprecated jQuery function calls.

jobara:
@amb26 I've updated from main, and ready for review.

amb26:
Why were all event bindings converted to the "on" form? It seems that, e.g. click() still exists and isn't deprecated
https://api.jquery.com/click/

src/components/reorderer/js/Reorderer.js
```diff
@@ -508,7 +508,7 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
         var handleClick = function (evt) {
             var handle = fluid.unwrap(thatReorderer.dom.fastLocate("grabHandle", this));
             if (fluid.dom.isContainer(handle, evt.target)) {
-                $(this).focus();
+                $(this).trigger("focus");
```
amb26, line 511: There are now several inconsistent means of updating focus in this file - 
i) $(el).focus();
ii) $(el).trigger("focus")
iii) fluid.focus(el)
we should normalise these so that we only use one style - presumably iii). 
See notes on line 594 below

src/components/reorderer/js/Reorderer.js
```diff
@@ -595,7 +591,7 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                 var handlePos = fluid.dom.computeAbsolutePosition(handle);
                 var handleWidth = handle.offsetWidth;
                 var handleHeight = handle.offsetHeight;
-                item.focus();
+                item.trigger("focus");
```
amb26, line 594: This is an interesting corner where start and stop have become more asymmetric. I imagine it might create problems if we made this function async like stop, but it might be worth trying to see what happens. At the very least, we could convert this to a fluid.focus() call to make it consistent with stop, even if something goes wrong if we try to await() it

src/framework/core/js/Fluid.js
```diff
@@ -1851,6 +1851,9 @@ var fluid = fluid || fluid_3_0_0; // eslint-disable-line no-redeclare
 
     // A function to tag the types of all Fluid components
     fluid.componentConstructor = function () {};
+    Object.defineProperty(fluid.componentConstructor, "name", {
```
amb26, line 1854: Explain why we did this because I already barely remember!

src/framework/core/js/FluidDocument.js
```diff
@@ -151,22 +151,27 @@ var fluid_3_0_0 = fluid_3_0_0 || {}; // eslint-disable-line no-redeclare
...
 
-    // These function (fluid.focus() and fluid.blur()) serve several functions. They should be used by
-    // all implementation both in test cases and component implementation which require to trigger a focus
-    // event. Firstly, they restore the old behaviour in jQuery versions prior to 1.10 in which a focus
-    // trigger synchronously relays to a focus handler. In newer jQueries this defers to the real browser
-    // relay with numerous platform and timing-dependent effects.
-    // Secondly, they are necessary since simulation of focus events by jQuery under IE
-    // is not sufficiently good to intercept the "focusin" binding. Any code which triggers
-    // focus or blur synthetically throughout the framework and client code must use this function,
-    // especially if correct cross-platform interaction is required with the "deadMansBlur" function.
```
amb26, line 162: Probably worth restoring the last line of this comment which I imagine is still valid - "Any code which ..."

src/framework/core/js/Fluid.js
amb26: Looks like the reason was so that we can detect an Infusion component in src/framework/core/js/jquery.keyboard-a11y.js without having to depend on Fluid.js

tests/component-tests/reorderer/js/ImageReordererTests.js
```diff
@@ -60,120 +60,120 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
...
-        function verticalNavigationTest(lightbox, upEvt, downEvt) {
+        async function verticalNavigationTest(lightbox, upEvt, downEvt) {
             // setup: force the grid to have four columns
             var lightboxRoot = fluid.jById(fluid.testUtils.imageReorderer.imageReordererRootId);
 
             lightboxRoot.removeClass("width-3-thumb");
             lightboxRoot.addClass("width-4-thumb");
 
-            fluid.testUtils.imageReorderer.focusLightbox();
+            await fluid.testUtils.imageReorderer.focusLightbox();
```
amb26, line 70: A good example of minimal violence

jobara:
> Why were all event bindings converted to the "on" form? It seems that, e.g. click() still exists and isn't deprecated
> https://api.jquery.com/click/

It was flagged by the jQuery Migrate tool, where it is marked as deprecated. https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfnclick-event-shorthand-is-deprecated

I think it's safer to just follow the migration tool in this regard because other using Infusion may also be using the migration tool and run into this as well. Additionally, it keeps things more consistent with the way other events are used.

src/components/reorderer/js/Reorderer.js
jobara: Doing a straight conversation to `fluid.focus(item)` seems to be okay. Making `start` async and using `await` breaks the interaction. I'll just convert and not use async/await here.

src/components/reorderer/js/Reorderer.js
jobara: I don't think case i) occurs. What you might have seen is an instance of calling focus on a regular DOM node. At any rate, I'll explore converting all of these to using `fluid.focus`.

src/components/reorderer/js/Reorderer.js
jobara: Actually it might be working, I accidentally left in a call to trigger("focus"). Will explore some more.

jobara: Okay there is a small breakage when using async/await. Using the layout reordered demo if you move some items and then click on some empty space, you can end up not being able to position some of the reorder able elements. This doesn't seem to happen with just `fluid.focus` and no async/await.

jobara:

src/components/reorderer/js/Reorderer.js
amb26: Thanks - could you add a code comment before the method reporting that it breaks if made async

jobara:
@amb26 further follow up on the deprecations. It seems like the docs are out of date.
https://github.com/jquery/jquery/issues/3214#issuecomment-398663964

The above thread is interesting. Basically some of the shorthands are deprecated, but there is no intention to remove them in the short term. However, their use is strongly discouraged.

src/components/reorderer/js/Reorderer.js
jobara: @amb26 I've added a comment. This is ready for more review.

demos/keyboard-a11y/js/five-star.js
```diff
@@ -84,25 +84,28 @@ var demo = demo || {};
             onCreate: [{
                 "this": "{that}.stars",
                 method: "mouseover",
-                args: {
-                    expander: {
-                        funcName: "demo.fiveStar.makeStarHandler",
-                        args: ["{that}", "{that}.hoverStars"]
+                args: [
```
amb26, line 87: This change doesn't look right - you've changed the args list but left the method as "mouseover"

jobara: You're right. Thanks for catching that. I've pushed up a fix.

PR #1010, opened 2020-09-30
FLUID-6554: Handle soft hyphens in content

jobara:
https://issues.fluidproject.org/browse/FLUID-6554

### DIFF
--- a/src/framework/preferences/js/SyllabificationEnactor.js
+++ b/src/framework/preferences/js/SyllabificationEnactor.js
@@ -46,7 +46,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         },
         markup: {
             separator: "<span class=\"flc-syllabification-separator fl-syllabification-separator\"></span>",
-            softHyphenPlaceholder: "<span class=\"flc-syllabification-softHyphenPlaceholder\"></span>"
+            softHyphenPlaceholder: "<span class=\"flc-syllabification-softHyphenPlaceholder fl-syllabification-softHyphenPlaceholder\"></span>"
         },
         model: {
             enabled: false
--- a/src/framework/preferences/js/SyllabificationEnactor.js
+++ b/src/framework/preferences/js/SyllabificationEnactor.js
@@ -46,7 +46,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
         },
         markup: {
             separator: "<span class=\"flc-syllabification-separator fl-syllabification-separator\"></span>",
-            softHyphePlaceholder: "<span class=\"flc-syllabification-softHyphenPlaceholder\"></span>"
+            softHyphenPlaceholder: "<span class=\"flc-syllabification-softHyphenPlaceholder\"></span>"
         },
         model: {
             enabled: false
@@ -158,7 +158,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
                     "{arguments}.0",
                     "{arguments}.1",
                     "{that}.options.markup.separator",
-                    "{that}.options.markup.softHyphePlaceholder"
+                    "{that}.options.markup.softHyphenPlaceholder"
                 ]
             },
             injectScript: {
@@ -379,16 +379,16 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
 
     /**
      * Insert separators, indicating the hyphenated positions, into a text node. This will inject the specified
-     * markup as the separators and split the original textnode at the hpyhenation points. If there are no places to
+     * markup as the separators and split the original textnode at the hyphenation points. If there are no places to
      * inject a hyphenation point, the node will be left unchanged. If the original textnode inclues soft hyphens (i.e.
      * &shy;) characters, these will be used as a hyphenation point with the original location in the text node
-     * replacded with the `softHyphenPlaceholderMarkup` to allow it to be replaced when syllabification is removed.
+     * replacded with the `softHyphenPlaceholderMarkup` to allow it to be restored when syllabification is removed.
      *
      * @param {HypherHyphenator} hyphenator - an instance of a Hypher Hyphenator
      * @param {DomNode} node - a DOM node containing text to be hyphenated
      * @param {String} separatorMarkup - the markup to be injected and used to indicate the separators/hyphens
      * @param {String} softHyphenPlaceholderMarkup - the markup to be injected and used to indicate the original
-     *                                               location of any soft hyphens inclded in `node`
+     *                                               location of any soft hyphens included in `node`
      */
     fluid.prefs.enactor.syllabification.hyphenateNode = function (hyphenator, node, separatorMarkup, softHyphenPlaceholderMarkup) {
         if (!hyphenator || !node.textContent) {
@@ -428,7 +428,7 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             parent.normalize();
         });
 
-        // replace placeholders with soft hyphens
+        // restore soft hyphens
         that.locate("softHyphenPlaceholder").each(function (index, elm) {
             var parent = elm.parentNode;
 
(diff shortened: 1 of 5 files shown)


jobara:
@cindyli and @amb26 further updates to better handle syllabification with soft hyphens in the original content. Could you please review this PR.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -350,63 +357,83 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+        node.parentNode.insertBefore(toInsert, node);
+        return node;
+    };
+
+    /**
+     * Insert separators, indicating the hyphenated positions, into a text node. This will inject the specified
+     * markup as the separators and split the original textnode at the hpyhenation points. If there are no places to
+     * inject a hyphenation point, the node will be left unchanged. If the original textnode inclues soft hyphens (i.e.
+     * &shy;) characters, these will be used as a hyphenation point with the original location in the text node
+     * replacded with the `softHyphenPlaceholderMarkup` to allow it to be replaced when syllabification is removed.
```
amb26, line 385: Typos hpyhenation and replacded. You probably want the last "replaced" to be "restored" for a bit of extra clarity

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -350,63 +357,83 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
      *
      * @param {HypherHyphenator} hyphenator - an instance of a Hypher Hyphenator
      * @param {DomNode} node - a DOM node containing text to be hyphenated
      * @param {String} separatorMarkup - the markup to be injected and used to indicate the separators/hyphens
+     * @param {String} softHyphenPlaceholderMarkup - the markup to be injected and used to indicate the original
+     *                                               location of any soft hyphens inclded in `node`
```
amb26, line 391: Typo inclded

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -37,14 +37,16 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+            softHyphenPlaceholder: ".flc-syllabification-softHyphenPlaceholder"
         },
         strings: {
             languageUnavailable: "Syllabification not available for %lang",
             patternLoadError: "The pattern file %src could not be loaded. %errorMsg"
         },
         markup: {
-            separator: "<span class=\"flc-syllabification-separator fl-syllabification-separator\"></span>"
+            separator: "<span class=\"flc-syllabification-separator fl-syllabification-separator\"></span>",
+            softHyphePlaceholder: "<span class=\"flc-syllabification-softHyphenPlaceholder\"></span>"
```
amb26, line 49: Typo name softHyphePlaceholder which at least was copy and pasted correctly at line 161 : P

jobara:
@amb26 my spelling woes continue. Thanks for spotting these. I've pushed up fixes, ready for more review.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -350,63 +357,83 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
         fluid.each(segs, function (seg) {
-            // Do not attempt to process empty segs, which could lead to an infinite recursion.
+            // If the text content starts with a soft hyphen (\u00AD, &shy;) replace it with an empty placeholder; which
+            // is used to put the soft hyphen back if syllabification is disabled.
             // See: https://issues.fluidproject.org/browse/FLUID-6554
-            if (seg.length) {
-                var separator = $(separatorMarkup)[0]; // converts the separatorMarkup markup string into a Dom Node
-                node = node.splitText(seg.length);
-                node.parentNode.insertBefore(separator, node);
+            if (node.textContent.startsWith("\u00AD")) {
```
cindyli, line 410: One problem is, when a soft hyphen is in a middle of a word, syllabification breaks the word into 2 syllables at where the soft hyphen is regardless if these 2 sections are proper syllables. 

For example, `back` is one single syllable. When a soft hyphen is in between `ba` and `ck`, hypher will mark them as 2 separate syllables, which might be misleading.

I was thinking about removing soft hyphens from texts before running them thru the hypher and inserting soft hyphens back. But the difficult part is how to keep track of positions of those soft hyphens.

src/framework/preferences/js/SyllabificationEnactor.js
```diff
@@ -350,63 +357,83 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
...
+                // converts the softHyphenPlaceholderMarkup markup string into a Dom Node
+                var placeholder = $(softHyphenPlaceholderMarkup)[0];
+                var newNode = fluid.prefs.enactor.syllabification.insertIntoTextNode(node, placeholder, 1);
```
cindyli, line 413: One issue with replacing soft hyphens with something else is hyphenation that indicates the join of the word is gone when the "syllable" setting is turned on:
Screen Shot 2020-10-01 at 4 24 26 PM

If the soft hyphen is still in place, the hyphenation persists when the "syllable" is on:
Screen Shot 2020-10-01 at 4 24 41 PM

amb26: Very good remarks, @cindyli!

src/framework/preferences/js/SyllabificationEnactor.js
amb26: One suggestion I can think of is to simply clone any DOM that the soft hyphens are removed from and hold the original, together with a pointer to its parent, it in a private fragment out of the document context. This can then be swapped back in again when the enactor is turned off.

cindyli: Removing soft hyphens without inserting them back when the "syllable" setting is on will lead to the issue below that hyphenations are gone.

jobara: @cindyli I can see your point. However, the ability to break on soft hyphens provides a means for an integrator to define where the syllables are split. Unfortunately the hyphenation engine we use it's 100% perfect. If I remember correctly, it's closer to 80% accurate. So providing an ability for an integrator to explicitly define where breaks should happen may be useful. Also for novel words or localization specific pronunciation that we may not cover. 

I'd suggest not using soft hyphens where a word wouldn't have a natural hyphenation point. Although I'm not sure if there are reasons that would contradict this nor is it something we could enforce in anyway. 

More thoughts/comments welcome.

src/framework/preferences/js/SyllabificationEnactor.js
jobara: @cindyli I've pushed up some changes which add a soft hyphen to the placeholder pseudo element. Please let me know if this addresses the issue.

src/framework/preferences/js/SyllabificationEnactor.js
amb26: I think it's reasonable to imagine that turning on the syllabification enactor represents the user's intention "I would like to substitute UIO's idea for where syllables should break for the original page author's, if any". So long as we restore the markup to its original condition when the enactor is deactivated, I think this will be accepted. Of course it would be desirable if we could somehow incorporate the "hint" from the markup author provided by the soft hyphen but this doesn't seem to be possible with hypher.

src/framework/preferences/js/SyllabificationEnactor.js
jobara: @cindyli pointed found a bug in the updated implementation if the soft hyphen appears in the last segment, it doesn't get replaced with the placeholder. This is because the algorithm removes the last segment as we don't want to add a separator after it. However, we do need to replace the soft hyphen in all segments.

jobara:
@cindyli and @amb26 ready for more review.

cindyli:
Could you add tests to make sure all soft hyphens, including when it's on the last syllable segment, are replaced by the soft hyphen placeholder? Thanks.

PR #972, opened 2019-11-26
FLUID-6145: Fully working immutable options world

amb26:

PR #1013, opened 2020-10-06
FLUID-6548: Update linting for Infusion based on new rules and linter

jobara:
Includes an update to use fluid-lint-all for linting.

https://issues.fluidproject.org/browse/FLUID-6548


src/framework/core/js/Fluid.js
```diff
@@ -205,27 +203,26 @@ var fluid = fluid || fluid_3_0_0;
...
         var activity = fluid.makeArray(fluid.describeActivity()); // Take copy since we will destructively modify
         fluid.popActivity(activity.length); // clear any current activity - TODO: the framework currently has no exception handlers, although it will in time
         if (fluid.failureEvent) { // notify any framework failure prior to successfully setting up the failure event below
-            fluid.failureEvent.fire(args, activity);
+            fluid.failureEvent.fire(messages, activity);
         } else {
-            fluid.logFailure(args, activity);
-            fluid.builtinFail(args, activity);
+            fluid.logFailure(messages, activity);
+            fluid.builtinFail(messages, activity);
```
jobara, line 218: @amb26 I want to highlight this change to make sure you see it. There is another similar one elsewhere in the code as well. The change here is to use the [rest parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) in stead of arguments. The rest parameter returns a named array of the arguments from that position.

src/framework/core/js/Fluid.js
```diff
@@ -2321,26 +2364,27 @@ var fluid = fluid || fluid_3_0_0;
...
      * fine-grained policy. If it is an Object, the values may also themselves be EL paths
      * representing that a default value is to be taken from that path.
-     * @param {...Object} options1, options2, .... - an arbitrary list of options structure which are to
+     * @param {...Object} sources - an arbitrary list of options structure which are to
      * be merged together. These will not be modified.
      */
 
-    fluid.merge = function (policy /*, ... sources */) {
-        var sources = Array.prototype.slice.call(arguments, 1);
+    fluid.merge = function (policy , ...sources) {
```
jobara, line 2387: @amb26 this is another case where I've converted to use a [rest parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters). In this case the other benefit is that the first parameter called `policy` is not included in the `sources` array, so we don't need to handle skipping it later in the code.

src/framework/core/js/Fluid.js
amb26: I assume that the use of the spread pattern makes the use of fluid.makeArray() unnecessary on the first line

jobara: @amb26 yes, the call to `fluid.makeArray` that had been on line 214 was removed for this reason. The call to `fluid.makeArray` on line 212 (formerly 215) seems be used for making a copy.

the-t-in-rtf:
The pull against fluid-grunt-lint-all associated with this has been merged and I have cut the 2.0.0 release.   Please update this pull to use the released version, and then we should be good to merge.

greatislander:
@jobara This one needs an update as well.

jobara:
> @jobara This one needs an update as well.

Was this in reference to a specific dependency or just about updating from "main"?

greatislander:
Updating from `main`.

greatislander:
The changes in [FLUID-6573](https://issues.fluidproject.org/browse/FLUID-6573) ([pull request](https://github.com/fluid-project/infusion/pull/1032)) should get the build working again here.

jobara:
The Codecov check for this PR is interesting. It seems the patch is only getting 72.72% for coverage, but for the project the coverage actually increased.

greatislander:
> The Codecov check for this PR is interesting. It seems the patch is only getting 72.72% for coverage, but for the project the coverage actually increased.

So I guess the overall lines of code decreased and thus the project coverage is higher overall, but the changed lines only have 72.72% coverage so that check is failing?

amb26:
I found that our current coverage thresholds are a little aggressive - my pull https://github.com/fluid-project/infusion/pull/972 similarly had the patch miss by 0.09% and was so thrown out ... even though, again, coverage improved overall. It might be an idea to replace the test with a simple check that coverage has not reduced overall or at least below 88%

greatislander:
We can disable the patch check if desired (or simply lower the threshold for patches) — see docs: https://docs.codecov.io/docs/commit-status#branches

jobara:
> We can disable the patch check if desired (or simply lower the threshold for patches) — see docs: https://docs.codecov.io/docs/commit-status#branches

That's an option. @amb26, @the-t-in-rtf, @cindyli any thoughts?

amb26:
I think eliminating the patch check is my favoured option - there are lots of reasons why a patch in itself may not appear to have adequate coverage (e.g. it consists mostly of deletions of uncovered code!)

jobara:
> I think eliminating the patch check is my favoured option - there are lots of reasons why a patch in itself may not appear to have adequate coverage (e.g. it consists mostly of deletions of uncovered code!)

I'm disabling the patch checks in this PR.

jobara:
@amb26 this PR is ready for review.

PR #1014, opened 2020-10-09
FLUID-6555: Update CSS to address linting issues with Stylelint config

greatislander:
Prerequisite for fluid-project/fluid-grunt-lint-all#20.


demos/pager/css/pager.css
```diff
@@ -35,43 +35,58 @@
...
     color: #343499;
-    background-color: #E6E7E8;
 }
 
 .demo-pager-container td {
-    padding: .5em .2em;
+    padding: 0.5em 0.2em;
 }
 
-.demo-pager-head-select { width:5%; }
```
the-t-in-rtf, line 46: Is there a rule that reports single line entries like this?  I see you've expanded lots of these, and yet there's one that is still there.

tests/framework-tests/preferences/css/test-theme.css
```diff
@@ -1,17 +1,17 @@
...
+    font-size: 24px;
 }
 
 span {
-  font-size: 12px;  /* Setting font size */
+    font-size: 12px;  /* Setting font size */
 }
 
-.sheep { cue: url( 'http://sheep.org' ); }
+.sheep { cue: url('http://sheep.org'); }
```
the-t-in-rtf, line 17: I'm confused as to whether single-line entries are checked by the new linting.  Why did you expand the other entries but not this one?

.stylelintrc
```diff
@@ -0,0 +1,15 @@
+{
```
the-t-in-rtf, line 1: Why is this file duplicated from the same file in [the fluid-grunt-lint-all pull](https://github.com/fluid-project/fluid-grunt-lint-all/pull/20/files)?  Is there any way to reused or `extend` the configuration used in that package instead?

the-t-in-rtf:
I reviewed the changes made to all files, in general I see what I'd expect from reading the rules, i.e. the changes fix white-space (including indenting), single-colons for `::before` and `::after`, parameter ordering.  The end result seems more consistent and readable.  I did have some small questions for you, but on the whole it looks like the new rules are helpful and not too disruptive.

.stylelintrc
greatislander: I think much in the way that the Fluid eslint configuration is installed as a package which is extended within Infusion's `.eslintrc` file, we'd need to create and publish a package (e.g. `stylelint-config-fluid`) which could be extended in `.stylelintrc`.

the-t-in-rtf: I was thinking we could resolve the path to one stored in fluid-grunt-lint-all, but I think your proposal is better given that we're trending towards improving support for (and presumably relying more on) central configurations like this.  

Can you file a ticket for that work?  I imagine you'd probably be best to write it up slash work on the actual thing.

I was going to say that we could do that later and avoid having three related pulls open at once, but I think any change to a central config is going to necessitate at least a small pull against fluid-grunt-lint-all and infusion.  If it's not too much work, I think it'd be cleaner to have a common config as part of the initial effort to standardise our approach to CSS.

greatislander: @the-t-in-rtf I haven't filed an issue (not sure where I should do that, to be honest) but I started working on this in https://github.com/greatislander/stylelint-config-fluid. I'll try to get it wrapped up this week.

the-t-in-rtf:
The pull against fluid-grunt-lint-all associated with this has been merged and I have cut the 2.0.0 release.   Please update this pull to use the released version, and then we should be good to merge.

greatislander:
@the-t-in-rtf Updated, but now the JS errors are showing up— how do you want me to handle that? Should we consolidate the changes from here into #1013 or vice versa?

the-t-in-rtf:
I guess @amb26 will need to finish reviewing and merge Justin's pull and you'll need to merge with upstream master.

greatislander:
This will need more work once #1024 is merged as we'll be able to lint the unprocessed Sass now as well.

jobara:
@greatislander I've updated https://github.com/fluid-project/infusion/pull/1013 to use fluid-lint-all. I disabled the stylelint check for now. You'll either need to pull in my changes and update the .fluidlintallrc.json file, or would you prefer that I merge in your changes from this PR into my mine?

greatislander:
@jobara My preference would be that you merge #1013 and then I update this PR with the changes from main.

jobara:
> @jobara My preference would be that you merge #1013 and then I update this PR with the changes from main.

Works for me. Thanks.

jobara:
@greatislander https://github.com/fluid-project/infusion/pull/1013 has been merged to main. You can update this PR now from main. Please add in the stylelint check for fluid-lint-all.

greatislander:
@jobara My updated configuration doesn't seem to be picking up the `.scss` files, see comment here: https://matrix.to/#/!JcoHDrfLedPQdFhAXn:matrix.org/$H_GltIQC2J4Fo0Em55rbEYHM8x2KgSo31go7SVzmtCU?via=matrix.org

greatislander:
NB: This branch is not ready to merge; blocked by https://github.com/fluid-project/fluid-lint-all/issues/12.

the-t-in-rtf:
@greatislander, there is a new release of `fluid-lint-all` with the fix for the default options merging and a new `showCheckedFiles` flag to assist in future troubleshooting.

greatislander:
@the-t-in-rtf The same issue appears to be happening even with the new `fluid.flatten` configuration. With an unmodified https://github.com/fluid-project/fluid-lint-all/blob/4cba8be7579aafeaf207fb00593c3a6bf5fc7f52/src/js/lint-all.js#L203, only CSS files are linted (there are 50). If I change the array order to put `.scss` first on that line of `fluid-lint-all`, only the 9 SCSS files are linted.

the-t-in-rtf:
Yes, I made the same expander definition error in the longer form.  I have [a second pull that actually fixes the problem](https://github.com/the-t-in-rtf/fluid-lint-all/pull/new/GH-12-1), there's a dev release, `1.0.4-dev.20210129T151721Z.b9ddeaa.GH-12-1`.

greatislander:
@amb26 @jobara This PR is now done and can be merged, with many thanks to @the-t-in-rtf!

src/components/overviewPanel/css/OverviewPanel.css
```diff
@@ -22,12 +22,17 @@
...
     font-weight: bold;
+    src: url('../../../lib/roboto/fonts/Roboto-Slab-Bold.woff');
 }
 
-
 .fl-overviewPanel-container {
+    font-family: "OpenSans", "Helvetica Neue", "Helvetica", Arial, sans-serif;
+    font-size: 1em;
+    height: 100%;
+    left: 0;
```
jobara, line 33: It seems that the reordering of the properties has moved some positioning properties away from the comment. For example `left: 0;` is now above the comment and `position: fixed;` and `top: 0;` are below. I think it would make sense to move the comment above the declaration and reword as necessary. Another option would be to remove the comment completely if you don't find it necessary. I'm open to other options as well.

src/components/pager/css/Pager.css
```diff
@@ -3,67 +3,84 @@
  * Data and UI separated into 2 parts: Layout and Color Themes
  */
 
-/***********************************************************/
+/**********************************************************   */
```
jobara, line 6: Does it have to have the space between the `*` at the end? If so, it might be better to just remove this comment break if it isn't really needed.

src/components/pager/css/Pager.css
```diff
@@ -3,67 +3,84 @@
...
-    color:#000;
-    text-decoration:none;
-    cursor:default;
+    background-color: transparent;
+    color: #000;
+    cursor: default;
+    text-decoration: none;
 }
 
+/**********************************************************   */
```
jobara, line 69: Similar to previous comment

src/components/pager/css/Pager.css
```diff
@@ -3,67 +3,84 @@
...
 
-/***********************************************************/
 /** THEMES - High Contrast **/
+
 /* Pager data */
 .fl-theme-bw .fl-pager .fl-pager-data {
+    /* Is this required? */
```
jobara, line 75: Is this a comment that you've added or was it moved from somewhere else?

src/components/reorderer/css/Reorderer.css
```diff
@@ -19,55 +21,56 @@
 
 /* automatically applied as the mouse hovers over re-orderable item */
 .fl-reorderer-movable-hover {
-    background-color: lightyellow; /* !important;;*/
+    background-color: lightyellow; /* !important;; */
```
jobara, line 24: This seems to be an odd comment. I wonder if we should just remove it.

src/components/textfieldControl/css/TextfieldSlider.css
```diff
@@ -23,78 +23,83 @@ http://brennaobrien.com/blog/2014/05/style-input-type-range-in-every-browser.htm
...
-  color: transparent;
+    animation-duration: 0.2s;
+    background: transparent;
+    border-color: transparent;
+    border-width: 0.5rem 0;
+    color: transparent;
+    height: 1rem;
 }
+
 .fl-textfieldSlider input[type=range].fl-slider::-ms-fill-lower {
```
jobara, line 83: someday we should review all these vendor prefixes.

src/framework/enhancement/css/ProgressiveEnhancement.css
```diff
@@ -1,4 +1,8 @@
 /* Progressive Enhancement: JS will reverse the display setup if it is enabled */
-.fl-progEnhance-enhanced {display:none}
+.fl-progEnhance-enhanced {
+    display: none;
+}
 
-.fl-progEnhance-basic {}
+.fl-progEnhance-basic {
+    /* Is this required? */
+}
```
jobara, line 8: I think we should probably remove this. @amb26 and @colinbdclark do you recall if there was a specific reason for the empty declaration?

src/components/pager/css/Pager.css
greatislander: I added it— I wasn't sure why the empty block was there.

src/components/reorderer/css/Reorderer.css
greatislander: I think so. I'm not sure why it's there.

src/components/textfieldControl/css/TextfieldSlider.css
greatislander: Ideally we could write unprefixed CSS and use Autoprefixer (see: [FLUID-6566](https://issues.fluidproject.org/browse/FLUID-6566)).

jobara:
merged at 97c2fcc179cf71bcc2ce9af26d8750d4a8d80c69

PR #1016, opened 2020-10-26
FLUID-6539: use native button element for switch

greatislander:
This PR refactors the Switch component to use a native `<button>` element. Three tests were removed:

- The test for the `tabindex`, which is no longer needed
- The two tests for triggering the switch with <kbd>space</kbd> and <kbd>enter</kbd>

In testing with @jobara we found that the latter tests didn't work; triggering a `keydown` event on a `<button>` using jQuery does not trigger the `click` handler bound to the button (see this behaviour reproduced: https://codepen.io/greatislander/pen/bGeqmrR). We felt that removing these tests was acceptable since the native browser behaviour of the button is adequately tested by way of the click event.


jobara:
Disabled state for switches will be worked on in [FLUID-6563](https://issues.fluidproject.org/browse/FLUID-6563).

jobara:
merged at daeff32bd67a593b4f5819b314752d970c1e6cc0

PR #1017, opened 2020-10-27
NOJIRA: exclude additional files from the NPM package

greatislander:
This PR excludes a few additional files from the NPM package:

- `.eslintignore`
- `.eslintrc.json`
- `.nycrc`
- `Gruntfile.js`
- `Vagrantfile`
- `provisioning/`
- `tests/`


amb26:
Unfortunately excluding all of "tests" is overly aggressive and will make the package useless for a number of its current purposes - since the entire Infusion testing system is in there including jqUnit, the IoC testing framework, etc. 

You should cut down on the excludes so that they retain tests/lib and tests/test-core - looking at the original set of excludes I think they were probably nearly correct in this area and just needed to be tweaked to include more recent additions such as bundle-tests

This isn't the best factoring of our layout since logically the "tests" area should simply contain tests proper, and is something we should deal with when rationalising Infusion's layout (e.g. when we reorganise it into modules where each module's individual tests accompany it)

greatislander:
@amb26 thanks, updated.

jobara:
@greatislander I thought I had written up a JIRA for cleaning this up, but I can't find it now. Thanks for catching it. In addition to @amb26's comments regarding the tests. I have a few other modification requests:

- Remove trigger-ci file from the repository
- Add to npmignroe
  - .github
  - CONTRIBUTING.md ( is there a reason to keep this in the npm package? )

greatislander:
@jobara You'll see that I removed the trigger-ci file in this PR. I can add `.github`; a nice solution to `CONTRIBUTING.md` might be to move it and `CODE_OF_CONDUCT.md` into `.github`. Community health files are properly recognized if they're in `.github` and then by ignoring that directory they could be excluded from the package as well. Thoughts?

greatislander:
@jobara I also wonder if we still need the `.project` file?

jobara:
> @jobara You'll see that I removed the trigger-ci file in this PR. I can add `.github`; a nice solution to `CONTRIBUTING.md` might be to move it and `CODE_OF_CONDUCT.md` into `.github`. Community health files are properly recognized if they're in `.github` and then by ignoring that directory they could be excluded from the package as well. Thoughts?

If they work from the .github directory I think that's fine. Although I'm trying to recall now if we wanted them to have more visibility in the repo itself. If we do move them to the .github repo you could actually probably just omit the CODE_OF_CONDUCT as it will be supplied by default from the .github repo in our organization.

jobara:
> @jobara I also wonder if we still need the `.project` file?

@amb26 do you use the .project file for your IDE?

jobara:
@greatislander could you please update this with "main"? There are merge conflicts at the moment.

greatislander:
@jobara ready for another look.

jobara:
@greatislander could you please update this with main. There are merge conflicts again.

greatislander:
@jobara Updated.

.npmignore
```diff
@@ -3,27 +3,23 @@
...
-tests/configs/
-tests/framework-tests/
-tests/manual-tests/
-tests/node-tests/
-tests/test-core/testTests/
-tests/all-tests.html
-tests/testem.js
+tests/
+!tests/lib
+!tests/test-core
```
jobara, line 25: I ran `npm pack` to see how things will be treated. However, it appears that the tests are all being included, not just tests/lib and tests/test-core

greatislander: Resolved in 2ba94f9.

.npmignore
```diff
@@ -20,6 +20,4 @@ examples/
 instrumented/
 products/
 reports/
-tests/
-!tests/lib
-!tests/test-core
+tests/*-tests
```
jobara, line 23: This is leaving in all-tests.html and testem.js, both of those should also be removed.

greatislander:
@jobara this is ready for another review.

.npmignore
```diff
@@ -3,27 +3,22 @@
...
 .eslintignore
 .eslintrc.json
-.gitignore
 .github
+.gitignore
 .npmignore
 .nycrc
 .project
 .settings
+.vagrant
```
jobara, line 12: this should no longer exist in the project after https://github.com/fluid-project/infusion/pull/1029

jobara:
merged at 8dfc99ed2c0c8f7e28b90419862f0d7f752c7cc8

PR #1019, opened 2020-10-27
NOJIRA: update engines to supported Node version

greatislander:
Related to https://issues.fluidproject.org/browse/FLUID-6556, this PR updates the `engines` configuration in `package.json` to indicate that we only support Node >= 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
         "lint": "grunt lint"
     },
     "engines": {
-        "node": ">=0.10.0"
+        "node": ">=12"
     },
     "dependencies": {
         "fluid-resolve": "1.3.0"


greatislander:
@amb26 @the-t-in-rtf I spoke about this with @jobara and he suggested that this may cause problems with older Node projects that use Infusion. Not sure if that'll be the case but wanted to check with you.

amb26:
Given that the default behaviour of failing an engine version check is just a warning - https://www.stefanjudis.com/today-i-learned/prevent-npm-install-for-not-supported-node-js-versions/ -  I'm happy with updating this to node 12. In practice, the projects that can't upgrade node versions are unlikely to want to upgrade Infusion either.

greatislander:
@jobara This is ready for review.

gtirloni (approved):

PR #1021, opened 2020-10-27
FLUID-6565: Prevent form submission from clicking switches in prefs editor

jobara:
https://issues.fluidproject.org/browse/FLUID-6565


greatislander (approved):
Good catch, thanks.

cindyli (approved):

PR #1023, opened 2020-10-28
FLUID-6567: run tests conditionally

greatislander:
This PR addresses [FLUID-6567](https://issues.fluidproject.org/browse/FLUID-6567) by:

1. Splitting code linting and tests into two separate workflows;
2. Only runing linting on `ubuntu-latest` with Node 14.x;
3. Only triggering tests when paths matching the following rules are modified in the commit or pull request:

- `package.json`
- `src/**/*.html`
- `src/**/*.js`
- `src/**/*.json`
- `tests/**/*.html`
- `tests/**/*.js`
- `tests/**/*.json`

This is a starting point and can certainly be improved based on more in-depth examination of the files which might impact test outcomes (it's likely that the paths can be made more specific).


.github/workflows/ci-lint.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+defaults:
+  run:
+    shell: bash
```
gtirloni, line 11: This can be removed since it's only using Linux. It was needed for macOS/Windows

.github/workflows/ci-test.yml
```diff
@@ -1,17 +1,33 @@
...
   pull_request:
     branches: [ main ]
+    paths:
+        - 'package.json'
+        - 'src/**/*.html'
+        - 'src/**/*.js'
+        - 'src/**/*.json'
+        - 'tests/**/*.html'
+        - 'tests/**/*.js'
+        - 'tests/**/*.json'
```
jobara, line 23: This duplication is unfortunate and error prone. Can you see if there is a way around it, or if it is just a limitation of GitHub Actions at the moment?

.github/workflows/ci-lint.yml
```diff
@@ -0,0 +1,40 @@
+name: Code linting
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  lint:
```
jobara, line 10: Should probably add a timeout to this task as well.

greatislander: It takes 60 seconds so I don't think that's necessary.

.github/workflows/ci-test.yml
greatislander: It's a limitation at the moment.

.github/workflows/ci-lint.yml
greatislander: (See here: https://github.com/fluid-project/infusion/pull/1023/checks?check_run_id=1321259265)

jobara: The purpose would be to prevent a stalled run from hanging for a day or so. This way it forces the run to stop early. The time length can be lower than the 20 min that is set for the other tests. I believe @gtirloni picked that number as it gave time for the tests to complete successfully.

gtirloni: +1 -- it's just to shield us from that because we have limited build minutes to use every month and I've seen my fair share of weird problems. It's good to have an upper limit (even if it's 20x the usual time).

.github/workflows/ci-test.yml
jobara, line 23: I think we should also add the Gruntfile.js to the list of tracked files. Possibly also the .nycrc file as that configures the code coverage reporting.

jobara:
Merged at fa21180c08becc8be94c6fd333ef331226a473f5

PR #1024, opened 2020-10-28
FLUID-6496: resolve build errors, migrate from Stylus to Sass

greatislander:
I investigated this issue following original investigation by @jobara and @cindyli. They had correctly identified the initial problem as the upgrade from Grunt 1.0.2 to Grunt 1.1.0 in 5f7b28236c47ba2a42749af81d886040a37e67c6, specifically the replacement of Grunt's old `file.mkdir` function with [mkdirp](https://www.npmjs.com/package/mkdirp). On further investigation I found that the root cause of the problem was the official [grunt-contrib-stylus](https://github.com/gruntjs/grunt-contrib-stylus) plugin, which has not received updates since 2016.

I explored the possibility of patching `grunt-contrib-stylus` or creating a fork, but ultimately decided to migrate the handful of stylesheets written in Stylus to Sass, using the [grunt-dart-sass](https://github.com/laurenhamel/grunt-dart-sass) plugin, which uses the Sass project's primary [Dart Sass](https://sass-lang.com/dart-sass) implementation (the official [grunt-contrib-sass](https://github.com/gruntjs/grunt-contrib-sass) plugin still requires the presence Ruby Sass on the host system, which reached [end of life as of March 2019](https://sass-lang.com/ruby-sass)).

This necessitated a couple of changes, mostly involving the logic in the [fonts partial](https://github.com/fluid-project/infusion/blob/944aa2d2ce3d2cde0cdfcf19ff0cde593a367e17/src/framework/preferences/css/sass/utils/_fonts.scss) and the replacement of Stylus [subscript operators](https://stylus-lang.com/docs/operators.html#subscript-) with [conventional selectors](https://github.com/fluid-project/infusion/pull/1024/files#diff-4b60994e1edb60c625d3850bf54da1ef2355b1fda36aaa665bf4e3666e07cf47L305-L449). I also [renamed the two partial files](https://github.com/fluid-project/infusion/tree/944aa2d2ce3d2cde0cdfcf19ff0cde593a367e17/src/framework/preferences/css/sass/utils) in the Preferences framework to use Sass's partial convention (see [documentation on partials](https://sass-lang.com/guide#topic-4)).

As a result of this work, this pull request also resolves [FLUID-6551](https://issues.fluidproject.org/browse/FLUID-6551).


src/components/switch/css/sass/Switch.scss
```diff
@@ -8,9 +8,9 @@ knobMargin = 0.2em;
 }
 
 .fl-switchUI-control {
-    width: knobSize * 3;
+    width: ($knobSize * 3);
     border: 0.15em solid var(--fl-fgColor, currentColor);
-    border-radius: knobSize + knobMargin;
+    border-radius: ($knobSize + $knobMargin);
```
jobara, line 13: Looking at these changes, I wonder if we should just convert this to vanilla CSS and use CSS variables and `calc`. There isn't much nesting in this file either, so not sure how much it's actually benefitting from pre-processing.

src/framework/preferences/css/sass/Enactors.scss
```diff
@@ -2,7 +2,7 @@
  * General prefs editor styling used to apply preferences
  * !important is used for all styles to ensure that Preferences Framework's enactors are able to override the page's default styles.
  */
-@import "utils/Fonts"
+@import "utils/fonts";
```
jobara, line 5: It seems that `@import` is deprecated in favour of `@use`. I think we might as well switch to `@use`.

src/components/switch/css/sass/Switch.scss
greatislander: It's not really benefitting from being Sass on its own, but it's used as a partial import here: https://github.com/fluid-project/infusion/blob/944aa2d2ce3d2cde0cdfcf19ff0cde593a367e17/src/framework/preferences/css/sass/PrefsEditor.scss#L205

Sass imports don't treat CSS files as partials, so if we try to import vanilla CSS into a Sass file it is left as a regular CSS import, which won't work. This is a great case for reorganizing the stylesheets to be more modular (see: [FLUID-6552](https://issues.fluidproject.org/browse/FLUID-6552)) but I feel that it's outside the scope of this JIRA.

src/framework/preferences/css/sass/Enactors.scss
greatislander: Resolved in 3d193de26. However, the import of Switch can't be replaced with a `@use` statement in its current form.

src/components/switch/css/sass/Switch.scss
greatislander: Correction: the issue here is the file extension, so I can replace this with vanilla CSS and import it. However, this doesn't address the recommendation to move away from `@import` usage.

greatislander: Resolved in 2b06baf.

jobara:
@greatislander could you please merge with master and address the merge conflict.

greatislander:
Done, @jobara.

jobara:
Merged at 99ef9fb2012822418e800b2ad86a0808b96cc4e6

PR #1027, opened 2020-10-29
FLUID-6539: remove activateHandler from switch

greatislander:
Follow-up to #1016.

### DIFF
--- a/src/components/switch/js/Switch.js
+++ b/src/components/switch/js/Switch.js
@@ -80,21 +80,12 @@ var fluid_3_0_0 = fluid_3_0_0 || {};
             toggleModel: {
                 funcName: "fluid.switchUI.toggleModel",
                 args: ["{that}"]
-            },
-            activateHandler: {
-                funcName: "fluid.switchUI.activateHandler",
-                args: ["{arguments}.0", "{that}.toggleModel"]
             }
         }
     });
 
     fluid.switchUI.toggleModel = function (that) {
         that.applier.change("enabled", !that.model.enabled);
-    };
-
-    fluid.switchUI.activateHandler = function (event, fn) {
-        event.preventDefault();
-        fn();
     };
 
 })(jQuery, fluid_3_0_0);


jobara:
@greatislander could you commit these changes against [FLUID-6539](https://issues.fluidproject.org/browse/FLUID-6539) as it's related to [changes](https://github.com/fluid-project/infusion/pull/1016/files#diff-295d61b1fe3790604f4fcf3bf9fedb688101c6d1ce6f37186129864220ebce24) for that. By doing this it will help keep track for the reasons these changes were made.

amb26:
This is fine but we should document somewhere (probably on https://docs.fluidproject.org/infusion/development/switchapi ) the now more restricted markup expectations of the component. These were never clearly explained, but I think it is now capable of binding to fewer types of markup than before. I'm just working now on porting this to the new renderer which will help clean up such issues a bit by letting it generate its own markup, but in the meantime we should explain this textually somehow in code comments as well as docs.

greatislander:
@amb26 I'm happy to update that but there's a bit of a conundrum. Those docs are currently valid for https://www.npmjs.com/package/infusion/v/3.0.0-dev.20200326T173810Z.24ddb2718. If I make the change to the docs site indicating that the switch component can only be initiated on a `<button>`, that won't be accurate for the current release. What's the best practice given versioning, releases, etc.?

amb26:
Well, that was sort of an open discussion as part of the 11ty update of the docs pull which is looks like you are inheriting : P
I don't have any particularly clear ideas on this, but one route might be to open an "in progress" pull against the docs repository which tags the added text with a placeholder saying something like "This change applies to versions of Infusion later than xxxxxx" and once we roll the next dev release we can fill in the placeholder and merge it.

In the long run I guess I was hoping for something similar to what node does for their docs - that is i) version manage the entire docs so entire snapshots are available for each release, and in addition ii) highlight significant additions or changes permanently with the release they were added in - e.g. for https://nodejs.org/docs/latest/api/v8.html#v8_v8_getheapcodestatistics it says "Added in v12.8.0" but you can also access https://nodejs.org/docs/v12.7.0/api/v8.html and see that it is not there.
Of course such a thing is dependent on being able to move Infusion over to a recognisable versioning system and release cadence, which is dependent on a lot of other work, but it would be nice to give the best approximation of something like this we can afford in the meantime.

greatislander:
@amb26 I haven't formally inherited that PR yet but I expect I will :smile:

I guess the best approach for this specific issue might be to have a version of the docs which reflects changes in the current development version, distinct from docs pinned to each release.

jobara:
@greatislander and @amb26 
We did look a bit at using tags for versions of the docs, but if we need to go back to fix things like typos or other errors it becomes complicated. Another possibility is to branch and deploy the branches as well.

We've also talked about the live site pointing at "main" and relying on the "versioned" ones to point at the releases. I don't think we've finalized any decisions but these were avenues to explore.

jobara:
Merged at 5734c1232a1ebcbd54c59b157f6e2bced7176719

PR #1029, opened 2020-10-29
FLUID-6569: Simplify automated testing using headless browsers.

jobara:
https://issues.fluidproject.org/browse/FLUID-6569


gtirloni:
I don't develop Infusion and I was wondering how people would debug a test failure in this case. Would they just open a browser and point it at all-tests.html and that's it? Just wondering if there is any workflow that would still be reliant on a Vagrant box but I guess not.

jobara:
> I don't develop Infusion and I was wondering how people would debug a test failure in this case. Would they just open a browser and point it at all-tests.html and that's it? Just wondering if there is any workflow that would still be reliant on a Vagrant box but I guess not.

Typically if there is a test failure I'd run it directly in a browser, as it's easier to debug that way.

gtirloni (approved):

greatislander (approved):
👍 This is great!

cindyli (approved):

PR #1031, opened 2020-11-03
FLUID-6571: generate a manifest of demos, examples and tests

greatislander:
- Running `npm run buildManifest` will write the manifest to `build/manifest.json` and write a simple index file to `build/index.html`.
- Running `npm run buildLatest` will generate a build of Infusion (using `grunt --source && grunt copy:all`) and write the manifest; this command is now used by Netlify to build the latest version from [`main`](../tree/main/).

## Other Notes

To allow names for demos, examples, and tests to be derived from the filesystem, some files in the `manual-tests` and `test-core` directories have been renamed. Where needed, references to these files have been updated elsewhere.


greatislander:
Not sure what's happening with the tests. I'll take a look shortly.

netlify.toml
```diff
@@ -0,0 +1,6 @@
+[build]
+  command = "npm run buildLatest"
+
+[[redirects]]
+  from = "/"
+  to = "https://build.fluidproject.org/"
```
jobara, line 6: I'm assuming this would cause the deploy preview to redirect to the build site. Is that correct? If so, I think it would still take away from the usefulness of the build preview. I can see how this would make sense for the deploy from main though.

greatislander: Oh, of course. Okay. I'm going to add some code to generate a static HTML file in the `buildManifest` script. Will update the PR later or tomorrow morning with that functionality.

greatislander: Added index generation in bf420ba.

greatislander:
@jobara A test failed for some reason while running `npm install`; can you restart the failed/cancelled tests?

jobara:
@greatislander maybe we should hold off on this for now, at least until we get the linting update merged in. It seems that you made some changes due to newer syntax failing linting, but that should be resolved soon. Also I think it would be preferable to reduce our dependencies, e.g. lodash, and instead make use of native JavaScript features. e.g. [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

greatislander:
> @greatislander maybe we should hold off on this for now, at least until we get the linting update merged in. It seems that you made some changes due to newer syntax failing linting, but that should be resolved soon. Also I think it would be preferable to reduce our dependencies, e.g. lodash, and instead make use of native JavaScript features. e.g. [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

Sounds good; I would much prefer to use newer syntax and template literals.

bin/build-manifest.js
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * Clean up output to ensure that special terms, acronyms,
+ * et cetera are properly formatted.
+ *
+ * @param {String} str Input string on which to perform find and replace.
+ * @return {String} The input string with replacements applied.
+ */
+var replaceValuesInString = function (str) {
+    var replacementPatterns = {
+        "a 11 y": "a11y",
```
amb26, line 18: How do these strange split-up strings arise?

bin/build-manifest.js
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * Determine the URL of a demo, example, or test based on file path.
+ *
+ * @param {String} path The file path of a demo, example, or test.
+ *
+ * @return {URL} The URL of the demo, example, or test.
+ */
+var getUrl = function (path) {
+    return new URL("https://build-infusion.fluidproject.org/" + path);
```
amb26, line 51: If this runs as part of a netlify deploy, shouldn't this URL point to the site's own structures? Might be a better place for a relative URL, perhaps. If we're not proposing moving the build site itself into this structure, it might be better to migrate this entire logic into the build site.

bin/build-manifest.js
greatislander: @amb26: It's a result of using `lodash.lowercase` to convert file and directory names into sentences. Here's an example (check the console): https://codepen.io/greatislander/pen/LYZBqyg

bin/build-manifest.js
greatislander: The manifest file will be ingested by the build.fluidproject.org site so it requires full URLs; you make an excellent point about the index.html file though. It should use relative URLs.

PR #1032, opened 2020-11-05
FLUID-6573: replace uglify and concat tasks with grunt-terser

greatislander:
This PR uses [terser](https://www.npmjs.com/package/terser) via [grunt-terser](https://www.npmjs.com/package/grunt-terser) to replace the `uglify` and `concat` Grunt tasks.

### Implementation Notes

- Terser uses `output.beautify` to output unminified JS, `buildSettings.expanded` and is used to set this value (other parameters that rely on `buildSettings.compress` have been changed to use `buildSettings.expanded`).
- Terser uses `output.preamble` instead of `banner` for opening comments.


Gruntfile.js
```diff
@@ -364,12 +364,11 @@ module.exports = function (grunt) {
                 }]
             }
         },
-        uglify: {
+        terser: {
             options: {
-                banner: "<%= banner %>",
```
jobara, line 369: It looks like terser calls this option `preamble`. see: https://github.com/terser/terser/blob/6b6cb9ff01ab5d94e7fa3bc32981fdcab83474a4/README.md#format-options

package.json
```diff
@@ -51,13 +51,13 @@
         "grunt": "1.3.0",
         "grunt-contrib-clean": "2.0.0",
         "grunt-contrib-compress": "1.6.0",
-        "grunt-contrib-concat": "1.0.1",
         "grunt-contrib-copy": "1.0.0",
         "grunt-contrib-uglify": "4.0.1",
```
jobara, line 55: grunt-contrib-uglify should be removed from the dependencies

Gruntfile.js
greatislander: Aha! I tried to find a reference but couldn't. I'll set that up.

package.json
greatislander: Resolved in d6090f6.

Gruntfile.js
greatislander: Resolved in a9356f0 and 69e9a16.

greatislander:
@jobara Can you trigger the tests again? One of them failed for mysterious reasons :(

jobara:
@greatislander we spoke last week about the source versions not working. Have you had a chance to explore or research that more? I believe the last we talked about would be the option to beautify the minified output, but I was concerned that it wasn't a true source build as it was still processed through the minifier.

greatislander:
@jobara I wasn't able to find any other approaches using terser. Something else that occurred to me is that it may be desireable to process the ES6 syntax for compatibility with older browsers. Is that something that's been discussed? It should probably be looked at as part of any changes to the build process.

jobara:
> @jobara I wasn't able to find any other approaches using terser. Something else that occurred to me is that it may be desireable to process the ES6 syntax for compatibility with older browsers. Is that something that's been discussed? It should probably be looked at as part of any changes to the build process.

No, I don't believe we've discussed that yet. Although I don't believe we currently support browser that would require transpiling to ES5.

greatislander:
@jobara noted (https://github.com/fluid-project/infusion/pull/1034#discussion_r521343594) that we need to make sure that the beautified version preserves comments.

greatislander:
> @jobara noted ([#1034 (comment)](https://github.com/fluid-project/infusion/pull/1034#discussion_r521343594)) that we need to make sure that the beautified version preserves comments.

@jobara: this is resolved in cfa3953.

jobara:
@greatislander I noticed that the whitespace formatting in the bundles are a bit off. It does make it harder to read, but maybe it isn't a huge issue. Example:

```javascript
    // Renders an activity stack in a form suitable to be sent to a modern browser's console
    // unsupported, non-API function
        fluid.renderActivity = function(activityStack, renderer) {
        renderer = renderer || fluid.renderOneActivity;
        return fluid.transform(activityStack, renderer);
    };
```

I poked around in the issues for Terser. A few things to note:

-  grunt-terser is using an old version of terser. It's at 4.3.9 and the latest is 5.5.1. A lot has [changed](https://github.com/terser/terser/blob/master/CHANGELOG.md) in between but I'm hoping none of those will affect us at the moment.
- it seems like beautify is going to be [deprecated](https://github.com/terser/terser/issues/599#issuecomment-599046945).
- there was a suggestion there to use [Prettier](https://prettier.io) which is something we could also look at in the future.
- another option in the future is to only build minified bundles and require source maps for debugging them.

amb26:
Perhaps another bit of impetus to ditch grunt

jobara:
As per @greatislander's suggestion I ran a diff between the source versions when building infusion-all from main and from this PR using Terser. Note in our current main, the source is not run through any minifier and just concatenated independently. In this PR, the output is always run through Terser and beautified for the source version.

[changes.diff.zip](https://github.com/fluid-project/infusion/files/5624808/changes.diff.zip)

Some of the changes I found when run through Terser are as follows:
- whitespace changes including converting tabs to spaces and other indentations
- inline comments moved to a new line
- functions declared in an argument are wrapped in `()`. e.g `myFunc(function () {return "test";});` becomes `myFunction((function() {return "test";}));`
- large numbers use scientific notation. e.g. `200000` becomes `2e5`. This did take place for minified files with uglify.
- single expressions distributed across multiple lines may be collapsed to a single line.
- extraneous braces meant for clarity may be removed from expressions. e.g. `return (fileUploadLimit === that.options.queueSettings.fileUploadLimit);` becomes `return fileUploadLimit === that.options.queueSettings.fileUploadLimit;`
- objects will be written with each property on a new line
- arrays written on a single line
- arrays have spaces around members. e.g. `["a", "b"]` becomes `[ "a", "b" ]`
- comment within an object literal are removed.
- unnecessary quotes from object keys are removed
- `()` removed when constructing a an instance without any arguments e.g. `new FormData();` becomes `new FormData;`. This also happened in uglify.

I only made it through about 2000k lines of the diff going in reverse order. I'm wondering though based on that sample how worth it is to produce a concatenated source in this fashion, as it's suitably different than the actual source. In particular shifting inline comments to a new line can confuse which line the comment is actually documenting. Would it be better to just output minified versions and rely on source maps for debugging. There are cases when using via as a node dependency where you may want to modify the file in place, in this case a source bundle would still be helpful. Should we return to using concat to concatenate the files instead of beautifying the minified source. If we use terser and beautify, should we change the name from source to beautified or something similar to indicate that it isn't really the source? @amb26, @cindyli , @greatislander what are your thoughts on this?

greatislander:
@jobara My thoughts are that we should switch back to using concat for source, use terser for the minified version only, and work on replacing Grunt with Gulp (and gulp-concat).

However, I'm curious about this comment:

> There are cases when using via [npm] as a node dependency where you may want to modify the file in place, in this case a source bundle would still be helpful.

So you'd install infusion with NPM, then copy the source files into your project so you can edit them? What would happen if the version in `node_modules` needed to be updated from a newer build? Just trying to understand the use case.

amb26:
> However, I'm curious about this comment:

There are cases when using via [npm] as a node dependency where you may want to modify the file in place, in this case a source bundle would still be helpful.

So you'd install infusion with NPM, then copy the source files into your project so you can edit them? What would happen if the version in node_modules needed to be updated from a newer release? Just trying to understand the use case.

Personally when I need to modify a file in place from a node dependency I simply modify the original source file. This is one reason why demos and other kinds of developer artefacts are coded against the original sources rather than bundles. I don't see a particular use case for a "readable source bundle" - either the code is bundled and unreadable/unmodifiable except via source map or it isn't.

jobara:
> @jobara My thoughts are that we should switch back to using concat for source, use terser for the minified version only, and work on replacing Grunt with Gulp (and gulp-concat).
> 
> However, I'm curious about this comment:
> 
> > There are cases when using via [npm] as a node dependency where you may want to modify the file in place, in this case a source bundle would still be helpful.
> 
> So you'd install infusion with NPM, then copy the source files into your project so you can edit them? What would happen if the version in `node_modules` needed to be updated from a newer release? Just trying to understand the use case.

Good question. This is usually just for quickly testing out a change where I can't or don't want to make a dev release or push to a GitHub branch. Also Infusion doesn't support npm-link, if I remember correctly, because of the dedupe-infusion. So in this use case I'm just modifying in the node_modules directory directly and just re-installl to toss the changes. 

Another case, in the chrome extensions I've had issues where the source-maps just don't appear/work at all.

greatislander:
> Also Infusion doesn't support npm-link, if I remember correctly, because of the dedupe-infusion.

Can you tell me more about this? I haven't come across it in the documentation.

amb26:
>> Also Infusion doesn't support npm-link, if I remember correctly, because of the dedupe-infusion.

> Can you tell me more about this? I haven't come across it in the documentation.

Some not terribly detailed JIRAs on this topic - https://issues.fluidproject.org/browse/FLUID-6140 , https://issues.fluidproject.org/browse/FLUID-5822 . By its nature, Infusion is a global programming environment and needs to ensure that there is exactly one version of itself loaded in an entire application, regardless of npm's opinion on the matter. The module driver of every Infusion inspects the whole tree at startup and automatically defers to the one at the highest level found in the filesystem.

jobara:
> > Also Infusion doesn't support npm-link, if I remember correctly, because of the dedupe-infusion.
> 
> Can you tell me more about this? I haven't come across it in the documentation.

I believe it has something to do with infusion deduping not being able to traverse across symlinks. Infusion needs a single instance that is run in node, so it removes the others, or something to that effect. @amb26  would be able to explain better.

amb26:
> @jobara My thoughts are that we should switch back to using concat for source, use terser for the minified version only, and work on replacing Grunt with Gulp (and gulp-concat).

I'm fine with this - although I don't mind about the differences in the source build. Should we push this forwards, or should we try to make a start on eliminating grunt? My preference is to move to pure npm scripts rather than to move to gulp.

greatislander:
@amb26 @jobara Thinking about this more, I think we should merge this. We know we want to revisit/remove Grunt altogether in the future but I think getting the uglify/concat tasks replaced will fix the build and let us move forward with [FLUID-6538](https://github.com/fluid-project/infusion/pull/1008), [FLUID-6548](https://github.com/fluid-project/infusion/pull/1013) and [FLUID-6555](https://github.com/fluid-project/infusion/pull/1014), giving us the new/updated linting capabilities while still allowing us to create custom builds. Thoughts?

PR #1033, opened 2020-11-09
FLUID-6574: use build for all browser tests

greatislander:
This PR modifies browser tests to use the uncompressed build of Infusion produced by `grunt --source` instead of individual dependency files to ensure that tests pass with a built version of Infusion`.

(Specific dependencies from the old test files can be identified by reviewing test files at [8dfc99e](https://github.com/fluid-project/infusion/tree/8dfc99ed2c0c8f7e28b90419862f0d7f752c7cc8/tests). The Orator test requires an additional dependency, [`src/components/textToSpeech/js/MockTTS.js`](https://github.com/fluid-project/infusion/blob/4f51ed963ef0eac6da725e656a2ea84b41415211/src/components/textToSpeech/js/MockTTS.js), which is not included in the Infusion build and perhaps should be moved into the test framework.)

This PR also addresses [FLUID-6575](https://issues.fluidproject.org/browse/FLUID-6575), adding NPM scripts to facilitate running tests and builds, even in instances where `grunt-cli` is not globally installed. They are as follows:

- `deps`: run `grunt loadDependencies`
- `build:all`: run `grunt`
- `build:all:source`: runs`grunt --source=true`
- `build:custom`: run `grunt custom`
- `build:custom:source`: run `grunt custom --source=true`
- `build:dists`: run `grunt buildDists`
- `build:sass`: run `grunt buildSass`
- `watch:sass`: run `grunt watch:buildSass`
- `test`: prepare and run browser and Node tests sequentially
- `test:browser:prepare`: run `build:all:source` to build file for browser tests
- `test:browser:run`: run browser tests
- `test:browser`: prepare _and_ run browser tests
- `test:bundles:prepare`: run `build:dists` to build files for browser bundle tests
- `test:bundles:run`: run browser bundle tests
- `test:bundles`: prepare _and_ run browser bundle tests
- `test:node`: run Node tests
- `test:report`: generate test reports

_Additional parameters may be passed to any build script after a double hyphen, for example:_

```bash
npm run build:custom -- --exclude="jquery"
```

Additionally, scripts have been updated to run binaries (`nyc`, `rimraf`, and `testem`) by name (e.g. `testem` instead of `node node_modules/testem/testem.js`).


jobara:
@greatislander regarding the comment in the description about MockTTS.js it is something I had considered in the past. I think at the time the tests weren't included in the release, and it was something that would be needed by packages using Infusion to write their own tests against TTS implementations. I still don't think we include the component tests in the npm package, and it wouldn't make sense in the general tests as it requires the TTS component. It's something that would be clearer when we split things off into smaller packages. But let me know if you have more thoughts/ideas about it.

jobara:
@greatislander where does this PR stand in relation to @amb26's [comments](https://issues.fluidproject.org/browse/FLUID-6574?focusedCommentId=42072&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-42072) on the [FLUID-6574](https://issues.fluidproject.org/browse/FLUID-6574) should this be closed, mothballed, or something else?

amb26:
Closed for inactivity, please feel free to reopen and resync if we'd like to do something similar

PR #1034, opened 2020-11-09
FLUID-6575: add NPM scripts to facilitate running tests and builds

greatislander:
This PR resolves [FLUID-6575](https://issues.fluidproject.org/browse/FLUID-6575), adding NPM scripts to facilitate running tests and builds, even in instances where `grunt-cli` is not globally installed. They are as follows:

- `deps`: run `grunt loadDependencies`
- `build:all`: run `grunt`
- `build:all:source`: runs`grunt --source=true`
- `build:custom`: run `grunt custom`
- `build:custom:source`: run `grunt custom --source=true`
- `build:dists`: run `grunt buildDists`
- `build:sass`: run `grunt buildSass`
- `watch:sass`: run `grunt watch:buildSass`
- `start`: host the working directory using [serve](https://www.npmjs.com/package/serve)
- `test`: prepare and run browser and Node tests sequentially
- `test:browser`: run browser tests
- `test:bundles:prepare`: run `build:dists` to build files for browser bundle tests
- `test:bundles:run`: run browser bundle tests
- `test:bundles`: prepare _and_ run browser bundle tests
- `test:node`: run Node tests
- `test:report`: generate test reports

_Additional parameters may be passed to any build script after a double hyphen, for example:_

```bash
npm run build:custom -- --exclude="jquery"
```

Additionally, scripts have been updated to run binaries (`nyc`, `rimraf`, and `testem`) by name (e.g. `testem` instead of `node node_modules/testem/testem.js`).


package.json
```diff
@@ -19,18 +19,26 @@
...
+        "pretest": "rimraf reports/* coverage/*",
+        "test": "run-s test:browser test:node",
+        "test:browser": "testem ci --file tests/testem.js",
+        "test:bundles": "run-s test:bundles:*",
+        "test:bundles:prepare": "npm run build:dists",
+        "test:bundles:run": "testem ci --file tests/bundle-tests/testem.js",
+        "test:node": "nyc node tests/node-tests/basic-node-tests.js",
+        "test:report": "nyc report -r text-summary -r html -r json",
+        "posttest": "npm run test:report",
+        "build:all": "grunt",
```
amb26, line 33: This is a great set of improvements, thanks. If this is the new preferred way of triggering builds, we should update the README section named "Package Types" to account for it. I think it would also make sense to add a new README section describing some of the other scripts, e.g. "deps"

package.json
```diff
@@ -19,18 +19,26 @@
...
+        "build:all:source": "grunt --source=true",
+        "build:custom": "grunt custom",
+        "build:custom:source": "grunt custom --source=true",
+        "build:dists": "grunt buildDists",
+        "build:sass": "grunt buildSass",
+        "deps": "grunt loadDependencies",
+        "lint": "grunt lint",
+        "watch:sass": "grunt watch:buildSass"
```
jobara, line 41: Please document these in the README and CONTRIBUTING files as needed.

greatislander:
Thanks for the review, @amb26 and @jobara. I'll proceed with README updates.

README.md
```diff
@@ -103,77 +103,95 @@ npm install
...
-grunt
+npm run build:all
 ```
 
-##### Custom Build
-
-Will only include the modules you request, and all of their dependencies, minus any that are explicitly excluded. Unlike
-the "all" build, none of the demos or tests are included with a custom package.
+By default all packages are minified. Running the build script with the `:source` suffix will allow you to maintain
+readable spacing and comments.
```
jobara, line 115: If we switch to Tenser is this still true? That is, will the comments be preserved or stripped out?

README.md
```diff
@@ -218,15 +236,28 @@ grunt custom --name="myPackage"
...
+To run only the browser tests:
+
+```bash
+npm run test:browser
+```
+
+Any browsers that Testem finds on your system will be launched sequentially with each browser running the full Infusion
+test suite. The results will be returned in your terminal in the [TAP](https://testanything.org/) format. Once you have
+run `npm install`, you can use the command `npx testem launchers` from the root of this repository to get a list of
+browsers that Testem can launch on your system.
```
jobara, line 260: This is no longer completely accurate. We actually only attempt to run the browser tests in Headless Chrome and Headless Firefox. We may modify that later, but that's it for now.

https://github.com/fluid-project/infusion/blob/8dfc99ed2c0c8f7e28b90419862f0d7f752c7cc8/tests/testem.js#L28-L29

README.md
greatislander: Looks like we need to explicitly enable them for the expanded output: https://terser.org/docs/api-reference#output-options

Can you add a comment to that effect to the terser PR, @jobara?

README.md
greatislander: @jobara should the instructions about using `npx testem launchers` be removed as it's no longer particularly helpful?

jobara: Hmm.. probably. It could help if an expected browser isn't running or if you wanted to see what else was available, but that's probably it.

greatislander:
@jobara Added `start` in ee38ff1.

jobara:
merged at 44bf2d380b17488d9523165eb973a698074a5a8b

PR #1039, opened 2020-11-24
FLUID-6582 Add pt_BR localization to preferences messages

gtirloni:


gtirloni:
I checked the output of codecov and it's suggesting I made changes to files that I didn't really change (e.g. Orator.js). I'm not sure what it's comparing against.

jobara:
> I checked the output of codecov and it's suggesting I made changes to files that I didn't really change (e.g. Orator.js). I'm not sure what it's comparing against.

I'm re-running the tests. I have a theory that some of the tests that have conditionals are having their checks time out in some runs and thus not running all of the tests and causing the coverage to drop.

jobara:
@gtirloni would you be able to add a translation for the simplify adjuster? It's currently only in the prefs framework demo, but we'd like to make a real preference and adjuster for it at some point. 

see: https://github.com/fluid-project/infusion/tree/main/demos/prefsFramework/messages

gtirloni:
> @gtirloni would you be able to add a translation for the simplify adjuster?

Done!

jobara:
Merged at 39773eb69938b6fa0bb9e3ae8f083775d2754f2f

PR #1040, opened 2021-02-10
FLUID-6598 Use Node.js 14.x and 15.x (deprecate 12.x)

gtirloni:


amb26:
Node 12.x is in "maintenance LTS" for a good while so I think we should continue to test with it

gtirloni:
@amb26 you're right, I completely misread the charts. Sorry about that, I had the impression 12.x would be going out of support this month.

PR #1042, opened 2021-03-25
FLUID-6607: remove normalize.css dependency

greatislander:
This PR required minimal CSS changes, as follows:

- In several components and throughout demos, `button`, `input`, and `select` elements were given the following CSS to preserve their expected font family and size: `font-family: inherit; font-size: 100%`.
- In the separated panel preferences editor stylesheet, a `margin: 0` rule was added to the `body` element to preserve the prefs editor panel's position flush at the top of the window.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 /.bundle/
 /src/lib/hypher/
 /src/lib/jquery/
-/src/lib/normalize/
 /src/lib/open-dyslexic/
 /src/lib/opensans/
 /src/lib/roboto/
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -200,7 +200,6 @@ module.exports = function (grunt) {
                 "src/lib/jquery/core",
                 "src/lib/jquery/plugins",
                 "src/lib/jquery/ui/js",
-                "src/lib/normalize",
                 "src/lib/open-dyslexic",
                 "src/lib/opensans",
                 "src/lib/roboto",
@@ -336,11 +335,6 @@ module.exports = function (grunt) {
                         "node_modules/jquery-ui/ui/widgets/tooltip.js"
                     ],
                     dest: "src/lib/jquery/ui/js/",
-                    expand: true,
-                    flatten: true
-                }, {
-                    src: "node_modules/normalize.css/normalize.css",
-                    dest: "src/lib/normalize/css/",
                     expand: true,
                     flatten: true
                 }, {
--- a/README.md
+++ b/README.md
@@ -226,7 +226,6 @@ Distribution bundles can be [viewed on unpkg](https://unpkg.com/browse/infusion/
 * jQueryUI
 * jQueryScrollToPlugin
 * jQueryTouchPunchPlugin
-* normalize
 * open-dyslexic
 * opensans
 * roboto
--- a/README.md
+++ b/README.md
@@ -230,7 +230,6 @@ Distribution bundles can be [viewed on unpkg](https://unpkg.com/browse/infusion/
 * open-dyslexic
 * opensans
 * roboto
-* url-polyfill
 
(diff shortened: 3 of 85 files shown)


greatislander:
Not sure why this test is failing: https://github.com/fluid-project/infusion/blob/ca1e594bd6883a5e002a50bb77521c5ecfd35fc0/tests/framework-tests/preferences/js/SeparatedPanelPrefsEditorTests.js#L148-L150

amb26:
I can't reproduce this failure in Firefox on Windows. I checked that the tests are properly using the "temporary" in-memory store so there shouldn't be any possibility for the store behaviour to be platform-dependent.

amb26:
What's extremely mysterious is that the failure in this branch is consistent - I've rerun the tests 3 times now, and on the other hand the pass in #1043 is also consistent which I've rerun twice

jobara:
I was able to reproduce this in FF on macOS, although not consistently.

It seems to be an issue with the panelIndex being included sometimes.

```JSON
{   
    "panelIndex": 0,  
    "preferences": {     
        "lineSpace": 2,     
        "textFont": "verdana",     
        "textSize": 1.8,     
        "theme": "bw"   
    } 
}
```

greatislander:
> It seems to be an issue with the panelIndex being included sometimes.

That's very odd. As far as I can tell the only changes my PR makes are removing Normalize from the module dependencies and HTML files and a couple of small CSS adjustments. That test failure is a very bizarre side effect.

amb26:
> It seems to be an issue with the panelIndex being included sometimes.

Thanks for the great investigation, @jobara - the "old" framework was pretty racy with model updates and so a small change in rendering timings may mean that the initial rendering races differently with the preferences updates. I think that the soundest update would be to adjust the test so that it only compares the preferences part of the model - I've written up this issue at https://issues.fluidproject.org/browse/FLUID-6610

jobara:
@amb26 and @greatislander some more info. It seems the panelIndex change is coming from [onScroll](https://github.com/fluid-project/infusion/blob/4e804fee39403575f893565548124f6e61b00490/src/framework/preferences/js/ArrowScrolling.js#L102-L111) event listener. This doesn't seem to fire in main, but does trigger from this branch.

jobara:
@amb26 and @greatislander it might be related to `fluid.prefs.arrowScrolling.scrollToPanel` which will attempt to [scroll](            scrollContainer.scrollLeft(scrollContainer.scrollLeft() + panels.eq(panelIndex).offset().left);) the container to the left to the specified panel.

jobara:
There's a strange issue with the overview panel in the Prefs framework demo. When the line space increase from 1.0 to 1.1 the line space in the overview panel actually decreases. Increasing the linespace further does increase the spacing. This isn't an issue in main.

greatislander:
> Moving a comment I filed [#1045 (comment)](https://github.com/fluid-project/infusion/pull/1045#issuecomment-811020214)
> 
> There's a similar issue with the toggle switch inputs. These can be seen in the same prefs framework demo.

This is now resolved.

jobara:
@greatislander could you please merge this with main?

jobara:
In Safari 14.0.3 (macOS 11), the Orator's "Play" button is smaller than the icon to the left. 

In the [Orator demo](https://build-infusion.fluidproject.org/demos/orator/) from main:
Screen Shot 2021-04-01 at 8 49 45 AM

In the Orator demo from this PR:
Screen Shot 2021-04-01 at 8 49 39 AM

greatislander:
> In Safari 14.0.3 (macOS 11), the Orator's "Play" button is smaller than the icon to the left.

This is fixed now, @jobara.

jobara:
merged at f346ada4eefc1b056c35aaab39885e87285c3594

PR #1043, opened 2021-03-25
FLUID-6608: remove url-polyfill dependency

greatislander:
The URL api is now available in all of our supported browsers (https://caniuse.com/url) so I think it can safely be removed.


jobara:
@greatislander could you please update the README to remove reference to url-polyfill as well?

PR #1046, opened 2021-04-13
FLUID-6594: Migrate build to use npm scripts

jobara:
Notable differences:
* only includes minified builds
* preamble/banner in minified js builds
  * In a single line
  * Indicates build options
  * Includes version using tag name or package version with "-dev" appended
* zip/archive includes version using tag name or package version with "-dev" appended
* default build compiles dists
* dist directory only contains minified js files

https://issues.fluidproject.org/browse/FLUID-6594


jobara:
### Windows build failure

Regarding the failing Windows builds, they are related to the [command substitutions](https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html) used by some of the scripts. These are used to pull in output from one npm script into another. Typically for complex commands or where piping is not supported. For example the [`build:util:concat`](https://github.com/jobara/infusion/blob/13dc880d2019ef276581a73f759f99c9923e4f99/package.json#L39) script.

```JSON
"build:util:concat": "terser $(npm run build:util:modules --silent) --source-map --format \"preamble='$(npm run build:util:preamble --silent)'\""
```
Command Substitutions are not directly supported in Windows. Although [stack overflow](https://stackoverflow.com/a/2768660/9878263) mentions how to achieve something similar. This does not appear to be cross-platform compliant though. I also have not been able to find an npm package to perform the task we need.

#### Note about Terser

Tersers [CLI API](https://terser.org/docs/cli-usage) takes in files to be concatenated immediately after the command name. We can also pass them in at the end if a `--` is placed between the flags and the list of files. If no input files are specified, it reads from stdio. It seems that reading from stdio expects code, not files names to be entered.

### Exploration using environment variables
I started exploring moving the output of `build:util:modules` to environment variables. To achieve that I added an additional output option for buildModules.js which could then be redirected into a temporary .env file. I used [dotenv-cli](https://www.npmjs.com/package/dotenv-cli) and [cross-vars-no-babel](https://www.npmjs.com/package/cross-var-no-babel) to read in the contents of the temporary file and access them through environment variables. 

```JSON
"build:util:concat": "dotenv -e .env_build_temp cross-var terser %FL_BUILD_FILES% --source-map --format \"preamble='$(npm run build:util:preamble --silent)'\"",
```

As described in this [post](https://www.genui.com/resources/env-variables-json) by Ray Knight, we need to use the `%variable%` syntax even on macOS to ensure that the variable is not replaced before we load our environment variables in. 

However, this also does not work because it treats the list of file names as a single string which gets interpreted by terser as a single file name and throws an `ENAMETOOLONG` error. 

### Other possible solutions

- Require all platforms use bash or powershell (see: https://stackoverflow.com/a/55303370)
- Find a way to configure the CI windows builds to use bash and require Windows users to manually configure this themselves
- Find another way of passing the list of files to Terser, and address any other command substitution instances. (mostly these are for the preamble/banner in the concatenated file).
- Find an alternative to Terser that would work better for our case
- Remove the notion of custom builds and manually write up the list of files to concat and minify for the distributions and infusion-all package.
- Move more of the logic into buildModules.js so that it also handles the concatenation and minification.

I'm leaning towards the last option, moving more of the logic into buildModules.js. @amb26 do you have any thoughts and suggestions?

amb26:
> Move more of the logic into buildModules.js so that it also handles the concatenation and minification.
I'm leaning towards the last option, moving more of the logic into buildModules.js. @amb26 do you have any thoughts and suggestions?

I think I agree - in practice there's nothing that can be done with elaborate shell subsitution that can't be done more stably and comprehensibly with a wrapper around the invocation of terser. I wouldn't say that this is "handling the concatenation in buildModules.js" as it is exposing an API for terser that is more directly oriented to our needs. It seems that all of the features that we would need are exposed via a programmatic API: https://github.com/terser/terser#api-reference

jobara:
Some lingering issues:

- `npm run build:pkg` and `npm run build:pkg:custom` don't work on windows do to access of environment variables.
  - It might be possible to support using [cross-var](https://www.npmjs.com/package/cross-var-no-babel) but looks like it may require to use Windows syntax `%variable%` which doesn't support fallbacks like the unix variant does `${variable:-fallback}`
- sourceMaps build with a root value `../` so the the dists will link to source files from src. However, with packages, the source is adjacent and the expectation is that the files will not be extracted within infusion's directory.
  - One solution is to only specify the the root `../` for dist builds.
  - Another options is to have all the builds be the same and copy src into the dist directory.

README.md
```diff
@@ -99,136 +99,121 @@ All other development dependencies will be installed by running the following fr
...
+The `FL_INCLUDE` environment variable takes in a comma-separated string of the [Modules](#modules) to be included in a
+custom package. If omitted, all modules will be included (demos and tests will not be included).
 
 ```bash
-npm run build:custom -- --include="inlineEdit, uiOptions"
+FL_INCLUDE="fluid-inline-edit, fluid-ui-options" npm run build:pkg:custom
 ```
 
-##### --exclude
+##### FL_EXCLUDE
```
amb26, line 151: I probably missed some relevant discussion on this - why did we change from a command line option to an env variable?

buildModules.js
```diff
@@ -0,0 +1,332 @@
...
+ * @property {String} execFile - Path to the file being executed.
+ *                                   enclosing DOM element
+ * @property {Array} files - An array of file paths.
+ * @property {String} [include] - (optional) A comma separated string of modules to be included.
+ * @property {String} [exclude] - (optional) A comma separated string of modules to be excluded.
+ */
+
+/**
+ * Processes the argv command line arguments into an object. Options are expected to be key/value pairs in the format of
+ * `--key="value"`. If no value is provided the option is set to `undefined`. Arguments provides which are not options
```
amb26, line 44: Typo provided

buildModules.js
```diff
@@ -0,0 +1,332 @@
...
+const dayjs = require("dayjs");
+const fg = require("fast-glob");
+const fs = require("fs");
+const mkdirp = require("mkdirp");
+const path = require("path");
+const {execSync} = require("child_process");
+const {minify} = require("terser");
+
+/**
+ * Command line arguments processed into an object.
```
amb26, line 30: Would the use of a prepackaged system such as https://www.npmjs.com/package/minimist help here?

buildModules.js
```diff
@@ -0,0 +1,332 @@
...
+ * @return {String} - the directory path
+ */
+build.getDir = (filePath, ...paths) => {
+    var basePath = path.dirname(filePath);
+    return path.join.apply(null, [basePath, ...paths]);
+};
+
+/**
+ *
+ * @param {Array} files - The set of file paths to the package.json files to source the Infusion modules from.
```
amb26, line 115: String[] ?

buildModules.js
```diff
@@ -0,0 +1,332 @@
...
+        }
+    });
+    return modules;
+};
+
+/**
+ * Returns the set of requested modules/packages based on their dependency requirements. If the include options is
+ * provided only the requested modules will be returned. If the exclude option is provied, the exclude modules will be
+ * removed.
+ * @param {Object} packages - All available packages
```
amb26, line 138: Object<String, String> ?

src/components/inlineEdit/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
amb26, line 1: Lovely to see this

package.json
```diff
@@ -52,37 +95,34 @@
     },
     "devDependencies": {
         "@fluid-project/hyphenation-patterns": "0.2.2-dev.20181115T221631Z.b2d0651",
+        "bestzip": "2.2.0",
         "browserslist-config-fluid": "0.1.0",
+        "cpy-cli": "3.1.1",
```
amb26, line 100: I don't see any use of this dependency in this pull

README.md
jobara: Simply, this is so to support passing the values down between npm scripts. You can use the `--` notation to pass some arguments down but it gets confusing when a scrips runs multiple other scripts or they go down a few levels. In migrating more functionality to buildModules.js we did remove some of this complexity in the scripts so we might actually be able to use the command line arguments again. I'll explore that.

package.json
jobara: it's used for the dependency copying, moving files from nodue_modules to the lib directory. However, the command is just `cpy` as this is just a wrapper package to add a CLI API.

buildModules.js
jobara: Thanks for the tip on using minimist.

README.md
jobara: In my latest push I have removed all of the environment variables out of the npm scripts related API. However, they are still optional for use with buildModules.js

jobara:
@amb26 I paired up with @jhung to test the builds on Windows using PowerShell. However, we ran into issues with the custom builds. Basically it seems that npm isn't respecting `--` separator and is parsing the args itself instead of passing them down. Well in some cases it passes down the values but not the flags. 

```powershell
# will trigger npm's help, should have outputted the help from buildModules.js
npm run build:pkg:custom -- -h

# will pass "fluid-undo" to the underlying npm script
npm run build:pkg:custom -- -i fluid-undo 

# will no pass any of the arguments down to the npm script
npm run build:pkg:custom -- --include=fluid-undo
```

This also seems to be the case with the custom builds from main, which also rely on the `--` separator to pass down args.

Note that using the `--` separator within npm scripts in package.json works fine. So this might just be a bug with PowerShell or NPM, but I haven't yet been able to locate any info on it.

There is a workaround for Windows. buildModules.js accepts environment variables. Which means that the following could be used to create a custom build:

```powershell
$env:FL_INCLUDE = 'fluid-undo'
npm run build:pkg:custom
```

Do you have any thoughts on this? Do you know if I'm doing something wrong, or running the commands incorrectly on Windows?

amb26:
> # will trigger npm's help, should have outputted the help from buildModules.js
npm run build:pkg:custom -- -h

That's odd because I can't reproduce this on my system. I tried, for example, "npm run lint -- -h" and got lint's help as expected.
I tried this on the native shell, Powershell, and Cygwin's bash shell and got the expected results on each. I also can't find any reports that powershell defeats this behaviour, perhaps there's something odd about @jhung's configuration?

If this issue remains stubborn, what I seem to find is recommendations for a helper script as in "solution 2" on this answer https://stackoverflow.com/a/51401577 or else this one https://stackoverflow.com/a/59248672

jobara:
@amb26 I setup a Windows VM to testing things out locally and things worked okay for me. So you might be right that it may be something specific to @jhung's config. I'm going to ask @BlueSlug if he can test on his Windows machine.

jobara:
@amb26 this PR is ready for more review. I paired with @BlueSlug to test the issues with Windows builds. We weren't able to reproduce the issues that @jhung saw. However, we did come across another issue with the naming of the zip files. This has been addressed as well.

@amb26 some things to test:
- distributions: `npm run build`
- infusion-all build: `npm run build:pkg`
- custom builds: `npm run build:pkg:custom -- -i fluid-undo -n myPackage -e fluid-framework` or some similar custom package
- tagged builds. If the build is generated from a tag, the concatenated files will mention the tag name in the banner instead of the dev version. e.g. with tag "infusion - myTag", without tag "infusion - 3.0.0-dev". Similarly the name of the zip file will include the dev version or tag name. To test simply run `git tag -a myTag -m "my tag comment"` but make sure to delete after testing by running `git tag -d myTag`

BlueSlug:
I've tested @jobara's changes since we paired earlier today. The issue we discovered has been solved, and I didn't note any further bugs.

During my testing, I discovered that the `name` option is ignored if no modules are included or excluded. This result was unexpected on my part (though it makes perfect sense), so perhaps it's worth clarifying in the documentation:

> value: "custom suffix" (String) only available to custom packages

could be amended to something like.

> value: "custom suffix" (String) only available to custom packages. If all modules are included and none are excluded, the name "all" will be used.

jobara:
@BlueSlug thanks for the suggestion. I've touched up the docs for the name option.

buildModules.js
```diff
@@ -0,0 +1,368 @@
...
+ *
+ * Environment Variables:
+ * - npm_package_name (from package.json): package name
+ * - npm_package_version (from package.json): package version
+ *
+ * System:
+ * - date: current date
+ * - branch: git branch the build is generated from
+ * - revision: git revision the build is generated from
+ * - tag: git tag the build is generated from (if HEAD corresponds to a tag)
```
amb26, line 191: Looks like this actually ends up in "version" via the git describe call

amb26:
Thanks, I found this all worked very well - only one very minor wrinkle that the "git describe" invocation didn't pick up my first attempt at a tag without an annotation since it was considered "lightweight". I think this is actually how we had traditionally created all tags in the past so perhaps we should cater for this case by using "git describe --tags" to include these tags

jobara:
> Thanks, I found this all worked very well - only one very minor wrinkle that the "git describe" invocation didn't pick up my first attempt at a tag without an annotation since it was considered "lightweight". I think this is actually how we had traditionally created all tags in the past so perhaps we should cater for this case by using "git describe --tags" to include these tags

@amb26 thanks for the feedback. I've modified the `git describe` command to use the `--tags` flag so that lightweight tags will also be picked up. I made some other doc/comment changes as well. Ready for more review.

PR #1048, opened 2021-05-12
FLUID-6619: announce contrast themes with VoiceOver/Edge

greatislander:
This restores accessible labels to VoiceOver with Edge/Chrome on macOS. Should be tested on with Firefox on Windows using NVDA/JAWS to make sure it doesn't regress the behaviour that the `aria-label` was added to address.

### DIFF
--- a/src/framework/preferences/js/Panels.js
+++ b/src/framework/preferences/js/Panels.js
@@ -778,9 +778,6 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                 theme: themeValue
             }));
 
-            // Aria-label set to prevent Firefox from reading out the display 'a'
-            label.attr("aria-label", themeValue);
-
             var labelTheme = theme[index];
             if (labelTheme === defaultThemeName) {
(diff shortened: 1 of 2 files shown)


jobara:
I tested this out (see configurations below). I appears to be outputting the labels of the contrast options. Note that I didn't have audio working in the Windows VM, so had to rely on the speech viewers for the screen readers.

I tested using the following configurations:

- macOS 11
 - VoiceOver
   - Chrome
   - Firefox
   - MS Edge
   - Safari
- Windows 10
 - JAWS
   - Chrome
   - Firefox
   - MS Edge
 - NVDA
   - Chrome
   - Firefox
   - MS Edge

greatislander:
@jobara: the decrease in coverage is unavoidable because the code is simpler (in other words, we are no longer adding an aria attribute and then testing to make sure it is there and has the correct value). I wonder if this might be a good opportunity to revisit the threshold configuration for CodeCov to account for changes like this?

jobara:
merged at 75c3c7e12a96f0ed168e8e2c9691ce5b061309c3

PR #1049, opened 2021-05-13
FLUID-6620: convert Textfield Stepper to number input

greatislander:
This PR resolves [FLUID-6620](https://issues.fluidproject.org/browse/FLUID-6620) by updating the Textfield Stepper component throughout Infusion to a standard `<input type="number>` ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number)). This resolves the perplexing behaviour in VoiceOver on Safari where announced values are represented as a decimal based on `aria-valuenow` / `aria-valuemax` for inputs with the `spinbutton` role instead of the displayed value (see: https://bugs.webkit.org/show_bug.cgi?id=221102).

## Additional Information

The `number` input type has native `stepUp()` and `stepDown()` methods for incrementing and decrementing its value. It might be good to leverage these somehow instead of using the custom logic defined in `fluid.textfieldStepper.step`, but that's a nice to have.

### DIFF
--- a/src/components/textfieldControl/css/TextfieldStepper.css
+++ b/src/components/textfieldControl/css/TextfieldStepper.css
@@ -23,11 +23,18 @@
 /* Textfield Styling */
 
 .fl-textfieldStepper .fl-textfieldStepper-textField {
+    -moz-appearance: textfield;
     font-family: inherit;
     font-size: 100%;
     margin: 0 0.5rem;
     text-align: center;
     width: 7rem;
+}
+
+.fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-inner-spin-button,
+.fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-outer-spin-button {
+    appearance: none;
+    margin: 0;
 }
 
 /* Textfield Button Styling */
--- a/src/components/textfieldControl/css/TextfieldStepper.css
+++ b/src/components/textfieldControl/css/TextfieldStepper.css
@@ -33,7 +33,7 @@
 
 .fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-inner-spin-button,
 .fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-outer-spin-button {
-    appearance: none;
+    -webkit-appearance: none;
     margin: 0;
 }
 
--- a/src/components/textfieldControl/js/TextfieldStepper.js
+++ b/src/components/textfieldControl/js/TextfieldStepper.js
@@ -80,13 +80,6 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                             // down arrow === 40
                             args: ["{that}.container", "keydown", 40, "{textfieldStepper}.decrease"]
                         }
-                    },
-                    modelListeners: {
-                        "value": {
-                            "this": "{that}.container",
-                            method: "attr",
-                            args: ["value", "{change}.value"]
-                        }
                     }
                 }
             },
--- a/src/components/textfieldControl/js/TextfieldStepper.js
+++ b/src/components/textfieldControl/js/TextfieldStepper.js
@@ -51,12 +51,17 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                                     "range.min": {
                                         "this": "{textfield}.container",
                                         method: "attr",
-                                        args: ["aria-valuemin", "{change}.value"]
+                                        args: ["min", "{change}.value"]
                                     },
                                     "range.max": {
                                         "this": "{textfield}.container",
                                         method: "attr",
-                                        args: ["aria-valuemax", "{change}.value"]
+                                        args: ["max", "{change}.value"]
+                                    },
+                                    "step": {
+                                        "this": "{textfield}.container",
+                                        method: "attr",
+                                        args: ["step", "{change}.value"]
                                     }
                                 }
                             }
@@ -74,18 +79,13 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                             listener: "fluid.textfieldStepper.bindKeyEvent",
                             // down arrow === 40
                             args: ["{that}.container", "keydown", 40, "{textfieldStepper}.decrease"]
-                        },
-                        "onCreate.addRole": {
-                            "this": "{that}.container",
-                            method: "attr",
-                            args: ["role", "spinbutton"]
                         }
                     },
                     modelListeners: {
                         "value": {
                             "this": "{that}.container",
                             method: "attr",
-                            args: ["aria-valuenow", "{change}.value"]
+                            args: ["value", "{change}.value"]
                         }
                     }
                 }
--- a/src/components/textfieldControl/js/TextfieldStepper.js
+++ b/src/components/textfieldControl/js/TextfieldStepper.js
@@ -68,19 +68,7 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                         }
                     },
                     attrs: "{textfieldStepper}.options.attrs",
-                    strings: "{textfieldStepper}.options.strings",
-                    listeners: {
-                        "onCreate.bindUpArrow": {
-                            listener: "fluid.textfieldStepper.bindKeyEvent",
-                            // up arrow === 38
-                            args: ["{that}.container", "keydown", 38, "{textfieldStepper}.increase"]
-                        },
-                        "onCreate.bindDownArrow": {
-                            listener: "fluid.textfieldStepper.bindKeyEvent",
-                            // down arrow === 40
-                            args: ["{that}.container", "keydown", 40, "{textfieldStepper}.decrease"]
-                        }
-                    }
+                    strings: "{textfieldStepper}.options.strings"
                 }
             },
             increaseButton: {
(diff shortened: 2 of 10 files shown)


src/components/textfieldControl/css/TextfieldStepper.css
```diff
@@ -23,13 +23,20 @@
 /* Textfield Styling */
 
 .fl-textfieldStepper .fl-textfieldStepper-textField {
+    -moz-appearance: textfield;
```
amb26, line 26: Which current browsers does this tweak display in? Is there a route to eliminating it someday?

src/components/textfieldControl/js/TextfieldStepper.js
```diff
@@ -74,18 +79,13 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
...
-                            method: "attr",
-                            args: ["role", "spinbutton"]
                         }
                     },
                     modelListeners: {
                         "value": {
                             "this": "{that}.container",
                             method: "attr",
-                            args: ["aria-valuenow", "{change}.value"]
+                            args: ["value", "{change}.value"]
```
amb26, line 88: It seems from the MDN docs that this attribute controls a default rather than a current value - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#value - what's the accessibility impact of removing this listener?

amb26: Sorry, I should more properly say, "replacing this listener with one that changes the actual control value", via some means like $.val() or the direct "value" property, etc.

jobara: This should already be handled by the textfield component. https://github.com/fluid-project/infusion/blob/75c3c7e12a96f0ed168e8e2c9691ce5b061309c3/src/components/textfieldControl/js/Textfield.js#L44-L50

src/framework/preferences/css/sass/PrefsEditor.scss
```diff
@@ -193,12 +193,19 @@ $icon-font: 'PrefsFramework-Icons';
...
         text-align: center;
         width: 4em;
     }
 
+    .fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-inner-spin-button,
+    .fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-outer-spin-button {
+        appearance: none;
+        margin: 0;
+    }
+
```
jobara, line 208: Shouldn't these new additions already be coming from TextfieldStepper.css?

tests/component-tests/textfieldControl/js/TextfieldStepperTests.js
```diff
@@ -89,7 +89,7 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 
         fluid.tests.textfieldStepper.verifyState = function (that, expectedValue, incBtnState, decBtnState) {
             jqUnit.assertEquals("The model value should be " + expectedValue, expectedValue, that.model.value);
-            jqUnit.assertEquals("The aria-valuenow should be " + expectedValue, expectedValue.toString(), that.textfield.container.attr("aria-valuenow"));
+            jqUnit.assertEquals("The value should be " + expectedValue, expectedValue.toString(), that.textfield.container.attr("value"));
```
jobara, line 92: @greatislander related to @amb26 comment above about the value of the input field, this should like get the actual value using something like `that.textfield.container.val()`.

src/components/textfieldControl/css/TextfieldStepper.css
greatislander: As the `-moz` prefix would suggest, it's needed for Firefox to display this input without the default stepper controls. The route to eliminating it would be for Firefox to provide an alternative method of doing this that doesn't require a browser prefix.

src/framework/preferences/css/sass/PrefsEditor.scss
greatislander: The prefs framework and UIO demos don't include the TextfieldStepper stylesheet— it appears that for some historical reason, those styles were duplicated in the Prefs Framework stylesheet.

src/components/textfieldControl/js/TextfieldStepper.js
greatislander: It appears that this listener can indeed be removed as it's handled in the textfield component. Resolved in https://github.com/fluid-project/infusion/pull/1049/commits/fe7b771984c45608edce74a6485453ea5cff0b5d.

tests/component-tests/textfieldControl/js/TextfieldStepperTests.js
greatislander: I've added a test to do this in 06f3533.

src/framework/preferences/css/sass/PrefsEditor.scss
jobara: Hmm... maybe it was to simplify the imports, not sure.

src/components/textfieldControl/css/TextfieldStepper.css
```diff
@@ -23,13 +23,20 @@
...
     font-family: inherit;
     font-size: 100%;
     margin: 0 0.5rem;
     text-align: center;
     width: 7rem;
 }
 
+.fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-inner-spin-button,
+.fl-textfieldStepper .fl-textfieldStepper-textField::-webkit-outer-spin-button {
+    appearance: none;
```
jobara, line 36: It appears this needs to be `-webkit-appearance: none;` for it to work on Safari.

How it looks as it currently is:
Number input in Safari

greatislander: Resolved in 33aa7dc.

jobara:
The input fields are generally smaller now. Ideally the input field should be the same height as the stepper buttons on the left and right. (It's a bit hard to parse, but you can see some of the older [designs](https://wiki.fluidproject.org/display/fluid/%28Floe%29+User+Interface+%28Learner%29+Options+Mobile+and+Responsive+Design) as a reference and [FLUID-5536](https://issues.fluidproject.org/browse/FLUID-5536))

e.g. from the [build site](https://build-infusion.fluidproject.org/demos/uioptions/) in Safari
Screen Shot 2021-05-13 at 9 44 02 AM

from Safari with the changes from this PR
Screen Shot 2021-05-13 at 9 45 14 AM

Other browsers are displaying the input size smaller as well, and may be partially related to removing normalize in [FLUID-6607](https://issues.fluidproject.org/browse/FLUID-6607)

jobara:
When using VoiceOver on macOS with Chrome and MS Edge, keyboard support doesn't work quite right. If you navigate to the input field, pressing the <kbd>↑</kbd> or <kbd>↓</kbd> shifts focus to the "+" button on the initial press. This prevents the arrow keys from incrementing/decrementing the values and the button is also not interactive with the keyboard. If you <kbd>shift</kbd>+<kbd>tab</kbd> back, you can use the <kbd>↑</kbd> or <kbd>↓</kbd> keys as expected.

greatislander:
> The input fields are generally smaller now.

Resolved in 1eab77d.

src/components/textfieldControl/js/TextfieldStepper.js
```diff
@@ -74,18 +79,6 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                             listener: "fluid.textfieldStepper.bindKeyEvent",
                             // down arrow === 40
                             args: ["{that}.container", "keydown", 40, "{textfieldStepper}.decrease"]
-                        },
```
jobara, line 77: This probably aren't necessary anymore and may be causing the issue with the arrow keys that I mentioned in VO and Chrome/MS Edge.

greatislander: Resolved in 6d0ef96.

tests/component-tests/textfieldControl/js/TextfieldStepperTests.js
```diff
@@ -148,17 +148,17 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                         changeEvent: "{stepper}.applier.modelChanged"
                     }, {
                         // decrease to minimum
-                        jQueryTrigger: fluid.tests.textfieldControl.createKeyEvent("keydown", 40),
-                        element: "{stepper}.dom.textfield"
+                        jQueryTrigger: "click",
+                        element: "{stepper}.dom.decreaseButton"
```
jobara, line 152: Did the keyboard interaction tests stop working after the change? It would be good to test if the up/down arrow keys still work to increment and decrement the model values.

greatislander: I reverted this change and now the tests hang locally (but on the TextFieldSlider tests, not this one). I'm going to see if they run in GitHub actions.

greatislander: Looks like they don't. No output, no messages.

jobara: The "No output" is probably because the IoC testing framework is waiting for an event that never fires. In the browser you get an error in the console for this, but jqUnit is kind of stuck. I filed a [FLUID-6622](https://issues.fluidproject.org/browse/FLUID-6622) about this. 

Anyways, the error is probably because the browser is ignore the synthetic keyboard interaction. There are some interactions that the browser requires to be user directed to actually work. It seems that this might be one of those cases. In that case, we won't be able to test the keyboard interaction, and you can go back to using the mouse click on the buttons.

Thanks for looking into this though.

greatislander: That was my instinct based on a previous conversation we'd had which was why I removed it. I'll revert the revert.

PR #1050, opened 2021-05-19
FLUID-6623: Backport KETTLE-49 system for writable DataSources to Infusion 3 branch

amb26:


jobara:
@amb26 are you able to amend the first commit to include the JIRA number?

src/framework/core/js/DataSource.js
```diff
@@ -14,7 +14,10 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 (function ($, fluid) {
     "use strict";
 
-    /** NOTE: Much of this work originated from https://github.com/fluid-project/kettle/blob/main/lib/dataSource-core.js **/
+    /** NOTE: Much of this work originated from
+     * https://github.com/fluid-project/kettle/blob/main/lib/dataSource-core.js and
+     * https://github.com/fluid-project/kettle/blob/main/lib/dataSource-url.js
+     ***/
```
jobara, line 20: Should these link to a particular branch/tag/revision instead of from main. That way it links to the actual versions of those files that it was derived from. For example, if those files are deleted or renamed the links will no longer work.

PR #1060, opened 2021-10-18
FLUID-6605, FLUID-6260: Use CSS custom properties for styling related Enactors

jobara:
Makes use of CSS custom properties for styling related enactors. Also split apart the CSS property definitions from the default styling application. This allows for the current approach of styles that will attempt to be "drop in" and work; while also providing an option for an integrator to more easily use their own styling by hooking directly into the custom properties in their own stylesheets.

- https://issues.fluidproject.org/browse/FLUID-6260
- https://issues.fluidproject.org/browse/FLUID-6605
- https://issues.fluidproject.org/browse/FLUID-6687
- https://issues.fluidproject.org/browse/FLUID-6688
- https://issues.fluidproject.org/browse/FLUID-6690


tests/framework-tests/preferences/html/LetterSpaceEnactor-test.html
```diff
@@ -53,11 +51,14 @@ <h2 id="qunit-banner"></h2>
...
-        <p class="flc-letterSpace" style="display: none; font-size: 16px">
+        <p class="flc-letterSpace" style="display: none; font-size: 16px; letter-spacing: var(--fl-letterSpace)">
             Sample Text
         </p>
         <div style="display: none; font-size: 16px; letter-spacing: 0.5em">
-            <p class="flc-letterSpace-existing" style="letter-spacing: 0.2em">
+            <p class="flc-letterSpace-existing" style="letter-spacing: var(--fl-letterSpace, 0.2em)">
+                Sample Text
+            </p>
+            <p class="flc-letterSpace-factor" style="letter-spacing: calc(calc(var(--fl-letterSpace-factor, 0) * 1em) + 0.2em)">
```
amb26, line 61: See below

src/framework/preferences/js/Enactors.js
```diff
@@ -358,6 +376,13 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                 "model.value": "value"
             }
         },
+        styles: {
+            enabled: "fl-lineSpace-enabled"
+        },
+        cssCustomProp: {
+            factor: "--fl-lineSpace-factor",
```
amb26, line 383: Given --fl-lineSpace-factor does not appear in any CSS definitions, I'm wondering why we store it in a custom property rather than just as an intermediate value in code?

src/framework/preferences/js/Enactors.js
```diff
@@ -310,6 +315,13 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                 "model.value": "value"
             }
         },
+        styles: {
+            enabled: "fl-textSize-enabled"
+        },
+        cssCustomProp: {
+            factor: "--fl-textSize-factor",
```
amb26, line 322: Similar to --fl-lineSpace-factor I don't see that this is used in a CSS definition

src/framework/preferences/css/sass/EnhanceInputs_base.scss
```diff
@@ -0,0 +1,6 @@
+.fl-input-enhanced {
+    --fl-enhance-factor: 1.25;
```
amb26, line 2: This value seems to be used in just one place. Is there a use case for adjusting it somehow?
Given it appears to be tied to the next value, if we keep it, could be named --fl-enhance-font-size-factor

src/framework/preferences/js/Enactors.js
jobara: the "factor" CSS custom properties provide hooks for integrators to use. The way that the programmatic style changing enactors typically work is to get the current size of the related CSS computed style, then apply the size increase to that. In this case, the line-height might be set to 1.2. Changing that in UIO to 2, would set line-height to 2.4. However, an integrator may have different line-heights set for the different parts of the page. If they only have the 2.4 value, they'd be setting all the line-heights to the same. With `--fl-lineSpace-factor` an integrator could set the line height like:

```css
body {
    line-height: var(--fl-lineSpace, 1.2);
}

.summary {
    line-height: calc(var(--fl-lineSpace-factor, 1) * 1.4);
}
```

jobara:
@amb26 do you have suggestions for the name "factor"? I guess it doesn't always have to be used as a multiplier, it could be used as an addition. Does the name still make sense?

src/framework/preferences/js/Enactors.js
```diff
@@ -405,20 +448,31 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
         return fluid.roundToDecimal(parseFloat(lineHeight) / fontSize, 2);
```
amb26, line 448: Note that the bug referenced on line 444 applied only to old versions of IE and was closed 11 years ago - we should be able to axe the work-around now

amb26: There's also another reference to the expired FLUID-4500 on line 433

src/framework/preferences/js/Enactors.js
jobara: Docs makes sense, I was also thinking a tutorial for integrating with them might help too. I've filed this as https://issues.fluidproject.org/browse/FLUID-6691

src/framework/preferences/js/Enactors.js
jobara: I've removed those and updated the tests.

jobara:
@greatislander I've updated the README file with information about the Sass files and CSS custom properties. I'm still not sure of better naming convention for the Sass files. Wondering if things are okay to merge as is, and we can continue to explore better naming options in future updates?

greatislander (approved):

greatislander (approved):

jobara:
@amb26 I made some additional changes since you last reviewed. Would you like to have another look and merge if you are happy?

PR #1064, opened 2021-11-02
FLUID-6696: Ensure that ToC is shown when initialized.

jobara:
https://issues.fluidproject.org/browse/FLUID-6696

### DIFF
--- a/src/framework/preferences/js/Enactors.js
+++ b/src/framework/preferences/js/Enactors.js
@@ -460,6 +460,11 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                 options: {
                     listeners: {
                         "afterRender.boilAfterTocRender": "{fluid.prefs.enactor.tableOfContents}.events.afterTocRender",
+                        // An integrator may wish to style the ToC container but not have it displayed when not
+                        // in use. In that case the container will likely be styled with `display: none` to hide it.
+                        // The following ensures, that the container is made visible when the ToC is enactor is enabled
+                        // and tableOfContents component has initialized.
+                        // see: https://issues.fluidproject.org/browse/FLUID-6696
                         "onReady.show": {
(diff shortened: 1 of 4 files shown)


jobara:
@jhung could you please take a look at the changes here and confirm if the address the issue you identified?

src/framework/preferences/js/Enactors.js
```diff
@@ -459,7 +459,11 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
                 createOnEvent: "onCreateTOCReady",
                 options: {
                     listeners: {
-                        "afterRender.boilAfterTocRender": "{fluid.prefs.enactor.tableOfContents}.events.afterTocRender"
+                        "afterRender.boilAfterTocRender": "{fluid.prefs.enactor.tableOfContents}.events.afterTocRender",
+                        "onCreate.show": {
```
amb26, line 463: Either this is a bug in the ToC core implementation in which case this fix should go there, or else it is part of the enactor's workflow, in which case it should manifest as an extra call to {enactor}.applyToc - either way we should have a comment here explaining the situation and reference to FLUID-6696

tests/framework-tests/preferences/html/Enactors-test.html
```diff
@@ -88,7 +88,7 @@ <h2 id="qunit-userAgent"></h2>
         </div>
 
         <div class="flc-tableOfContents">
-            <div class="flc-toc-tocContainer"></div>
+            <div class="flc-toc-tocContainer" style="display: none"></div>
```
amb26, line 91: Add a comment here explaining that this attribute is explicitly a test for FLUID-6696

jobara: I've added comments to both the test html and the enactor code.

PR #1065, opened 2021-11-22
FLUID-6697, FLUID-6698: Improvements to DataBinding for applications with materialised DOM binding

amb26:
- suggested by work on rewriting TextFieldControls. Renamed dom.attrs to dom.attr for consistency with dom.class. Better validation for "target" of model relays. Permit transforming short-form relays from non-model sources. fluid.transform.parseClasses transform, and allow relay from constants.

### DIFF
--- a/src/framework/core/js/DataBinding.js
+++ b/src/framework/core/js/DataBinding.js
@@ -1133,8 +1133,8 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 
     fluid.funcToSingleTransform = function (that, mrrec) {
         if (mrrec.func) {
-            if (mrrec.args !== undefined ^ mrrec.source === undefined) {
-                fluid.fail("Error in model relay definition: short-form relay must specify either args or a source and not both");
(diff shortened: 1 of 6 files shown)


jobara:
This is probably outside the scope of this PR, but wondering how the `fluid.materialisers.domOutput` might handle the distinction between jQuery's `prop` and `attr` methods?

amb26:
> This is probably outside the scope of this PR, but wondering how the fluid.materialisers.domOutput might handle the distinction between jQuery's prop and attr methods?

Well, I think I have decided in the meantime to no longer believe in the "prop" methods - because they address material that can't be adequately serialised back out of the DOM again - and if you can't serialise it, it isn't "materialised". One of the main principles of new renderer is that material should have "just the same effect" (as far as possible) on the server as on the client, and if you can't get it into the markup, it's hard to say that what you've done really has such a meaning. I think the "properties" idea was well-intentioned, in that it was an attempt to apply some light kind of "type system" on top of the markup which just dealt in strings, but I think that we can probably safely leave it out of scope and go back to "attr" just as we did in the beginning.

PR #1067, opened 2021-12-13
FLUID-6703, FLUID-6704, FLUID-6706: Fixes supporting prefs framework rewrite

amb26:
FLUID-6703: Fix and test case for id materialisation precedence
FLUID-6704: Move priority of template fetch to after model resolution
FLUID-6706: Support synchronous template fetches

### DIFF
--- a/src/framework/core/js/ResourceLoader.js
+++ b/src/framework/core/js/ResourceLoader.js
@@ -733,6 +733,10 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 
     fluid.resourceLoader.loaders.resourceText.noPath = true;
 
+    // Implementation note: Whilst there will not be a type error if the user writes a non-promise return, there will
+    // be a race failure as the "resource" initialiser in fluid.resourceFromRecord loses to the transformEvent listener,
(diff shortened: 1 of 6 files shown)


src/framework/core/js/ResourceLoader.js
```diff
@@ -733,6 +733,10 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 
     fluid.resourceLoader.loaders.resourceText.noPath = true;
 
+    // Implementation note: Whilst there will not be a type error if the user writes a non-promise return, there will
+    // be a race failure as the "resource" initialiser in fluid.resourceFromRecord loses to the transformEvent listener,
+    // overwriting its result with the FetchOne value as, e.g. fluid.getForComponent resolves synchronously. As a result
+    // we have to say we can't support synchronous "early" resources.
```
jobara, line 739: Isn't it possible to wrap this in a promise, or is it the fact that it would still resolve immediately a problem?

amb26: Yes, the race is to do with timing rather than the return type, but thanks for drawing my attention back to this comment which I can now delete/update since the issue is resolved

PR #1068, opened 2021-12-16
FLUID-6705: prevent line-space from affecting position of sliding panel tab

jobara:
https://issues.fluidproject.org/browse/FLUID-6705


cindyli:
Merged at 0104629aa6f3980617801aec12a8ff0ab874024b

PR #1069, opened 2022-02-01
FLUID-6710: Removing all file-scope closures from the framework preparing for 4.0.0 release

amb26:


demos/keyboard-a11y/js/five-star.js
```diff
@@ -11,194 +11,193 @@ You may obtain a copy of the ECL 2.0 License and BSD License at
...
 */
 
-var demo = demo || {};
-
-(function ($, fluid) {
-    "use strict";
+"use strict";
 
-    fluid.registerNamespace("demo.fiveStar");
+var demo = demo || {};
```
jobara, line 16: Out of curiosity, why isn't enough to just call `fluid.registerNamespace("demo.fiveStar");`

jobara: Oh I guess because of the function declarations.

src/framework/renderer/js/fluidRenderer.js
```diff
@@ -11,10 +11,9 @@ You may obtain a copy of the ECL 2.0 License and BSD License at
 https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 */
 
-(function ($, fluid) {
-    "use strict";
-
+"use strict";
 
+(function ($) {
```
jobara: Is this still needed for jQuery? It is removed from other files.

tests/component-tests/reorderer/js/ImageReordererTests.js
```diff
@@ -13,8 +13,9 @@ https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 
 /* global jqUnit */
 
+"use strict";
+
 (function ($) {
-    "use strict";
```
jobara: Can this be removed?

src/framework/renderer/js/fluidRenderer.js
amb26: I left the closure in since there are a lot of local variables that would otherwise leak out into the global scope, and it is not worth reengineering this file much since it will be deleted in the next release. I could remove the jQuery argument from the closure however.

tests/component-tests/reorderer/js/ImageReordererTests.js
amb26: The closure, you mean? There is an obscure timing-related issue about initialising the variable "k" in the inner scope that I didn't feel the need to look into the causes of, so I left it in.

tests/component-tests/textfieldControl/html/TextfieldStepper-test.html
```diff
@@ -62,5 +61,7 @@ <h2 id="qunit-userAgent"></h2>
         <button class="flc-textfieldStepper-increase">+</button>
     </div>
 
+    <script src="../js/TextfieldStepperTests.js"></script>
```
jobara: I wonder if we could use [`defer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-defer) for these cases.

jobara:
@amb26 .fluidlintallrc.json can be updated to remove the reference tin infusion-1.5.js

tests/component-tests/reorderer/js/ImageReordererTests.js
jobara: ok

src/framework/renderer/js/fluidRenderer.js
jobara: I wonder if we should remove jQuery. On one hand it probably doesn't matter, on the other, it might be confusing to and make someone in the future question why it's there.

tests/component-tests/textfieldControl/html/TextfieldStepper-test.html
amb26: Good suggestion - this seems to work

src/framework/renderer/js/fluidRenderer.js
amb26: Simplified this and ImageReordererTests closures

jobara:
@amb26 I did a quick look at the dynamic reorderer manual test and grid reorderer demo. They are both failing with an error from Reorderer.js line 463.

```log
Uncaught TypeError: dom.fastLocate(...)[(intermediate value)] is not a function
    hoverStyleHandler http://localhost:3000/src/components/reorderer/js/Reorderer.js:463
    togo http://localhost:3000/src/framework/core/js/FluidIoC.js:2849
    fire http://localhost:3000/src/framework/core/js/Fluid.js:1764
    initMovable http://localhost:3000/src/components/reorderer/js/Reorderer.js:564
    jQuery 2
Reorderer.js:463:75
    hoverStyleHandler http://localhost:3000/src/components/reorderer/js/Reorderer.js:463
    togo http://localhost:3000/src/framework/core/js/FluidIoC.js:2849
    fire http://localhost:3000/src/framework/core/js/Fluid.js:1764
    initMovable http://localhost:3000/src/components/reorderer/js/Reorderer.js:564
    jQuery 2
```

amb26:
> @amb26 I did a quick look at the dynamic reorderer manual test and grid reorderer demo. They are both failing with an error from Reorderer.js line 463

Thanks for checking this out, @jobara - I have a one-line fix for this issue, are you happy for me to add it to this branch, given I think it is an issue with the main branch rather than the work here?

jobara:
> > @amb26 I did a quick look at the dynamic reorderer manual test and grid reorderer demo. They are both failing with an error from Reorderer.js line 463
> 
> Thanks for checking this out, @jobara - I have a one-line fix for this issue, are you happy for me to add it to this branch, given I think it is an issue with the main branch rather than the work here?

If it isn't too much hassle it might be clearer in a separate PR given the size of this one, it could easily be overlooked if we need to look back on the change at some point.

amb26:
> If it isn't too much hassle it might be clearer in a separate PR given the size of this one, it could easily be overlooked if we need to look back on the change at some point. And given the issue, maybe we should merge that one fix first? What do you think?

Thanks @jobara - I've written up the issue at https://issues.fluidproject.org/browse/FLUID-6711

We can put it in immediately after this goes in - no point issuing a pull now since it will just conflict

examples/framework/preferences/minimalFootprint/js/loadScripts.js
```diff
@@ -11,59 +11,53 @@ You may obtain a copy of the ECL 2.0 License and BSD License at
 https://github.com/fluid-project/infusion/raw/main/Infusion-LICENSE.txt
 */
 
-var fluid_4_0_0 = fluid_4_0_0 || {}; // eslint-disable-line no-redeclare
-var fluid = fluid || fluid_4_0_0; // eslint-disable-line no-redeclare
-
-(function (fluid) {
```
jobara, line 17: This file is actually used to load Infusion only when needed. So the `fluid` namespace does not exist yet, and will need to be added for the functions below.

jobara:
@amb26 a couple more references to infusion-1.5.js

- in .fluidlintrc.json [line 63](https://github.com/amb26/infusion/blob/116eaba60b6397c32df3e65e037284af619385b7/.fluidlintallrc.json#L63)
- in .eslintignore [line 2](https://github.com/amb26/infusion/blob/116eaba60b6397c32df3e65e037284af619385b7/.eslintignore#L2)

PR #1073, opened 2022-05-06
FLUID-6728, FLUID-6729: Cases from ADTKINS involving sourcePath

amb26:
https://issues.fluidproject.org/browse/FLUID-6728 - sourcePath of Number 0 doe not get interpolated
https://issues.fluidproject.org/browse/FLUID-6729 - sourcePath cannot be injected into model

Preparing for 4.1.0 Infusion release

### DIFF
--- a/src/components/inlineEdit/package.json
+++ b/src/components/inlineEdit/package.json
@@ -15,7 +15,7 @@
     "dependencies": {
         "jquery": "3.6.0",
         "jquery-ui": "1.12.1",
-        "fluid-framework": "4.0.0",
+        "fluid-framework": "^4.0.0",
         "fluid-undo": "4.0.0",
         "fluid-tooltip": "4.0.0"
     }
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "infusion",
     "description": "Infusion is an application framework for developing flexible stuff with JavaScript",
-    "version": "4.0.0",
+    "version": "4.1.0",
     "author": "Fluid Project",
     "bugs": "http://issues.fluidproject.org/browse/FLUID",
     "homepage": "http://www.fluidproject.org/",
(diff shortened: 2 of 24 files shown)


src/components/inlineEdit/package.json
```diff
@@ -15,7 +15,7 @@
     "dependencies": {
         "jquery": "3.6.0",
         "jquery-ui": "1.12.1",
-        "fluid-framework": "4.0.0",
+        "fluid-framework": "^4.0.0",
```
cindyli, line 18: Is there a reason to change the version type of `fluid-framework` from "4.0.0" to "^4.0.0"?

cindyli: > Is there a reason to change the version type of `fluid-framework` from "4.0.0" to "^4.0.0"?

After reading the code below, seems the reason is to have the installer fetch the latest 4.1.0 if it's available. This makes sense at the dev stage. Shall we take a note when cutting a new infusion version,  all dependencies still use the same exact version rule?

amb26: No, I don't think so. We are moving to a model where the framework will be released and versioned separately from the components, and this version update is meant to communicate that the framework has been bumped one version number but the components haven't. The caret is meant to leave room for further such updates - e.g. that inlineEdit version 4.0.0 is expected to be compatible with any framework version 4.x

amb26:
> The test coverage is lower than the configured 90%. Shall this be improved?

Yes although primarily I plan to improve it by removing untested and obsolete material from the framework. I don't believe the codecov figure is very accurate though since it is always reporting false positives in coverage reduction as it is in this pull - in fact this pull request leaves code coverage unchanged. I am also getting a rate limit error in accessing the coverage statistics from codecov today

src/components/inlineEdit/package.json
cindyli: I see. Thanks.

PR #1074, opened 2022-05-07
FLUID-6730: Removing some obsoleted utilities

amb26:


jobara:
@amb26 could you please update the change log?

jobara:
@amb26 what about the following:

- fluid.keys -> [Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)
- fluid.values -> [Object.values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values)
- fluid_prefix, fluid.guid, and/or fluid.allocateGuid -> [Crypto.randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)

amb26:
> * fluid.keys -> [Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)
> * fluid.values -> [Object.values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values)

These two aren't exact replacements since as the doc comment at https://github.com/fluid-project/infusion/blob/main/src/framework/core/js/Fluid.js#L819 explains, fluid.keys returns properties in the prototype chain as well as own properties, which we sometimes rely on.

> * fluid_prefix, fluid.guid, and/or fluid.allocateGuid -> [Crypto.randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)

This isn't a good replacement since we lose the ability to associate component ids with order of construction, which is extremely helpful when debugging, as well as the crypto ids being much bigger and slower to generate. Also crypto.randomUUID is not implemented on node 14.x which is still in LTS.

CHANGELOG.md
```diff
@@ -21,6 +21,13 @@ Infusion from v2.0.0 onwards adheres to [Semantic Versioning](https://semver.org
 
 ### More Info
 
+## [4.1.0] - 2022-05-09
```
jobara, line 24: Will you actually be cutting a release for this? If so I guess the date needs to be updated now, if not these can all go to the unreleased section until we're ready to do the release.

CHANGELOG.md
```diff
@@ -21,6 +21,13 @@ Infusion from v2.0.0 onwards adheres to [Semantic Versioning](https://semver.org
 
 ### More Info
 
+## [4.1.0] - 2022-05-11
```
jobara, line 24: The link is missing for 4.1.0.

amb26: > The link is missing for 4.1.0.

Good catch, thanks!

PR #1080, opened 2022-05-12
FLUID-6735: Test case for issue

jobara:
Test case for:

https://issues.fluidproject.org/browse/FLUID-6735


amb26:
Cheers @jobara - I reworked this test case in #1081 which resolves the circular reference problem. As I remarked in the channel, the problem is caused by the "misaligned" material in the parent grade which can't avoid trying to resolve a reference into its model from its options as it constructs, which is not permitted in Infusion. If/when Infusion 6 is written, this class of problem will be removed by removing the distinction between "models" and "general options" but until then I'm not sure we can do much better about issuing a helpful diagnostic although I can try. Hopefully this reworking is good enough to let you make progress on your panelIndex issue.

jobara:
@amb26 I was able to get past this issue in UIO+ by passing in a grade that contained the model configuration. Similar to what you described except I used the gradeNames directly.

amb26:
Reworked test case into #1081 which has been merged

PR #1082, opened 2022-05-27
FLUID-6738: Implementation and test case for "rootMaker" workflow function

amb26:


amb26:
Mothballed since @duhrer has said this functionality is not soon required

PR #1083, opened 2022-07-01
FLUID-6741: Test case and fix for construction of lensed component triggered by resource via model listener

amb26:


tests/all-tests.html
```diff
@@ -74,7 +74,7 @@
             ],
             "Testing Framework Tests": [
                 "/test-core/testTests/html/ConditionalTestUtils-test.html",
-                "/test-core/testTests/html/IoCTesting-test.html",
+//                "/test-core/testTests/html/IoCTesting-test.html",
```
cindyli, line 77: Shall this be uncommented?

cindyli: Ah, I saw your PR https://github.com/fluid-project/infusion/pull/1084 and now understand why `IoCTesting-test.html` is commented out. Did you figure out later why this test hangs on CI?

amb26: Cheers, good catch - actually #1084 is already merged into this pull and so the test no longer hangs

PR #1084, opened 2022-07-01
NOJIRA: Test CI

amb26:


amb26:
Hi @cindyli, @gtirloni - some problem seems to have developed in GitHub CI that our browser tests are hanging partway through the sequence. It's pretty bizarre that this is happening on every GitHub platform, and also is something I can't reproduce locally. Any ideas?

amb26:
Cheers all, you can now ignore this - it was a genuine bug, as far as I can tell a latent race condition exposed by the upgrade to Chrome 103, fixed in #1085

PR #1088, opened 2022-07-18
FLUID-6745: Improve TextNodeParser so that it can be configured with a free hash of selectors to ignore

amb26:
Also improve impl so it does not report text nodes that consist entirely of whitespace, corresponding adjustment to Orator and its tests


jobara:
@amb26 I was testing the prefsFramework demo using FF 102. I noticed that some of the words are not being highlighted. For example the left navigation the "Resources" and "Multimedia" headers are skipped, with the highlighting remain gin on "Impacts" For the rest of the navigation, the first word of each line is not highlighted. 

Screen Shot 2022-07-18 at 6 42 29 PM

Did you say there was also an issue with Safari? Is there something you wanted me to test specifically for that?

jobara (approved):

PR #1089, opened 2022-07-20
FLUID-6736: fix self voicing in safari

cindyli:
To resolve [FLUID-6736](https://issues.fluidproject.org/browse/FLUID-6736)

The self voicing stops working in Safari because `voice` is not provided when creating `SpeechSynthesisUtterance` instances at speaking. Before, when the `voice` value is missing, browsers will use the default voice. This is no longer supported in Safari.

The fix is to find the corresponding voice for the language of the speech. When the language is not given, use "en-US" as default.

### DIFF
--- a/src/components/textToSpeech/js/TextToSpeech.js
+++ b/src/components/textToSpeech/js/TextToSpeech.js
@@ -77,7 +77,8 @@ fluid.defaults("fluid.textToSpeech", {
         utteranceOnStart: null
     },
     members: {
-        queue: []
+        queue: [],
+        voices: "@expand:{that}.getVoices()"
     },
     components: {
         wndw: {
@@ -267,11 +268,6 @@ fluid.textToSpeech.requestControl = function (that, control, change) {
     }
 };
 
-fluid.textToSpeech.getVoiceByLang = function (that, lang) {
-    var voices = that.getVoices() || [];
-    return voices.find(voice => voice.lang === lang);
-};
-
 /*
  * After an utterance has finished, the utterance is removed from the queue and the model is updated as needed.
  */
@@ -290,6 +286,25 @@ fluid.textToSpeech.handleEnd = function (that) {
         var newModel = $.extend({}, that.model, resetValues);
         that.applier.change("", newModel, "ADD", "handleEnd.reset");
     }
+};
+
+// Find the voice by the language code
+// 1. If the language code is not provided, use the default language;
+// 2. If the voice for the language code is not found, fall back to the first voice that supports the same country
+// code;
+// 3. If the voice is still not found, return the voice of options.defaultLanguage.
+fluid.textToSpeech.getVoiceByLang = function (that, lang) {
+    lang = lang || that.options.defaultLanguage;
+
+    var voiceTogo = that.voices.find(voice => voice.lang === lang);
+
+    if (!voiceTogo) {
+        // find the first voice that matches the country code
+        var indexOfSeparator = lang.indexOf("-");
+        var countryCode = indexOfSeparator > 0 ? lang.substring(0, indexOfSeparator) : lang;
+        voiceTogo = that.voices.find(voice => voice.lang.startsWith(countryCode));
+    }
+    return voiceTogo ? voiceTogo : that.voices.find(voice => voice.lang === that.options.defaultLanguage);
 };
 
 /**
@@ -325,7 +340,7 @@ fluid.textToSpeech.queueSpeech = function (that, text, interrupt, options) {
     }
 
     options = options || {};
-    options.voice = options.voice || that.getVoiceByLang(options.lang || that.options.defaultLanguage);
+    options.voice = options.voice || that.getVoiceByLang(options.lang);
     var utteranceOpts = $.extend({}, that.model.utteranceOpts, options, {text: text});
 
     // The setTimeout is needed for Safari to fully cancel out the previous speech.
--- a/src/components/textToSpeech/js/TextToSpeech.js
+++ b/src/components/textToSpeech/js/TextToSpeech.js
@@ -197,7 +197,7 @@ fluid.defaults("fluid.textToSpeech", {
         },
         getVoiceByLang: {
             funcName: "fluid.textToSpeech.getVoiceByLang",
-            args: ["{that}", "{arguments}.0"]
+            args: ["{that}.voices", "{that}.options.defaultLanguage", "{arguments}.0"]
         },
         speak: {
             func: "{that}.invokeSpeechSynthesisFunc",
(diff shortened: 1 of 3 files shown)


src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -260,6 +266,12 @@ fluid.textToSpeech.requestControl = function (that, control, change) {
     }
 };
 
+fluid.textToSpeech.getVoiceByLang = function (that, lang) {
+    lang = lang || "en-US";
```
jobara, line 270: rather than hardcoding to "en-US" this should use [Navigator.languages](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages) or [Navigator.language](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language) with a final fallback to something that is configured in the options.

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -145,7 +145,8 @@ fluid.defaults("fluid.textToSpeech", {
         utteranceOpts: {
             // text: "", // text to synthesize. Avoid using, it will be overwritten by text passed in directly to a queueSpeech
             // lang: "", // the language of the synthesized text
-            // voice: {} // a WebSpeechSynthesis object; if not set, will use the default one provided by the browser
+            // voice: {} // a WebSpeechSynthesis object; it must be set in safari. In other browsers, if not set,
+            //              will use the default one provided by the browser
```
jobara, line 149: I think you should probably [file a bug](https://bugs.webkit.org) against Webkit and indicate the issue in the comments here. According to [Bug 237614](https://bugs.webkit.org/show_bug.cgi?id=237614) they should have updated to the latest version of the Web Speech API. Which indicates that the [voice attribute](https://wicg.github.io/speech-api/#dom-speechsynthesisutterance-voice) is nullable and states 

> If this attribute is unset or null at the time of the [speak()](https://wicg.github.io/speech-api/#dom-speechsynthesis-speak) method call, then the user agent must use a user agent default voice.

I updated my computer this morning and got the latest Safari 15.6 and also tried the [Safari Technology Preview](https://developer.apple.com/safari/resources/) with the build site, and neither work. So unfortunately this is still an issue.

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -312,6 +324,8 @@ fluid.textToSpeech.queueSpeech = function (that, text, interrupt, options) {
         that.cancel();
     }
 
+    options = options || {};
+    options.voice = that.getVoiceByLang(options.lang);
```
jobara, line 328: I think this should be something like the following, so that a supplied voice in the options would be used, and the getVoice would be a fallback. 

```suggestion
    options.voice ??= that.getVoiceByLang(options.lang);
```

jobara:
@cindyli I noticed that your changes didn't work the first time I tried. But did later. I wonder if there is something similar to that chrome issue I mentioned where the getVoices doesn't populate a list right away.

cindyli:
> @cindyli I noticed that your changes didn't work the first time I tried. But did later. I wonder if there is something similar to that chrome issue I mentioned where the getVoices doesn't populate a list right away.

@jobara, I tried Chrome and Safari, cannot reproduce the same issue. But I noticed if I first used the old code in Safari when the speaking didn't work, then switched to the new code branch, refreshed the same page, text-to-speech did remain the not-working state. In this case, closing and re-opening Safari fixed the problem.

Could you give a try and let me know? Thanks.

jobara:
> @jobara, I tried Chrome and Safari, cannot reproduce the same issue. But I noticed if I first used the old code in Safari when the speaking didn't work, then switched to the new code branch, refreshed the same page, text-to-speech did remain the not-working state. In this case, closing and re-opening Safari fixed the problem.
> 
> Could you give a try and let me know? Thanks.

I tried again today. The Prefs Framework demo was working, but the Orator demo wasn't.

cindyli:
> The Prefs Framework demo was working, but the Orator demo wasn't.

Good catch, @jobara. Orator demo didn't work because [the language code specified in index.html](https://github.com/cindyli/infusion/blob/main/demos/orator/index.html#L2) is "en-CA" so that `Navigator.language` returns "en-CA" as the default language. However, the voice list returned by the speech synthesis doesn't support a voice for "en-CA".

@jobara, @amb26, how do we fall back in this case? My thought is we either hardcode the [`defaultLanguage`](https://github.com/cindyli/infusion/blob/FLUID-6736/src/components/textToSpeech/js/TextToSpeech.js#L65) to "en-US" or compute `defaultLanguage` to the first language code starting with "en-" in the speech synthesis voice list. 

What do you think?

jobara:
> > The Prefs Framework demo was working, but the Orator demo wasn't.
> 
> Good catch, @jobara. Orator demo didn't work because [the language code specified in index.html](https://github.com/cindyli/infusion/blob/main/demos/orator/index.html#L2) is "en-CA" so that `Navigator.language` returns "en-CA" as the default language. However, the voice list returned by the speech synthesis doesn't support a voice for "en-CA".
> 
> @jobara, @amb26, how do we fall back in this case? My thought is we either hardcode the [`defaultLanguage`](https://github.com/cindyli/infusion/blob/FLUID-6736/src/components/textToSpeech/js/TextToSpeech.js#L65) to "en-US" or compute `defaultLanguage` to the first language code starting with "en-" in the speech synthesis voice list.
> 
> What do you think?

Ah I was wondering about that situation. I suppose picking some other "en-*" from the list would be sufficient for now. It's difficult to know what would be the best fallback for all situations though. That is, "en-US" would be fine for Canada, but may not be the best for another locale. I guess we can hope for 2 potential future improvements. 

1. The safari issue gets fixed so we can go back to letting it select which voices to use by default
2. We provide a mechanism for the user to pick which synthesizer to use

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -292,6 +288,25 @@ fluid.textToSpeech.handleEnd = function (that) {
     }
 };
 
+// Find the voice by the language code
+// 1. If the language code is not provided, use the default language;
+// 2. If the voice for the language code is not found, fall back to the first voice that supports the same country
+// code;
+// 3. If the voice is still not found, return the voice of options.defaultLanguage.
```
amb26, line 295: Could you convert this into a proper doc comment?

src/components/textToSpeech/js/TextToSpeech.js
```diff
@@ -292,6 +288,25 @@ fluid.textToSpeech.handleEnd = function (that) {
...
+fluid.textToSpeech.getVoiceByLang = function (that, lang) {
```
amb26, line 296: Dependence on the compoinent is minimal and might work better with "voices" and "defaultLanguage" hoisted into arguments

cindyli:
Self voicing stops working in Safari due to [a webkit bug](https://bugs.webkit.org/show_bug.cgi?id=243055).

This pull request worked around the Safari bug by always assigning a speech synthesis voice at speaking instead of leaving the voice option to `null`. This change breaks some tests because the "fluid.textToSpeech.utterance" mock was not complete since it did nothing to override the construction of the native “utterance” member. We made a mock Utterance and found that the tests hang through expecting the “onStop” event to be fired at the expected time. To save the time spent on this, we made [the mock getVoices()](https://github.com/cindyli/infusion/blob/FLUID-6736/src/components/textToSpeech/js/MockTTS.js#L165) return actual speech synthesis voice list instead of a list of objects. This fixed broken tests.

Moreover, Ned will not use orator from infusion, and SelfVoicing is not configured as part of the out of the box configuration of UIO on any of our supported sites, so we can punt on this issue until we come to make the UIO 2 release - in the meantime Safari may have resolved their bug.

This PR is mothballed.

jobara:
I re-tested the issue with Safari 17.5 and wasn't able to reproduce the issue. It seems that the underlying Safari bug has been resolved.

PR #1093, opened 2022-09-26
FLUID-6750: Fix and test case for model-driven localisation on startup

amb26:


src/framework/core/js/ResourceLoader.js
```diff
@@ -127,33 +127,28 @@ fluid.fetchResources = function (resourceSpecs, callback, options) {
...
-                        var togo = {
-                            loader: resourceSpec.loader
-                        };
-                        togo[pathKey] = oneExploded;
-                        return togo;
-                    }, fluid.fetchResources.prepareRequestOptions);
-                }
-            }
+        // If options.defaultLocale is set, it will replace any defaultLocale set on an individual resourceSpec
+        resourceSpec.resolvedDefaultLocale = resourceSpec.defaultLocale || resourceFetcher.options.defaultLocale;
```
cindyli, line 131: The comment says, "If options.defaultLocale is set, it will replace any defaultLocale set on an individual resourceSpec". But the code in the next line has resourceSpec.defaultLocale taking precedent over options.defaultLocale. Shall the code be changed?

tests/framework-tests/core/js/DataBindingTests.js
```diff
@@ -903,23 +903,51 @@ fluid.tests.fluid6442.checkIt = function (component) {
...
+            func: "fluid.tests.fluid6442.checkIt",
+            args: "{that}"
+        },
+        "onResourceError.failTest": function (err) {
+            jqUnit.fail("Failure fetching resource: " + err);
+        }
+    }
+});
+
 jqUnit.asyncTest("FLUID-6442: Modelised relocalisation should be accessible via resources", function () {
```
cindyli, line 918: I have a question for this test:

1. If the model driven localization occurs at the startup with the initial model value `model.resourceLoader.locale === "en"`, the fetched message value of `component.resources.messages.parsed.courses` should be from `message1_en.json` file and equals to "These courses will require a lot of grading".
2. At line 924, when a model change on `resourceLoader.locale` is triggered, the `component.resources.messages.parsed.courses` value should be fetched from `message1_en_ZA.json` and equals to "These courses will require a lot of marking".

It seems the resource fetch at the startup doesn't happen for this test. I wonder why?

src/framework/core/js/ResourceLoader.js
amb26: Good catch, it is the comment that is wrong and has been for many years :)

tests/framework-tests/core/js/DataBindingTests.js
amb26: It doesn't happen because the model update occurs synchronously, before the I/O can resolve - as the comment reads, "Change the locale before the queued I/O can resolve. In addition this tests promise cancellation propagation.". Given the I/O is in progress, it is cancelled in favour of the new model update and so only one fetch is seen.

cindyli: Got it. Thanks for the explanation.

PR #1094, opened 2022-10-11
FLUID-6749: iframeless UIO

cindyli:
In the separated panel version of UIO, the preferences editor no longer renders in an iframe. It now renders in the same content page. The changes preferences are applied to the preferences editor using custom css properties.

The issue ticket: [FLUID-6749](https://issues.fluidproject.org/browse/FLUID-6749)


tests/framework-tests/preferences/js/PrefsEditorTestsCommon.js
```diff
@@ -39,3 +39,20 @@ fluid.tests.mergeMembers = function (memberDef) {
     }
     return togo;
 };
+
+// An add-on grade to override distributeOptions in fluid.prefs.separatedPanel component whose targets are
+// outside of the separatedPanel component
+fluid.defaults("fluid.tests.prefs.overrideUnOutreach", {
```
amb26, line 45: Slightly odd name - perhaps something like overrideBuilderDependencies or something which mentions the name of which component is not present in the test configuration?

PR #1098, opened 2022-11-18
FLUID-6754: add system-ui font option to UIO

jobara:
Adds the system-ui font to UIO

https://issues.fluidproject.org/browse/FLUID-6754

### DIFF
--- a/src/framework/preferences/css/sass/utils/_fonts.scss
+++ b/src/framework/preferences/css/sass/utils/_fonts.scss
@@ -2,6 +2,7 @@
 
 $fonts: (
     ".fl-font-arial": "Arial",
+    ".fl-font-system-ui": "system-ui",
     ".fl-font-verdana": "Verdana",
     ".fl-font-times": "Georgia, Times, Times New Roman, serif",
     ".fl-font-comic-sans": "Comic Sans MS, sans-serif",
--- a/src/framework/preferences/messages/textFont_fa.json
+++ b/src/framework/preferences/messages/textFont_fa.json
@@ -1,5 +1,6 @@
 {
     "textFont-default": "پیش فرض",
+    "textFont-system": "System Font",
     "textFont-times": "Times New Roman",
     "textFont-comic": "Comic Sans",
     "textFont-arial": "Arial",
--- a/src/framework/preferences/messages/textFont_fr.json
+++ b/src/framework/preferences/messages/textFont_fr.json
@@ -1,5 +1,6 @@
 {
     "textFont-default": "Défaut",
+    "textFont-system": "Police Système",
     "textFont-times": "Times New Roman",
     "textFont-comic": "Comic Sans",
     "textFont-arial": "Arial",
--- a/src/framework/preferences/messages/textFont_pt_BR.json
+++ b/src/framework/preferences/messages/textFont_pt_BR.json
@@ -1,5 +1,6 @@
 {
     "textFont-default": "Padrão",
+    "textFont-system": "Fonte do Sistema",
     "textFont-times": "Times New Roman",
     "textFont-comic": "Comic Sans",
     "textFont-arial": "Arial",
--- a/src/framework/preferences/css/sass/PrefsEditor.scss
+++ b/src/framework/preferences/css/sass/PrefsEditor.scss
@@ -68,6 +68,10 @@ $icon-font: 'PrefsFramework-Icons';
             margin: 0 auto;
 
             option {
+                &.fl-font-system-ui {
+                    font-family: var(--fl-font-family, system-ui);
+                }
+
                 &.fl-font-times {
                     font-family: var(--fl-font-family, "Times New Roman", serif);
                 }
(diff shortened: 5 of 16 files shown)


cindyli (approved):
This PR looks good to me. "System font" is also the best name I can come up with. Let's see what @amb26 thinks. He's much better at naming.

jobara:
We had a good discussion about naming today. The suggestion is to change "System Font" to "Device Font" as it is more understandable and descriptive about where that font is originating from. Another point raised is that "Default" is now ambiguous, we've landed on changing the name to "No Preference" to indicate that it isn't based on a font that is being applied by UIO.

src/framework/preferences/messages/textFont_fr.json
```diff
@@ -1,5 +1,6 @@
 {
-    "textFont-default": "Défaut",
+    "textFont-default": "No Preference",
+    "textFont-system": "Device Font",
```
jobara, line 3: @BlueSlug could you help with translating these two strings to French?

src/framework/preferences/messages/textFont_pt_BR.json
```diff
@@ -1,5 +1,6 @@
 {
-    "textFont-default": "Padrão",
+    "textFont-default": "No Preference",
+    "textFont-system": "Device Font",
```
jobara, line 3: @gtirloni would you be able to help translate these two strings?

src/framework/preferences/messages/textFont_fa.json
```diff
@@ -1,5 +1,6 @@
 {
-    "textFont-default": "پیش فرض",
+    "textFont-default": "No Preference",
+    "textFont-system": "Device Font",
```
jobara, line 3: Need Farsi translations.

src/framework/preferences/css/sass/utils/_fonts.scss
```diff
@@ -2,6 +2,7 @@
 
 $fonts: (
     ".fl-font-arial": "Arial",
+    ".fl-font-system-ui": "system-ui",
```
amb26, line 5: To avoid confusing ourselves in future, should we rename all these uses of "system" to "device" also?

jobara: Good question. The actual generic font-family is called [system-ui](https://w3c.github.io/csswg-drafts/css-fonts-4/#system-ui-def). So the right hand side can't be changed. We could change the name of the class to device if that's preferred. I'm fine either way, I guess we need to decide if we want the class name to reflect the font family or the name exposed via the adjuster.

src/framework/preferences/messages/textFont_fa.json
jobara: The Farsi translations have been added.

src/framework/preferences/messages/textFont_pt_BR.json
jobara: The recommendation from @gtirloni is:

No Preference => Sem Preferência
Device Font => Fonte do Dispositivo

src/framework/preferences/messages/textFont_fr.json
jobara: @BlueSlug has recommended

-     "No Preference" => "Pas de préférence"
-     "Device Font" => "Police de l’appareil"

jobara:
@amb26 and @cindyli we have the translations in place now. Ready for more review.

src/framework/preferences/messages/contrast_fa.json
```diff
@@ -1,5 +1,5 @@
 {
-    "contrast-default": "پیش فرض",
+    "contrast-default": "No Preference",
```
cindyli, line 2: The translation should be in Farsi instead of English.

The same comment applies to translation of "No Preferences" in contrast_fr.json, contrast_pt_BR.json.

jobara: Thanks for catching this. I had forgotten that those files were updated as well. Interesting thing I had missed before is that the localization preferences already used "No Preference" instead of "Default".

cindyli:
Steps to produce an issue using the code in this pull request:

1. Use a browser to open `examples/framework/preferences/localizationPreference/urlPath/fa/`. Note that the languauge is set to Farsi;
2. Open another browser tab to open `demos/prefsFramework/`;
3. Open UIO, locate "text style" panel, open its drop down, choices are 

Screenshot 2022-12-15 at 10 50 40 AM

Note that "No preference" is translated to Farsi but not other choices. Shall all choices be translated to Farsi?

jobara:
> Note that "No preference" and "device font" are translated to Farsi but not other choices. Shall all choices be translated to Farsi?
> 
> This issue happens for all language translations.

I believe in this case it is the correct behaviour. When we originally sent these off for translation they were sent the whole file, and those font names weren't translated. I believe it's because they are proper names so there isn't a translation, and/or they are more commonly known in their English name.

cindyli (approved):

jobara:
@amb26 did you have more thoughts on this PR? Should we talk more about the naming changes you were asking about before?

greatislander (approved):

PR #1102, opened 2023-03-12
FLUID-6764: Fix to jquery.standalone for use with AudioWorklets.

colinbdclark:
This PR updates the global scope detection code in jquery.standalone.js to support use within an AudioWorkletGlobalScope, which is a super-minimal environment that provides no references to the global scope at all.

I haven't included any tests for this fix, because AudioWorkletGlobalScope doesn't even provide a means to import or load scripts, so we'd have to prepare a specialized build just for this purpose, which would include a custom AudioProcessor class and wrap everything in an ES6 module. I did, however, confirm that all the current unit tests continue to pass.


amb26:
Thanks so much for the fix, will issue a release soon!

PR #1104, opened 2024-07-11
FLUID-6763: Add Atkinson Hyperlegible font to UIO

jobara:
- Adds the Atkinson Hyperlegible font to the UIO Text Font preference.
- Updates fonts to be sourced via Fontsource npm packages
- Updates the Node versions to run in CI to the latest LTS versions.


src/components/overviewPanel/package.json
```diff
@@ -13,8 +13,8 @@
     },
     "dependencies": {
         "jquery": "3.6.0",
-        "opensans-webkit": "1.1.0",
-        "roboto-fontface": "0.10.0",
+        "opensans": "5.0.29",
```
greatislander, line 16: @jobara It looks like you have changed to the `@fontsource` namespaced packages elsewhere— is there a reason these haven't changed? I may be misunderstanding the structure or the role of this package.json file.

jobara: @greatislander these sub package.json files are used for the [internal build system](https://github.com/fluid-project/infusion?tab=readme-ov-file#how-do-i-create-an-infusion-package). They refer to other package listings for internal modules. In this case they are defined in src/lib/package.json which you can see edited in this PR as well.

greatislander (changes requested):
I note that Atkinson Hyperlegible is not listed in the 'external libraries' section of the readme— should it be?

jobara:
> I note that Atkinson Hyperlegible is not listed in the 'external libraries' section of the readme— should it be?

I'm not sure I'm thinking of the same place as you are. But it is mentioned https://github.com/fluid-project/infusion/pull/1104/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R232 which is the first item in the list. It appears higher because of the names being alphabetized. Although they aren't all actually in the correct alphabetical order so we could fix that up or maybe group by type. Let me know if that's what you mean or if it was supposed to be somewhere else.

PR #1111, opened 2024-11-26
NOJIRA: (ci) replace deprecated workflows with current versions

greatislander:
Many of the workflows in use (e.g. `actions/checkout@v2`) are about to be deprecated. This PR updates them accordingly and configures the required token for `codecov/codecov-action@v5`. Also note that this PR is unable to switch to using the NPM caching provided by the `setup-node` action because the `package-lock.json` file is not committed to Git. We may want to reconsider this.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -18,30 +18,30 @@ jobs:
         node-version: [18.x, 20.x, 22.x]
 
     steps:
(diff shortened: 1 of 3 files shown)


jobara:
Regarding the package-lock.json file, that's part of the suggested topics for the Fluid and Infusion discussion scheduled for Dec 2, 2024.

PR #1115, opened 2024-12-06
FLUID-6785: add package-lock.json to version control

greatislander:
This PR also bumps the `package.json` engines property to stipulated that Node 18 is the minimum supported version, which I believe is correct based on the current CI configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
         "watch:sass:min": "npm run build:sass:min -- --watch"
     },
     "engines": {
-        "node": ">=12"
+        "node": ">=18"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
         "watch:sass:min": "npm run build:sass:min -- --watch"
     },
     "engines": {
-        "node": ">=12"
+        "node": ">=18"
```
amb26, line 88: Please could you drop this back to node 16 since I need to support this in a few environments - cheers!

